Issue #1, opened 2015-12-29
Add openhab-online distro


dvanherbergen:
The current distro is large and contains all addons.
A minimal openhab-online could be created, which is minimal and could download all features on demand from a remote repo.

kaikreuzer:
Let's hope that we do not get into too much trouble regarding a "update feature" functionality, which will be asked for soon after ;-)

PR #3, opened 2015-12-31
First version of online distribution.

dvanherbergen:
Initial implemenentation for #1 
- Move shared distribution resources to dedicated folder
- Add first version of online distribution


pom.xml
```diff
@@ -104,6 +104,14 @@
             <name>JCenter Repository</name>
             <url>https://jcenter.bintray.com/</url>
         </repository>
+         <repository>
```
kaikreuzer: this has already been added by https://github.com/openhab/openhab-distro/pull/2, so please rebase!

dvanherbergen:
Rebase done.

distributions/distribution-resources/src/main/resources/etc/org.ops4j.pax.url.mvn.cfg
```diff
@@ -56,47 +56,28 @@ org.ops4j.pax.url.mvn.useFallbackRepositories=false
...
-#    http://repository.springsource.com/maven/bundles/release@id=spring.ebr.release, \
-#    http://repository.springsource.com/maven/bundles/external@id=spring.ebr.external, \
-#    http://zodiac.springsource.com/maven/bundles/release@id=gemini, \
-#    http://repository.apache.org/content/groups/snapshots-group@id=apache@snapshots@noreleases, \
-#    https://oss.sonatype.org/content/repositories/snapshots@id=sonatype.snapshots.deploy@snapshots@noreleases, \
-#    https://oss.sonatype.org/content/repositories/ops4j-snapshots@id=ops4j.sonatype.snapshots.deploy@snapshots@noreleases, \
-#    http://repository.springsource.com/maven/bundles/external@id=spring-ebr-repository@snapshots@noreleases
+# This contains the openhab, smart home and default repositories.
+#
+org.ops4j.pax.url.mvn.repositories= \
```
kaikreuzer: These repos now all also end up in the offline distro. We have specifically removed them because we noticed that they lead to a bad startup performance, since Karaf checks all those repos (and runs into timeouts if offline).

Also for the online distro I would prefer to have only "our" repos (eclipse, jcenter, jfrog) configured to make sure that we have under control, what is available to the runtime. All stuff from Karaf itself should imho be already contained in the runtime and there should be no need to access anything of that remotely.

dvanherbergen: OK, I will update.

dvanherbergen: Do you want to keep 'our repos' in the offline install or not?

distributions/openhab-online/readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+
+1. start openhab
+	start.sh
+
+2. view available features
+	feature:list
+
+3. install a feature e.g. install rest docs
+	feature:install openhab-misc-restdocs
```
kaikreuzer: I think we need to discuss how to deal with "minimal" vs. "online" distro.
For me, the online distro should result in the same functional setup as the offline distro - the only difference is the fact that the one contains the artefacts locally, so that they can be installed from there, while the other uses a remote repo.

I am also thinking about how to best handle the "openhab-demo" feature installation - at the moment, people have to do it manually in the shell, which is definitely not an easy start, so we need to change this.
I was wondering about having a config option in the addons.cfg (where you can also specify which bindings etc. should be installed), something like "runtime = minimal", "runtime = normal", "runtime = demo". The distro could then always come as minimal and install at first startup additional features (with the benefit that you can also strip it down again by changing this configuration value). This would be exactly the same for online and offline distro. What do you think?

distributions/distribution-resources/src/main/resources/etc/org.ops4j.pax.url.mvn.cfg
kaikreuzer: No, because this would not solve the slow startup / timeout issues.

kaikreuzer: What would be nice through is a possibility to update the content on an offline distro as well. I was actually hoping that we could have a addons.kar file somewhere in the distro, from where features are resolved. Users would have been able to download a newer version and replace it locally. Even for the online version, we could have the same kar made available at a single place. But somehow @maggu2810 struggled getting it working this way. I don't remember the details, but maybe kars are not meant to be used that way?

distributions/openhab-online/readme.md
dvanherbergen: > For me, the online distro should result in the same functional setup as the offline distro 

A separate minimal and online. OK, understood. That makes sense.

> I am also thinking about how to best handle the "openhab-demo" feature installation - at the moment, people have to do it manually in the shell, which is definitely not an easy start, so we need to change this.

My preference would go to extending the dashboard a little bit with a default admin/feature installation page. So that a user can launch openhab, click on the URL displayed in the console on startup (we should add that) to go to the dashboard and easily install new features like the demo. Installation itself should of course be triggered using REST services.

> I was wondering about having a config option in the addons.cfg

I would rather start from the concept that once a user has installed openHAB, they would never have to download a new runtime again. They should be able to just update from within openHAB itself. Of course, this will only work for the online runtime. For the offline, you would still have to download. But won't this be the exception rather than the rule?

dvanherbergen: > My preference would go to extending the dashboard a little bit with a default admin/feature installation page.

This would also allow us to not specify all repository urls by default, but add them programmatically when needed. E.g only install the snapshot repos when the user has selected the 'include snapshots' checkbox in the admin page.

kaikreuzer: > My preference would go to extending the dashboard a little bit with a default admin/feature installation page

I was thinking about that once as well, but then decided to include the "extension management in UI" only in the Paper UI, which will be present for a "normal" setup and it comes with all nice UI features like event, touch device support etc.

Two further reasons why purely relying on an admin UI is not good:
- It means that the minimal runtime already requires an http server, which is actually pretty big. If you think of an "headless" openHAB instance, e.g. for only running a specific binding and sending data through MQTT, an http server is not needed.
- There must be a way to specify your setup, so that it can be backed up or automatically constructed. This is why I introduced the "addons.cfg" file already.

> This would also allow us to not specify all repository urls by default, but add them programmatically when needed

This again is rather a configuration setting for me (use snapshots or not) than a UI functionality. Of course you can also provide such features in the UI later on, but the basis must be a configuration setting.

> they would never have to download a new runtime again

This is something that I would leave out of scope for the moment - it would already be a big gain if we can easily update the bindings/extensions. Updating the runtime itself will be much more difficult as you would also have to cover files like "start.sh" or whatsoever. For the moment the upgrade should be easy enough by simply allowing to manually replace the "runtime" folder, while keeping the "conf" and "userdata" folder in place.

dvanherbergen: > It means that the minimal runtime already requires an http server, which is actually pretty big. If you think of an "headless" openHAB instance, e.g. for only running a specific binding and sending data through MQTT, an http server is not needed.

OK, we have a difference in understanding here. I always considered minimal runtime as the smallest possible openHAB 'master' instance. For running only a specific binding in a slave configuration, I believe a dedicated 'slave/node' runtime would be more appropriate as we could tailor it for that use, .e.g for me a slave runtime, would:
- be preconfigured to run in a slave configuration
- be autodiscovered by the master instance
- be fully configurable from the master instance


> why purely relying on an admin UI is not good

I agree that config should also be possible. But UI is in my opinion also needed for the master instance. If a user finds it too difficult to enter 'feature:install openhab-demo' on the command line, then filling out a config file will be a challenge too :wink: 

> that it can be backed up or automatically constructed

Backed up, I understand. I'm not sure what you mean with automatically constructed.

kaikreuzer: > I always considered minimal runtime as the smallest possible openHAB 'master' instance.

Well, this is then more or less what the openhab-runtime feature (https://github.com/kaikreuzer/openhab-core/blob/master/features/karaf/src/main/feature/feature.xml#L24) is meant for.
What I call "minimal" rather corresponds to the openhab-runtime-base feature (https://github.com/kaikreuzer/openhab-core/blob/master/features/karaf/src/main/feature/feature.xml#L14).

>  then filling out a config file will be a challenge too

Then openHAB isn't the right tool for that guy, because he will also have to go on and create item, sitemap and rule files, not to forget about the binding config files :-)
But yes: I think the config file could be delivered with "demo", so that this is automatically installed, if the user does not have any other configuration (e.g. because he only updated his instance and has kept his old "conf" folder).

> I'm not sure what you mean with automatically constructed.

Well, having a "conf" folder where all the individual setup is described - through item files, binding configuration files, etc., but hence also through the configuration of what addons have to be available (at the moment people copy their "addons" folder with the jars over for this purpose - it is more or less a replacement for the "addons" folder.
Think also about providing a preconfigured docker image for the RaspPi or other. There are many cases where you simply want to say, which addons should be part of your runtime.

kaikreuzer: > I'm not sure what you mean with automatically constructed.

Note that all of this is in place with the addon.cfg already. I'll enhance this with the "runtime" setting tomorrow, so you don't have to worry about this now in the context of this PR. For the distribution, it would therefore be enough if you simply have the feature openhab-runtime-base pre-installed - ok?

dvanherbergen: > For the distribution, it would therefore be enough if you simply have the feature openhab-runtime-base pre-installed - ok?

OK, I will update the repo config files and add the shk repo to the online, so that it becomes 'fully' online.

dvanherbergen:
Readme removed, mvn url cfg file split and temp shk repo added.
Changes squashed into the commit.

kaikreuzer:
Thanks. Shall we change the offline distro to openhab-runtime-base as well then, so that both are in sync?

Installing features in the online distro gives me those logs:
```
14:46:00.770 [WARN ] [n.internal.config.MavenRepositoryURL] - Repository spec http://oss.jfrog.org/libs-snapshot@oh-snapshot-repo/ does not contain an identifier. This is deprecated & discouraged & just evil.
14:46:00.771 [WARN ] [n.internal.config.MavenRepositoryURL] - Repository spec https://repo.eclipse.org/content/repositories/snapshots@esh-snapshot-repo/ does not contain an identifier. This is deprecated & discouraged & just evil.
```
Any idea what this means?

dvanherbergen:
> Shall we change the offline distro to openhab-runtime-base as well then, so that both are in sync?

OK, I will add that change.

> Any idea what this means?

Yes, I'll fix that.

dvanherbergen:
Done.

kaikreuzer:
looks good, thanks!

dvanherbergen:
There is one issue left. I trashed my local maven repo to make sure online was really online and it turns out that there are some errors when it tries to resolve the openhab-demo and openhab-esh-addons. These feature jars are created in the openhab-distro, but they are not published to a maven repo.
Can we add publishing of those jars in the distro or do you want to keep the distro a packaging only project? In the latter case, we should move the openhab-demo and openhab-esh-addons feature to a different project.

kaikreuzer:
I just tried to add the deployment to JFrog in the build plan, unfortunately, it failed: https://openhab.ci.cloudbees.com/job/openHAB-Distribution/40/console

Maybe I first have to get them on Bintray and request that snapshots should be published to JFrog.
Won't be able to do further look into it this year, will be back next year :-)

dvanherbergen:
no worries. enjoy the end of the year...

kaikreuzer:
you too!

kaikreuzer:
Stuff is online now: http://oss.jfrog.org/oss-snapshot-local/org/openhab/distro/ :-)

PR #8, opened 2016-01-03
minor cleanups

maggu2810:
* fix indentation of root pom
* remove trailing whitespaces in root pom
* remove repository definition of child pom, because the repos are already defined in root pom


maggu2810:
* fix indentation of offline dist pom
* remove trailing whitespaces in offline dist pom

Issue #10, opened 2016-01-03
Performance issues on embedded ARM (e.g. Raspberry Pi 2)


kaikreuzer:
This is a follow up of https://github.com/openhab/openhab2/issues/514, where we assumed that switching to G1 GC fixes most of the performance issues (see https://github.com/openhab/openhab2/issues/514#issuecomment-162176235).
Unfortunately, G1 is not supported on ARM, if you run the openHAB runtime on a Raspberry Pi 2, you will get
```
Java HotSpot(TM) Client VM warning: G1 GC is disabled in this release.
```
See also https://blogs.oracle.com/henrik/entry/oracle_releases_jdk_for_linux which states:
"Some examples of missing features include the G1 GC".

I did now tests on a Raspberry Pi 2 (with java 1.8.0_40-b27) and measured these times for startup (until the dashboard and all UIs of the demo are available and the CPU usage dropped below 95%):
 - first start: 22 minutes
 - subsequent start: 11 minutes

Even the installation of an additional binding through the Paper UI takes already more than 30 seconds - this is just adding a small bundle to the OSGi container...

The current openHAB 2 distribution (non-karaf) takes less than 30 seconds for startup, so the issue is clearly with Karaf.
It seems that hardly anybody before us has ever tried to use Karaf 4 on a Raspberry Pi. The "plain" Karaf 4.0.3 e.g. sets the "-server" jvm flag, which isn't available on ARM. And even this empty Karaf container alone takes more than a minute for the startup.

**THIS IS A SEVERE BLOCKER FOR MOVING AHEAD WITH OPENHAB 2!!!**

One of the goals of openHAB 2 was to make it better suited for embedded environments - the current solution is in contrast UNUSABLE on it.

I guess we might need some input and suggestions from the Karaf team in order to take a decision on how to move forward. @ANierbeck, @cschneider, @jbonofre: Was it a bad idea from the start to use Karaf on embedded ARM computers or do you see some hope for us?

ANierbeck:
This is a rather strange thing. Did you run a vanilla Karaf on a PI before, any measurements there?
I know that @jgoodyear did some testing of OpenDayLight (ODL) on a PI before, afaik he didn't spot such an issue.

kaikreuzer:
> Did you run a vanilla Karaf on a PI before, any measurements there?

Not before, but the "empty Karaf container alone takes more than a minute for the startup" was a test of vanilla Karaf today.

kaikreuzer:
[...]

PR #19, opened 2016-01-05
Add temporary patch of pax-web-jetty to avoid slow startup times.

dvanherbergen:
Temporary solution for #10 

### DIFF
--- a/distributions/distribution-resources/src/main/resources/etc/org.ops4j.pax.logging.cfg
+++ b/distributions/distribution-resources/src/main/resources/etc/org.ops4j.pax.logging.cfg
@@ -30,10 +30,6 @@ log4j.logger.org.openhab = INFO
 log4j.logger.org.eclipse.smarthome = INFO
 log4j.logger.org.jupnp = ERROR
 
-# workaround for https://github.com/openhab/openhab2/issues/563
-log4j.logger.org.ops4j.pax.web.service.jetty = ERROR
-log4j.logger.org.ops4j.pax.web.extender.whiteboard = ERROR
-
 # CONSOLE appender not used by default
 log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
--- a/distributions/openhab-offline/pom.xml
+++ b/distributions/openhab-offline/pom.xml
@@ -43,6 +43,13 @@
             <type>xml</type>
             <scope>runtime</scope>
         </dependency>
+        <!-- Temporary patch for https://github.com/openhab/openhab-distro/issues/10 -->
+        <dependency>
+          	<groupId>org.openhab.distro.deps</groupId>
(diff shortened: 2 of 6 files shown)


kaikreuzer:
Shall we maybe place the jar somewhere on Bintray (https://bintray.com/openhab/mvn) and include it from there? Then we would not need to have it in the source repo.

dvanherbergen:
That would work too. We'll need to give it a different name then to indicate the patch.
I don't have upload rights to the openhab bintray. If you can deploy it there with a new name, I will update the PR to use that one.

maggu2810:
Hm, if you use another GAV for the artifact, then you have to change the feature, to use the other one.
I think for the workaround it is ok to replace the jar file.
With this PR the bundled jar is replaced and so the feature work out of the box.

Or do I miss something?

kaikreuzer:
I was rather thinking to only update the assembly descriptor to pick up the file as a Maven dependency than from the local project - this way, it should not have any impact on the features as the result is identical to the current PR.

dvanherbergen:
During the assembly we could rename it back to the original name.

maggu2810:
Ah okay.
If the assembly plugin support the reference using GAV then this is IMOH the cleaner solution.

kaikreuzer:
Here you go: https://dl.bintray.com/openhab/mvn/org/openhab/distro/deps/pax-web-jetty/4.2.3/:pax-web-jetty-4.2.3.jar

maggu2810:
It is up to you, but shouldn't we set the version (in the repo) to 4.2.3.sp1 or something similar to indicate it is a patched / modified 4.2.3 one?

kaikreuzer:
https://dl.bintray.com/openhab/mvn/org/openhab/distro/deps/pax-web-jetty/4.2.3.sp1/pax-web-jetty-4.2.3.sp1.jar

Better? :-)

dvanherbergen:
PR updated to use mvn dependency.

kaikreuzer:
Thanks a lot!

JSurf:
Is this fix included in the current oh beta ? I tested oh2 offline distro in a raspbian jessie/java8.openjdk based docker container on a rpi2. Host OS is arch linux. Startup times are terribly slow ~ 20 minutes for first start. If yes it does not help for me

JSurf:
Just tried with the online distro makes no difference, startup is also painfully slow no difference to the offline distro

dvanherbergen:
@JSurf, there is something else happening on your system. I just did a test with rpi1. The initial startup takes 8 minutes. Subsequent starts take about 3 minutes. 
Can you try installing openHAB natively, not inside a docker container?

JSurf:
Tested outside docker environment. Seems my issues are related to my docker container/environment.
Startup times without docker seem to be Ok

JSurf:
Turned out that openjdk on raspbian seems to be super slow. Changed to oracle jdk provided by the rpi foundation. Now startup is very fast

PR #20, opened 2016-01-05
Check platform before setting G1GC.

### LINKED ISSUE #17: Warning  displayed when starting runtime on an ARM computer
Since G1 is not supported on JDK for ARM, calling start.sh displays
```
Java HotSpot(TM) Client VM warning: G1 GC is disabled in this release.
```
I wonder if we can tweak the start scripts to only add the jvm arg for G1 on non-ARM systems?

dvanherbergen:
Resolves #17 


kaikreuzer:
Wow, you are so productive, I should enter more issues :-)

dvanherbergen:
Feel free, but todays work is done :-)

PR #21, opened 2016-01-05
set configuration for demo package to use demo.sitemap as default

kaikreuzer:


dvanherbergen:
LGTM

PR #22, opened 2016-01-05
added an initial welcome echo when starting the runtime

kaikreuzer:
As startup can take a while, a first message for the user is always welcome - this is taken from the current openHAB 2 runtime.


dvanherbergen:
I couldn't spot any typos... :wink:

kaikreuzer:
And you thought your day's work is over ;-)
Thanks!

dvanherbergen:
It is now. Shutting down. Catch you later.

 :sleeping:

PR #25, opened 2016-01-06
Add default java.util.logging config to disable lock messages at startup

### LINKED ISSUE #15: openhab.log contains karaf lock messages
The first entries in openhab.log are 
```
Jan 04, 2016 10:29:15 PM org.apache.karaf.main.Main launch
INFO: Installing and starting initial bundles
Jan 04, 2016 10:29:20 PM org.apache.karaf.main.Main launch
INFO: All initial bundles installed and set to start
Jan 04, 2016 10:29:20 PM org.apache.karaf.main.lock.SimpleFileLock lock
INFO: Trying to lock /home/pi/oh/userdata/lock
Jan 04, 2016 10:29:20 PM org.apache.karaf.main.lock.SimpleFileLock lock
INFO: Lock acquired
Jan 04, 2016 10:29:20 PM org.apache.karaf.main.Main$KarafLockCallback lockAquired
INFO: Lock acquired. Setting startlevel to 100
```
I would very much prefer that they were not logged. As they are logged before log4j is configured, I am not sure how to get rid off them, though.

dvanherbergen:
Resolves #15 

### DIFF
--- a/distributions/distribution-resources/src/main/resources/userdata/etc/java.util.logging.properties
+++ b/distributions/distribution-resources/src/main/resources/userdata/etc/java.util.logging.properties
@@ -1,0 +1,6 @@
+# set default format layout
+java.util.logging.SimpleFormatter.format = %1$tF %1$tT.%1$tL [%4$s] [%3$s] - %5$s%6$s%n
+
+# disable lock messages at startup
+org.apache.karaf.main.Main.level = WARNING
+org.apache.karaf.main.lock.SimpleFileLock.level = WARNING
--- a/distributions/distribution-resources/src/main/resources/etc/org.apache.karaf.log.cfg
+++ b/distributions/distribution-resources/src/main/resources/etc/org.apache.karaf.log.cfg
@@ -1,22 +1,3 @@
-################################################################################
-#
-#    Licensed to the Apache Software Foundation (ASF) under one or more
-#    contributor license agreements.  See the NOTICE file distributed with
-#    this work for additional information regarding copyright ownership.
-#    The ASF licenses this file to You under the Apache License, Version 2.0
-#    (the "License"); you may not use this file except in compliance with
-#    the License.  You may obtain a copy of the License at
-#
-#       http://www.apache.org/licenses/LICENSE-2.0
-#
-#    Unless required by applicable law or agreed to in writing, software
-#    distributed under the License is distributed on an "AS IS" BASIS,
-#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#    See the License for the specific language governing permissions and
-#    limitations under the License.
-#
-################################################################################
-
 #
 # This configuration file is used to configure the default values for the log:display
 # and log:exception-display commands.
--- a/distributions/distribution-resources/src/main/resources/etc/org.ops4j.pax.logging.cfg
+++ b/distributions/distribution-resources/src/main/resources/etc/org.ops4j.pax.logging.cfg
@@ -1,22 +1,3 @@
-################################################################################
-#
-#    Licensed to the Apache Software Foundation (ASF) under one or more
-#    contributor license agreements.  See the NOTICE file distributed with
-#    this work for additional information regarding copyright ownership.
-#    The ASF licenses this file to You under the Apache License, Version 2.0
-#    (the "License"); you may not use this file except in compliance with
-#    the License.  You may obtain a copy of the License at
-#
-#       http://www.apache.org/licenses/LICENSE-2.0
-#
-#    Unless required by applicable law or agreed to in writing, software
-#    distributed under the License is distributed on an "AS IS" BASIS,
-#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#    See the License for the specific language governing permissions and
-#    limitations under the License.
-#
-################################################################################
-
 # Root logger
 log4j.rootLogger = WARN, out, osgi:*
 log4j.throwableRenderer=org.apache.log4j.OsgiThrowableRenderer
--- a/distributions/distribution-resources/src/main/resources/etc/system.properties
+++ b/distributions/distribution-resources/src/main/resources/etc/system.properties
@@ -1,22 +1,3 @@
-################################################################################
-#
-#    Licensed to the Apache Software Foundation (ASF) under one or more
-#    contributor license agreements.  See the NOTICE file distributed with
-#    this work for additional information regarding copyright ownership.
-#    The ASF licenses this file to You under the Apache License, Version 2.0
-#    (the "License"); you may not use this file except in compliance with
-#    the License.  You may obtain a copy of the License at
-#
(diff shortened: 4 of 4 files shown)


distributions/distribution-resources/src/main/resources/userdata/etc/java.util.logging.properties
```diff
@@ -0,0 +1,25 @@
+################################################################################
+#
+#    Licensed to the Apache Software Foundation (ASF) under one or more
```
kaikreuzer: this license header does not really apply here, does it?

kaikreuzer:
Just for my understanding: Why is this put in userdata/etc? Is this the instance root folder, where jul looks for its configuration file?

distributions/distribution-resources/src/main/resources/userdata/etc/java.util.logging.properties
dvanherbergen: I guess not. It was copy pasted. There are a few other cfg files that have it too.

dvanherbergen:
Karaf sets the jul config file property to karaf.base/etc, which is userdata/etc.  It's not so nice, now we have 3 /etc folders in total...

distributions/distribution-resources/src/main/resources/userdata/etc/java.util.logging.properties
dvanherbergen: Shall I just remove the license for all? Or replace?

kaikreuzer:
> Karaf sets the jul config file property to karaf.base/etc

Can we reconfigure it?

dvanherbergen:
No, unfortunately it's not configurable. Moving karaf.base back to karaf.home would solve it, but I seem to remember that we have already been down that street :wink:

dvanherbergen:
Do we need to keep the runtime/etc folder? If we could merge that into runtime/karaf/etc, it would already simplify things.

dvanherbergen:
License info removed.

kaikreuzer:
> Do we need to keep the runtime/etc folder? If we could merge that into runtime/karaf/etc

I would rather remove the runtime/karaf/etc (as this is rather an internal/hidden/karaf-only folder and keep runtime/etc. Or if we expect users to do changes within this files, we might only use userdata/etc.
Btw, userdata/etc currently also has a jetty.xml. How did that get there?

distributions/distribution-resources/src/main/resources/userdata/etc/java.util.logging.properties
kaikreuzer: I think on the config files, it can be indeed removed.

dvanherbergen:
> userdata/etc currently also has a jetty.xml

Pax web generates it in the karaf.base/etc. So I guess it came back after moving karaf.base to userdata again. I can move our jetty.xml there to prevent the default pax web one from being generated.

> I would rather remove the runtime/karaf/etc 

I don't think we can do that. We can only move that one by moving the entire karaf.home. 

> if we expect users to do changes within this files

There is for example the org.ops4j.pax.logging.cfg file, which users could change (if we don't move to logback)

kaikreuzer:
> I can move our jetty.xml there to prevent the default pax web one from being generated.

Probably makes sense. Especially, as people might want to customize their jetty.xml, so userdata is a good place for it in any case.

> I don't think we can do that. We can only move that one by moving the entire karaf.home.

Really? I thought Karaf has the karaf.etc variable for redirecting it.
But anyhow, merging these two will result in problems, since ESH reads and processes files from runtime/etc in a specific way and we should not mix that with the Karaf mechanism.
Mid-term, I think we can probably completely remove the entries in runtime/etc by packaging them within the appropriate features.

maggu2810:
It seems karaf.base is used more frequently then expected (perhaps it make sense if we start using instances).
On a "clean" the karaf.data is removed (for K404 cleaned up) so if we store some configurations in karaf.base/etc we should perhaps change the directory layout a little bit.
But to fulfil all the requirements I have no idea what would the best option to fit @kaikreuzer needs.
If a clean should remove all stuff in userdata it needs to be set to karaf.data.
But perhaps we could use userdata/data and userdata/base. The clean will remove userdata/data and the base would not be touched.
If we want to clean additional stuff, we could do it in start.{sh,bat} ourself (additional).

dvanherbergen:
>  I thought Karaf has the karaf.etc variable for redirecting it.

Ah, yes. I forgot about that one.

kaikreuzer:
> to fit @kaikreuzer needs.

I hope these are not just my needs - if so, please shout at me and question those :-)

Just to quickly summarize, what I think we should have:
- conf: read-only configuration provided by the user. Not to be overwritten by any updates
- runtime: all files of the runtime that are read-only and not necessary to customize in any way by the user. On an update, this folder might be completely replaced by a new version
- userdata: the only writable folder for all data generated and modified at runtime. Should be kept untouched on an update

maggu2810:
I am fine with this three definitions, I pointed to stuff we already discussed long time ago. So no problem ATM (you remember that I started with directories named "karaf-base", "karaf-data", "karaf-home" and "karaf-etc" that are placed under runtime and userdata -- such stuff).

So I assume:
* karaf.base and karaf.data should be located "somewhere" under "userdata"
* karaf.home and karaf.etc should be located "somewhere" under "runtime"

dvanherbergen:
> runtime: all files of the runtime that are read-only and not necessary to customize in any way by the user. On an update, this folder might be completely replaced by a new version

In that case, we shouldn't we move the keystore, logback and quartz config files from runtime/etc to conf?

> userdata: the only writable folder for all data generated and modified at runtime. Should be kept untouched on an update

karaf.base is still a difficult one to place. It contains runtime info which shouldn't be updated by the user. And on an update, I believe it should be replaced. So it belongs in the runtime, but since much of its' content is generated on first startup, it requires the folder to be writable.
The standard Karaf approach for a single instance is to have karaf.home and karaf.base in the same location. I would prefer to follow this approach or to split the runtime into runtime/karaf (for karaf.home) and runtime/instances (for karaf.base), where only the runtime/karaf would then be read only.

kaikreuzer:
> we shouldn't we move the keystore, logback and quartz config files from runtime/etc to conf?

- keystore: No, rather to userdata/etc as we generate the ssl certificate through the runtime
- logback: This is not used and should be removed (see https://github.com/openhab/openhab2/issues/553)
- quartz: This is for me internal runtime stuff. Nothing to be changed or touched by the user

>  runtime/karaf (for karaf.home) and runtime/instances (for karaf.base)

This contradicts the requirements stated above that the runtime folder should be read-only. So the instance stuff has to go into userdata as we have it now.

dvanherbergen:
> keystore: No, rather to userdata/etc as we generate the ssl certificate through the runtime
logback: This is not used and should be removed (see openhab/openhab2#553)

I'll create a separate PR for this. The license info in this one was removed.

maggu2810:
So, this one could be merged and for the remaining discussion we should create a new issue (sorry, I started perhaps the discussion in this PR).

dvanherbergen:
#26 created. This one can be merged.

PR #27, opened 2016-01-06
Move keystore & jetty to new locations

### LINKED ISSUE #26: Move keystore and jetty.xml to userdata/etc
See https://github.com/openhab/openhab-distro/pull/25#issuecomment-169307600

dvanherbergen:
File shuffle to resolve #26 


maggu2810:
Am I missing something or will the clean handling be changed?
After a clean userdata/etc is not present anymore (and so the configuration files).
We could install the configuration files through a feature in userdata/etc, so it will be installed afterwards.

dvanherbergen:
Given that the jetty.certificate bundle already generates the keystore if it is absent, it might as well also install the default jetty.xml file...

kaikreuzer:
As we do not really have a use case for "clean" yet, I am ok with this change for now.

> We could install the configuration files through a feature in userdata/etc, so it will be installed afterwards.

Are config files of features automatically re-written after a clean or does this only happen when a feature is installed? Or does a clean anyhow mean that no features are installed afterwards and need to be reinstalled?

maggu2810:
> Are config files of features automatically re-written after a clean or does this only happen when a feature is installed? Or does a clean anyhow mean that no features are installed afterwards and need to be reinstalled?

The whole karaf.data is removed, so there is nothing installed (IMHO :wink:)

maggu2810:
> As we do not really have a use case for "clean" yet, I am ok with this change for now.

But then we should mask the argument in start.{sh,bat}. Or do you want that a "./start.sh clean" results in a broken installation?

kaikreuzer:
> so there is nothing installed

Not even the boot features? Then it should be fine.

> Or do you want that a "./start.sh clean" results in a broken installation?

As long as we do not claim that calling "./start.sh clean" is something people should do, I do not mind too much.

PR #35, opened 2016-01-06
added files required for launching runtime in IDE

kaikreuzer:
This is a preparation for updating the IDE setup for openHAB development.
With these files, it is possible to import everything necessary into the Eclipse workspace and launch the runtime from there without requiring any code bundles in the workspace.


dvanherbergen:
>  IDE setup for openHAB development.

Binding development or openHAB core development?
I'm not sure I understand your goal with this PR. Shouldn't this be added to openHAB2 instead?
Will binding developers really have to fork 2 repos in order to work on a new binding?

kaikreuzer:
I want to remove the distribution from openhab/openhab2, so that only the addons remain in that repo.
The "minimal" IDE setup should only check out openhab/openhab-distro and allow launching a runtime from the IDE. The user can then import whatever binding in his workspace that he likes.
A developer working on an openHAB1 binding would not at all need to check out openhab/openhab2.

dvanherbergen:
> I want to remove the distribution from openhab/openhab2, so that only the addons remain in that repo. 

Removing the distribution from openhab2 makes sense, but why not have both addons and a launch in openhab2? That way developers only need to checkout openhab2 to create new bindings. I believe that in practice, most developers will want to have other bindings in their workspace anyway to use as an example.

The distro workspace on the other hand contains things the developer doesn't need at all and in my opinion, this will only lead to confusion. A binding developer should not have to build a distro.

> A developer working on an openHAB1 binding would not at all need to check out openhab/openhab2.

openhab1 could also be trimmed to just the addons and a launch config. Or alternatively depend on openhab2 for the launch, that would still be acceptable to me as this is the future direction.

kaikreuzer:
openhab-distro will be THE place to go for developers from next Monday on. It aggregates ESH, OH1 and OH2 bindings, I will also move the documentation to it etc. There will be only one IDE setup for everyone. If you want to develop OH1 bindings, why should you have to checkout the openHAB2 repo?

>  That way developers only need to checkout openhab2 to create new bindings.

See the discussion at https://community.openhab.org/t/restructuring-of-repositories-types-of-add-ons/4277 - we might have many different repos for different addons in the future. People would only have to checkout openhab-distro and the repo with the addons they want to work on.

> openhab1 could also be trimmed to just the addons

This is what will happen for 1.9.0-SNAPSHOT. openhab/openhab will be a container for openHAB1-compatible addons, openhab/openhab2 will be a container for openHAB2-compatible addons. And we might add further repos in future.

> The distro workspace on the other hand contains things the developer doesn't need

The distro so far does not have any workspace. And the repo is the smallest we have got, since it mainly contains some xml files. But the openhab2 repo contains a lot of stuff that developers don't need: Namely all the addons.

> A binding developer should not have to build a distro.

Nobody is asked to run a Maven build on that repo.

Another reason is that I do not want to duplicate the demo resources. They are part of the openhab-distro repo and thus this repo needs to be checked out in any case.

kaikreuzer:
Thanks!
Did I convince you or did I talk you to death? ;-)

dvanherbergen:
80% convinced :-) Let's give it a go and see how it evolves..

dvanherbergen:
The openhab-distro IDE setup works flawlessly on Mac. I like the use of the two root folders 'Other Projects' and 'Infrastructure'.

PR #42, opened 2016-01-08
added fsinternetradio binding

kaikreuzer:

### DIFF
--- a/features/openhab-esh-addons/src/main/feature/feature.xml
+++ b/features/openhab-esh-addons/src/main/feature/feature.xml
@@ -7,6 +7,10 @@
         <feature>esh-binding-hue</feature>
     </feature>
 
+    <feature name="openhab-binding-fsinternetradio" description="Frontier Silicon Internet Radio Binding" version="${esh.version}">
+        <feature>esh-binding-fsinternetradio</feature>
+    </feature>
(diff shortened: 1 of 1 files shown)


dvanherbergen:
Adding this feature causes my build to fail with:
```
Unable to resolve org.eclipse.smarthome.binding.fsinternetradio/0.8.0.201601082217: missing requirement [org.eclipse.smarthome.binding.fsinternetradio/0.8.0.201601082217] osgi.wiring.package; filter:="(osgi.wiring.package=org.jupnp.model.meta)"]]
```

kaikreuzer:
Wait for this to be finished, then it will be green - but you are right, please do not merge before the ESH build is available in the maven repo.

dvanherbergen:
I did do a local smarthome build with the latest changes first. The build had a little hiccup on the automation part, but then I continued the rest of the build, so I was expecting the distro to build correctly locally. Did you do a local build that passed?

kaikreuzer:
Yes, I did.

dvanherbergen:
OK, maybe I should do a clean first. I'll try again.

dvanherbergen:
New build is available in maven repo and build works like a charm after a clean.

PR #47, opened 2016-01-09
Set management feature as not installed by default

### LINKED ISSUE #16: When doing a "logout", there is a JMX warning in the log
Using "logout" on the shell shuts down the runtime, but in the log the following line appears:
```
2016-01-04 22:47:12.313 [WARN ] [org.apache.aries.jmx.core           ] - Task rejected for JMX Notification dispatch of event [org.osgi.framework.ServiceEvent[source={javax.management.MBeanServer}={service.id=126, service.bundleid=47, service.scope=singleton}]] - Dispatcher may have been shutdown
```

Question: Can we completely remove the Aries JMX stuff or shall we simply change the logging conf to not display that?

dvanherbergen:
Fixes #16

### DIFF
--- a/distributions/openhab-offline/pom.xml
+++ b/distributions/openhab-offline/pom.xml
@@ -105,6 +105,7 @@
                 <configuration>
                     <installedFeatures>
                         <feature>wrapper</feature>
+                        <feature>management</feature>
                     </installedFeatures>
                     <startupFeatures>
                         <feature>eventadmin</feature>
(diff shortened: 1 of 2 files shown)


distributions/openhab-offline/pom.xml
```diff
@@ -105,6 +105,7 @@
                 <configuration>
                     <installedFeatures>
                         <feature>wrapper</feature>
+                        <feature>management</feature>
```
kaikreuzer: why did you add it here?

dvanherbergen: This way, the jars are included in the offline install but not activated, so that you can still do 'feature:install management' to install it offline.

kaikreuzer:
ok, works well - thanks!

PR #48, opened 2016-01-09
Set default JAVA_HOME if none is set.

### LINKED ISSUE #29: Warning if no JAVA_HOME is set
When launching the runtime without $JAVA_HOME being set, the message
```
karaf: JAVA_HOME not set; results may vary
```
is displayed. I don't think we really require JAVA_HOME anywhere and all seems to work well without it - so I wonder if we should remove this message.

dvanherbergen:
Resolves #29 

### DIFF
--- a/distributions/distribution-resources/src/main/resources/bin/setenv
+++ b/distributions/distribution-resources/src/main/resources/bin/setenv
@@ -87,3 +87,90 @@ case "$ARCH" in
       *) export EXTRA_JAVA_OPTS="-XX:+UseG1GC -Xmx256m" ;;
 esac
 
+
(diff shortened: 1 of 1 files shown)


dvanherbergen:
This PR implements @maggu2810's suggestion to set the JAVA_HOME using the code from the karaf script.

kaikreuzer:
Should this logic maybe better be moved to https://github.com/openhab/openhab-distro/blob/master/distributions/distribution-resources/src/main/resources/bin/setenv#L38?

dvanherbergen:
I've moved the logic to setenv.sh.

PR #69, opened 2016-01-14
remove trailing whitespace, split java opts

maggu2810:
I removed the trailing whitespace after <code>`arch`</code>

The other change for extra java opts has been introduced, so we could separate the common java opts and the arch specific ones.


dvanherbergen:
Looks good to me...

PR #70, opened 2016-01-14
fixed target platform location to release version

kaikreuzer:
NOTE: The old p2 repo location broke, since it did not include the version in the url - so this PR is important to be merged!

### DIFF
--- a/launch/openhab.target
+++ b/launch/openhab.target
@@ -122,8 +122,8 @@
 <repository location="http://www.jmdns.org/update-site/3.4.2/"/>
 </location>
 <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-<unit id="org.openhab.feature.p2.feature.group" version="2.0.0.a3"/>
-<repository location="https://dl.bintray.com/kaikreuzer/openhab-core/"/>
+<unit id="org.openhab.feature.p2.feature.group" version="0.0.0"/>
+<repository location="https://dl.bintray.com/kaikreuzer/openhab-core/p2/2.0.0.b1"/>
 </location>
 </locations>
 </target>


dvanherbergen:
Just to clarify, I'm not suggesting to go to pom-first builds, but to still use Tycho and manifest first, but just feed Tycho pom dependencies instead of p2 dependencies.

kaikreuzer:
> but just feed Tycho pom dependencies instead of p2 dependencies.

How is this done?

dvanherbergen:
This is described in the section [POM dependencies consider] (https://wiki.eclipse.org/Tycho/Target_Platform). From the description it should work, but I haven't actually tried it yet.

kaikreuzer:
I know the consider-pom feature, but afaik, this means that you have to add the dependencies in the pom.xmls of your bundles. So you end up doing manifest-first + pom-first at the same time or am I missing something?

kaikreuzer:
And note that the main reason for the openhab.target file is to make the IDE work (and this is also the reason, why it is at this "cyclic" location, so that it is easy to have it in the workspace) - and afaik, the consider-pom stuff simply would not work for that.

dvanherbergen:
As far as I understood it, it means we would only have to add the pom dependencies to the parent pom which contains the target-platform-configuration plugin. None of the other bundles would have to be changed, they would still be manifest first.

dvanherbergen:
> the consider-pom stuff simply would not work for that.

Agreed. But I'm not suggesting we should get rid of the openhab.target at all. Just that we should only use if for the IDE and for the maven builds use the pom dependency.

dvanherbergen:
Anyway, I'm not sure if it will even work. So I will try to do a test first...

kaikreuzer:
Ok. If it is possible to only define all dependencies in a parent pom, this might be an option for the addons-projects indeed - feel free to have a try if that works!

maggu2810:
I created such a pom sometimes for debugging.
Using Karaf console you can do: <code>bundle:list -l -t 0</code>
I have written a python script that parses that output (copy / paste) and create the dependency section of the pom.

dvanherbergen:
> I have written a python script that parses that output (copy / paste) and create the dependency section of the pom

That sounds convenient. Any chance you could run that and paste the output here? It will save me a lot of manual labour...

maggu2810:
I will have a look at this tomorrow and contact you.

dvanherbergen:
Thanks!

PR #73, opened 2016-01-15
added meta files for apt-repo distribution

teichsta:
This PR just contains the metafiles. The files still have to be adjusted to the new openHAB2 file structure which i hope @theoweiss will help with :-)

Signed-off-by: Thomas Eichstädt-Engelen <thomas@openhab.org>


kaikreuzer:
Thanks. So as this still is the "old" code, I suggest that we leave this PR unmerged for now and have @theoweiss (and others) create PRs against this PR first.

teichsta:
since the oh1 bundles build needs these metafiles we imho have the options to:

* either leave this PR unmerged and remove the jdeb plugin reference from all OH1 poms
* merge this PR to have the metadata in place and refine it with new PRs

dvanherbergen:
> since the oh1 bundles build needs these metafiles

There should be no dependency form oh1-addons (or any other openhab project) on the openhab-distro project. This will cause a circular build dependency.

kaikreuzer:
> There should be no dependency form oh1-addons (or any other openhab project) on the openhab-distro project.

I agree. So go for "remove the jdeb plugin reference from all OH1 poms" instead.

teichsta:
> There should be no dependency form oh1-addons (or any other openhab project) on the openhab-distro project. 

yes you're right … so there could be a third option to move the metafiles into the OH1 repo. Assuming that we'd like to introduce deb file generation for OH2 as well, we would than have to rethink the process.

kaikreuzer:
> we'd like to introduce deb file generation for OH2 as wel

The deb packages in openhab1 are only kept for existing 1.8 users.
For openHAB 2 (i.e. the new distro), there should imho be no packages for single addons, only one for the distro. Addons are managed through Karaf/Maven from now on.

teichsta:
will move to Add-on specific metafiles to the OH1 repo!

Issue #87, opened 2016-01-16
Add java version check in start.sh


dvanherbergen:
When starting openHAB2, it should check if the user is using at least Java 1.8. If not a clear error message should be displayed.

PR #88, opened 2016-01-16
Add java version check to setenv.

dvanherbergen:

### DIFF
--- a/distributions/distribution-resources/src/main/resources/bin/setenv
+++ b/distributions/distribution-resources/src/main/resources/bin/setenv
@@ -91,7 +91,7 @@ esac
 export EXTRA_JAVA_OPTS="${EXTRA_JAVA_OPTS_COMMON} ${EXTRA_JAVA_OPTS_ARCH} ${EXTRA_JAVA_OPTS}"
 
 
-# The 3 functions below are copied from the karaf script to set JAVA_HOME to a default value.
+# The functions below are copied from the karaf script to set JAVA_HOME to a default value.
 # This avoids Karaf printing a warning message at startup.
 
 detectOS() {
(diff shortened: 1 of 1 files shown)


dvanherbergen:
This PR adds a Java version check for the shell script. I'll create another PR for the bat file tomorrow.
Addresses #87

maggu2810:
You are using egrep, awk and sed now.
On common linux distributions all that commands (set and awk, perhaps not egrep) should be available.
Custom rootfs could miss some commands.
As Chris already runs in trouble, perhaps he could check which commands are (un)available on his system.

dvanherbergen:
It's a copy paste from the karaf script. If it doesn't work, we have bigger issues...

kaikreuzer:
>  from the karaf script. If it doesn't work, we have bigger issues...

Well, we already have big issues through Karaf (and since the vanilla Karaf does not even work on a RaspPi, I doubt that anybody ever tested it on other embedded platforms), so I agree with @maggu2810 that there is a risk that this breaks things, so we should be careful.

kaikreuzer:
@dvanherbergen Is this check not done in Karaf for Windows?

dvanherbergen:
I did not see it in the karaf.bat.

maggu2810:
For Oracle 1.7, 1.8 Java this one would work on the very most Linux systems:
<code>java -version 2>&1 | head -1 | cut -d\" -f2</code>

Kai, could you test this on MAC?

We could then using a "case ... in 1.8*) ..." (similar to the arch one).

kaikreuzer:
Works on Mac as well.

maggu2810:
Why still using egrep, awk, ...? If we could rely on head and cut only? Do you think there are differences on OpenJDK?

Issue #18, opened 2016-01-05
Karaf 4.0.8 should be tested


maggu2810:
``` text
Jean-Baptiste Onofré, 08:48
...
Today I'm fixing the latest Jira for 4.0.4, I should be able to submit 4.0.4 release to vote tomorrow
```

We should test K404 as soon as its voting round starts and we should vote +1 (if all is fine) or -1 (is something is broked) for the release.
##

maggu2810:
FYI: the vote has been started
http://karaf.922171.n3.nabble.com/VOTE-Apache-Karaf-4-0-4-release-tp4044726.html
After the release freeze we should test this immediately to guarantee that we have no regressions using K404 for the next developmemt round.

maggu2810:
K404 has been released:
http://karaf.922171.n3.nabble.com/ANN-Apache-Karaf-4-0-4-released-tp4044829.html

kaikreuzer:
I just had a first try (see https://github.com/kaikreuzer/openhab-distro/tree/karaf404)
Build works fine, but the offline-distro does not start:
```
2016-01-16 01:07:00.038 [WARN ] [url.mvn.internal.AetherBasedResolver] - Error resolving artifactorg.ops4j.pax.web:pax-web-jetty:jar:4.2.4:Could not find artifact org.ops4j.pax.web:pax-web-jetty:jar:4.2.4
shaded.org.eclipse.aether.resolution.ArtifactResolutionException: Could not find artifact org.ops4j.pax.web:pax-web-jetty:jar:4.2.4
        at shaded.org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:444)[4:org.ops4j.pax.url.mvn:2.4.5]
        at shaded.org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifacts(DefaultArtifactResolver.java:246)[4:org.ops4j.pax.url.mvn:2.4.5]
        at shaded.org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifact(DefaultArtifactResolver.java:223)[4:org.ops4j.pax.url.mvn:2.4.5]
        at shaded.org.eclipse.aether.internal.impl.DefaultRepositorySystem.resolveArtifact(DefaultRepositorySystem.java:294)[4:org.ops4j.pax.url.mvn:2.4.5]
```
It seems as if we need to update some other versions with it.
@dvanherbergen Can you help on that?

[...]

PR #89, opened 2016-01-17
[build fix] karaf 4.0.4-snapshot no longer available. 

dvanherbergen:

### DIFF
--- a/features/openhab-aggregate-kar/pom.xml
+++ b/features/openhab-aggregate-kar/pom.xml
@@ -30,7 +30,7 @@
             <plugin>
                 <groupId>org.apache.karaf.tooling</groupId>
                 <artifactId>karaf-maven-plugin</artifactId>
-                <version>4.0.4-SNAPSHOT</version>
+                <version>${dep.karaf.version}</version>
                 <extensions>true</extensions>
                 <configuration>
                     <startLevel>80</startLevel>


maggu2810:
Have you bumped the dep.karaf.version?
You know that we are using > 4.0.3 because 4.0.3 will not work for some snapshots?

dvanherbergen:
> Have you bumped the dep.karaf.version?

No. That is still on 4.0.3. The only one using 4.0.4-snapshot was the openhab-aggregate-kar.
Before we can move everything to 4.0.4, https://github.com/openhab/openhab-distro/issues/18#issuecomment-172134178 needs to be fixed.

> Have you considered that we are using > 4.0.3 because 4.0.3 will not work for some snapshots?

I've tested the distro build with this change to 4.0.3 and that works OK. Which snapshot are you referring too?

maggu2810:
The KAR creation for snapshots artifacts could be broken in <= 4.0.3. That is the reason why we have used > 4.0.3 (4.0.4-SNAPSHOT) for the KAR creation only.
https://github.com/apache/karaf/pull/102

dvanherbergen:
OK, so should we bump the version for the KAR creation or try to bump everything?

maggu2810:
We could bump the KAR creation to 4.0.4 as a fix (keep a TODO that we should remove the explicit version as soon as >= 4.0.4 is used for the whole repo).

The bump of the whole distribution could be more work and should be done in a separate PR. As I mentioned earlier, all the changed configurations need to be checked. If the unchanged / upstream configuration file contains more / less stuff now, we have to "merge" that change to our modified one, too.

I asked JBO to include the mechanism to edit configuration files by the build process some time ago and it was able to include that mechanism in K404.
See: https://github.com/apache/karaf/pull/80
Using this feature we could perhaps drop some hard coded files.

PR #90, opened 2016-01-17
Update arch to uname

### LINKED ISSUE #86: arch command used in setenv not supported on synology
The check for ARM should be updated to not fail on Synology, see:

See https://community.openhab.org/t/oh2-port-conflict-karaf/6339/8

dvanherbergen:
Resolves #86 


maggu2810:
I would assume uname is available on all systems. Should we use arch as fallback if uname is not present?

dvanherbergen:
I would assume that uname would be also there if arch is there. We could add a check for the existence of /usr/bin/uname.

PR #95, opened 2016-01-18
added task to install compat layer when testing openHAB 1 add-ons

kaikreuzer:

### DIFF
--- a/docs/sources/features/compatibilitylayer.md
+++ b/docs/sources/features/compatibilitylayer.md
@@ -9,9 +9,11 @@ Currently, the compatibility layer focuses on the official APIs, i.e. an add-ons
 ## How to use openHAB 1.x Add-ons that are not part of the distribution
  
 While the openHAB distribution already contains many add-ons of openHAB 1, there are still quite some of them missing - please help testing them - if they are confirmed to be working, they can be included in the distribution.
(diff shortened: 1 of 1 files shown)


maggu2810:
You can remove one space between "-" and "A" (<code>+ -  As with</code>).

kaikreuzer:
fixed.

lolodomo:
@kaikreuzer : JUst one little question: why the compat layer is not a bundle installed by default ?

kaikreuzer:
Why should it? If someone does not use any 1.x add-ons there is no need for it.
And once an add-on is added to the distro (see step 6 on https://github.com/openhab/openhab-distro/blob/master/docs/sources/features/compatibilitylayer.md), the compat layer is automatically installed when installing the add-on (so the user does not have to do that).

lolodomo:
If automatic, why mentioning to do it manually ?

kaikreuzer:
Because this is the documentation for developers on "How to use openHAB 1.x Add-ons that are not part of the distribution"

maggu2810:
@lolodomo If you install addons / bundles using features / extension management, the compat layer will be installed if necessary. As this dependencies are defined.
If you install bundles using the addons directory / file deployment, you have to ensure yourself that dependencies will be installed.

lolodomo:
That is clearer with this last explanation.
Thank you.

PR #104, opened 2016-01-22
Update openhab.target to openhab-deps-repo

dvanherbergen:

### DIFF
--- a/launch/openhab.target
+++ b/launch/openhab.target
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<?pde version="3.8"?><target name="openHAB Target Platform" sequenceNumber="146">
+<?pde version="3.8"?><target name="openHAB Target Platform" sequenceNumber="149">
 <locations>
 <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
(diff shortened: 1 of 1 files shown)


dvanherbergen:
I did some initial testing on this and it seems to work OK, but it would be great if someone could test this more.

maggu2810:
I checked out this PR, started IDE, ensure the new target platform is set and did a clean and rebuild all.
Seems to be working.

kaikreuzer:
openhab-core seems to be working in the target platform. This means that you cannot start the runtime within the IDE anymore!

dvanherbergen:
Not sure how I missed that. My IDE runtime seemed to work OK. Anyway, I've added the openhab-core again.

kaikreuzer:
Looks much better now!
One last question: Why do we still need the luna update site? Can't these dependencies be included in the openhab-deps-repo? Or is this on purpuse, because it only adds the Equinox framework for launch from the IDE and the Maven compilation obviously does not require it?

dvanherbergen:
> the Maven compilation obviously does not require it

Indeed, I started from the builds and tried to keep the dependencies as small as possible to get the best performance in the dependency resolution. 

Currently the openhab-deps-repo already includes two features: one for the compile dependencies and one for tests. We could add a third feature for the IDE runtime. I'm not sure if it will add much value other than keeping the target file a little cleaner.

kaikreuzer:
Ok, makes sense then. Thanks!

Issue #68, opened 2016-01-13
UPnP stream port needs to be changed if Jetty port is changed


kaikreuzer:
The port for upnp callbacks is defined in https://github.com/openhab/openhab-distro/blob/master/distributions/distribution-resources/src/main/resources/runtime/etc/services.cfg#L23
and jUPnP registers a servlet on Jetty to process those incoming requests.
If the user changed the Jetty port to something different than 8080, this will break.
We need to find a way to feed in the current jetty port to jupnp. Maybe this requires a PR for jUPnP itself.

kaikreuzer:
Addressed by https://github.com/jupnp/jupnp/pull/46

kaikreuzer:
Once a new jUPnP version is available, we should remove the line mentioned above from services.cfg.

dvanherbergen:
Closed via #110

PR #110, opened 2016-01-30
updated to JUPnP 2.1.0.RC1, which requires new config settings

kaikreuzer:

### DIFF
--- a/launch/home/etc/services.cfg
+++ b/launch/home/etc/services.cfg
@@ -20,8 +20,8 @@ org.openhab.autoapprove:enabled=false
 org.eclipse.smarthome.binding.sonos:opmlPartnerID=IAeIhU42
 
 org.jupnp:multicastResponsePort=0
-org.jupnp:streamListenPort=8080
-org.jupnp:threadPoolSize=200
+org.jupnp:asyncThreadPoolSize=30
+org.jupnp:threadPoolSize=15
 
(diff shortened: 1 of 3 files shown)


kaikreuzer:
fixes https://github.com/openhab/openhab-distro/issues/68

launch/home/etc/services.cfg
```diff
@@ -20,8 +20,8 @@ org.openhab.autoapprove:enabled=false
 org.eclipse.smarthome.binding.sonos:opmlPartnerID=IAeIhU42
 
 org.jupnp:multicastResponsePort=0
-org.jupnp:streamListenPort=8080
-org.jupnp:threadPoolSize=200
+org.jupnp:asyncThreadPoolSize=30
+org.jupnp:threadPoolSize=15
```
dvanherbergen: I'm not an expert on how upnp works, but 45 threads in total, isn't that a bit much?

kaikreuzer: No, it is a pretty good setting.
Note that up to now, we had 200 threads that were always started at startup. Now we have pools only temporarily increase if necessary, up to a maximum of 45 threads. (and when running a upnp discovery, you can easily have 500-1000 incoming requests to process)

dvanherbergen: Thanks for the explanation.

Issue #109, opened 2016-01-30
Provide generic OPENHAB_JAVA_OPTS env variable


dvanherbergen:
[Currently](https://community.openhab.org/t/solved-how-to-define-a-serial-port-on-raspberry-for-openhab2-for-my-razberry-zwave-binding/6693/5) you have to edit the setenv.sh to add additional java options. 
To make this easier, it would be nice to have an OPENHAB_JAVA_OPTS env variable, which can be set by the user and which is appended to the java options to start openhab.

maggu2810:
See also: https://github.com/openhab/openhab-distro/pull/111#issuecomment-177456732

kaikreuzer:
Solved through documentation update in #111

PR #111, opened 2016-01-31
Describe additonal JVM config option

dvanherbergen:
Makes #109 unnecessary. 

### DIFF
--- a/docs/sources/getting-started.md
+++ b/docs/sources/getting-started.md
@@ -73,6 +73,16 @@ export OPENHAB_HTTP_PORT=8011
 export OPENHAB_HTTPS_PORT=8444
 ```
 
+## Configuring additional JVM options
+
+In a standard setup, you don't need to configure JVM arguments. There may however be cases where bindings require additional JVM settings, for example to use non standard serial ports.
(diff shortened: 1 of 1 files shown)


docs/sources/getting-started.md
```diff
@@ -73,6 +73,15 @@ export OPENHAB_HTTP_PORT=8011
 export OPENHAB_HTTPS_PORT=8444
 ```
 
+## Configuring additional JVM options
```
kaikreuzer: Maybe we should directly refer to the cases, where this is needed (we are here on a "getting-started"), which is specifically defining a non-standard serial port.

dvanherbergen: Text updated to make it more clear it is not needed by default.

maggu2810:
Instead of using EXTRA_JAVA_OPTS or OPENHAB_JAVA_OPTS I would prefer to edit the system.properties as it is already done for other declarations:
https://github.com/openhab/openhab-distro/blob/66fc97431479c7b7c8a4ce635bf2f823da976107/distributions/distribution-resources/src/main/resources/etc/system.properties#L137

dvanherbergen:
As far as I understood, users shouldn't edit anything in the runtime folder because it would be replaced when they install a new version of openHAB.

PR #115, opened 2016-02-02
Fixes #114. Add fancier splash screen based on simple ascii art.

### LINKED ISSUE #114: Brand openhab console a bit more
Karaf provides capabilities to customize welcome screen when launching instance/getting there over ssh. Its very simple addon which clearly marks what kind of software user is running.

splatch:
This pull request provides a simple customization for shell console. 
openhab splash

### DIFF
--- a/distributions/distribution-resources/src/main/filtered-resources/etc/branding.properties
+++ b/distributions/distribution-resources/src/main/filtered-resources/etc/branding.properties
@@ -1,8 +1,13 @@
+# if you are looking here you might be interested in visiting this page:
+# http://www.network-science.de/ascii/ - this header is made with slant font. Enjoy!
 welcome = \r\n\
-\u001B[36m\u001B[1m-------------------------------------------------------------------------\r\n\
-Welcome to openHAB ${project.version}\r\n\
--------------------------------------------------------------------------\r\n\
-\u001B[0m\r\n\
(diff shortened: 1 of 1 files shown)


maggu2810:
Looks very nice.
Have you removed the trailing whitespace after 'openhab&gt;' by intention?
<code>openhab&gt;bundle:list</code> vs. <code>openhab&gt; bundle:list</code>

kaikreuzer:
Nice! I do not like the red dots in front of "empowering" too much (makes it too many different colors for my taste). How about making them grey (as the text) or remove the line completely?

splatch:
@maggu2810 yeah, my editor trimmed it out. WIll bring it back.
@kaikreuzer I could not find matching colors for bars from one hand and tried to refer slogan from logo. I'm fine to drop this line. These bars might not be visible when someone is using red background (I doubt if anyone does), also there is no way to make background with default color used in terminal, thus it might be tricky.

The same splash on white background
openhab splash white

With red background
openhab splash red

With silver background
openhab splash silver

kaikreuzer:
> tried to refer slogan from logo.

Nice idea, but I think visually it is cleaner if we drop it.

splatch:
Ammended commit. Below you can find how it looks without slogan text:

black
blue
green
navy
red
sad
silver
white
yellow

kaikreuzer:
Cool - thanks!

maggu2810:
@splatch :+1:

PR #147, opened 2016-02-25
Bump karaf to version 4.0.4

### LINKED ISSUE #18: Karaf 4.0.8 should be tested

dvanherbergen:
Resolves #18.

### DIFF
--- a/distributions/openhab-offline/pom.xml
+++ b/distributions/openhab-offline/pom.xml
@@ -47,7 +47,7 @@
         <dependency>
           	<groupId>org.openhab.distro.deps</groupId>
   			<artifactId>pax-web-jetty</artifactId>
-  			<version>4.2.3.sp1</version>
+  			<version>4.2.4</version>
 			<scope>provided</scope>
   		</dependency>
     </dependencies>
--- a/distributions/openhab-offline/src/main/descriptors/archive.xml
+++ b/distributions/openhab-offline/src/main/descriptors/archive.xml
@@ -97,9 +97,9 @@
         </file>      
         <!-- add patch for https://github.com/openhab/openhab-distro/issues/10 -->
         <file>
-            <source>target/assembly/system/org/openhab/distro/deps/pax-web-jetty/4.2.3.sp1/pax-web-jetty-4.2.3.sp1.jar</source>
-            <outputDirectory>runtime/karaf/system/org/ops4j/pax/web/pax-web-jetty/4.2.3/</outputDirectory>
-            <destName>pax-web-jetty-4.2.3.jar</destName>
+            <source>target/assembly/system/org/openhab/distro/deps/pax-web-jetty/4.2.4/pax-web-jetty-4.2.4.jar</source>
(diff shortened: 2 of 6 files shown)


dvanherbergen:
I only did a quick test of the offline distro, but the dashboard and basic ui seem to be working fine. I didn't check for any other issues that 4.0.4 could cause.

kaikreuzer:
I have some odd behavior of the shell. First of all it starts twice with the "openHAB" welcome screen.
And then it is really sluggish when I type commands and even if I manage to type them in, they are not getting in time to the interpreter:
```
openhab> logout
Command not found: lgou
openhab> feature:list
Command not found: otfurlst
openhab> 
```
Do you have the same?

dvanherbergen:
I didn't notice the duplicate logo before.. but yes, I get the same behaviour. I'm wondering if something started twice..

dvanherbergen:
It looks like something is started twice. After changing my config from demo to standard and restarting, it only showed one logo for a second while it was reconfiguring. Then the second logo appeared.

dvanherbergen:
The problem with the console seems like a regression issue: https://community.openhab.org/t/osgi-console-not-using-all-text-input/5282

kaikreuzer:
Hm, interesting, the thread above is from pre-karaf, so it might be some general issue with the gogo shell.

kaikreuzer:
@dvanherbergen Any idea about how to fix the "double-intro" bug?

dvanherbergen:
When openHAB starts, you can see that it launches the console twice, there are 2 threads 'Karaf shell pipe' and 2 threads 'Karaf local console user openhab'.
If you then connect using the karaf/bin/client, then a third one for each is added. The client console itself works as expected. So for some reason it is starting these threads twice on regular startup. No idea why it would do this though...

kaikreuzer:
If there are 2 shells simultaneously, it also explains why the keyboard entries are not correctly processed. I hope you find out what is wrong there!

maggu2810:
K405 is on vote and contains e.g. this bugfix: https://issues.apache.org/jira/browse/KARAF-4072

So we should try to use K405 (perhaps) and not bump to K404 anymore.

Perhaps we need to use K404 for the plugin and K405 for the distribution, but let's wait for the final decision (see: http://karaf.922171.n3.nabble.com/VOTE-Apache-Karaf-4-0-5-release-tp4046085p4046126.html).

kaikreuzer:
Superseded by https://github.com/openhab/openhab-distro/pull/172

PR #188, opened 2016-04-25
removed tmp-shk maven repo for online distro (as all required content…

kaikreuzer:
… should be part of the pre-installed features)

@maggu2810 Can you confirm that my assumption is right that this repo is not relevant for any add-ons and is only required to construct the core distributions?


maggu2810:
I think so

kaikreuzer:
So ok to merge? Or shall I do it myself?

PR #211, opened 2016-05-17
Update services.cfg

kgoderis:
Inclusion of opml URL definition as not done/omitted in ESH

Signed-off-by: Karel Goderis <karel.goderis@me.com>


kaikreuzer:
Thanks. Actually, we should remove it from services.cfg and add it right in here: https://github.com/kaikreuzer/openhab-core/blob/master/features/openhab-esh-addons/src/main/feature/feature.xml#L27, that would be even better.

maggu2810:
Why not here https://github.com/eclipse/smarthome/blob/4ab38d71/features/karaf/src/main/feature/feature.xml#L125

maggu2810:
Is this openHAB specific. What is the URL used for?

kaikreuzer:
Because we were not allowed to refer to this url within ESH. That's the whole point why we designed it as a configuration property.

PR #213, opened 2016-05-23
deb files from maven build

theoweiss:
Signed-off-by: Theo Weiss <theo@m1theo.org> (github: theoweiss)

This is the successor of https://github.com/openhab/openhab-distro/pull/138 with squashed commits and Signed-off-by.


kaikreuzer:
Many thanks - let's merge and test this!

theoweiss:
For the bintray repo I think the easiest way would be to use a new repository and adapt the openhab1 shell scripts https://github.com/openhab/openhab/blob/1.8/distribution/src/deb/bintray-upload-debs.sh and https://github.com/openhab/openhab/blob/1.8/distribution/src/deb/bintray-upload-fake-packages.sh . What do you think?

kaikreuzer:
Don't ask me, you are the expert 👍

PR #215, opened 2016-05-26
MapDB, Prowl, and Twitter seem to be working in OpenHAB 2

mannkind:
Currently MapDB can be used via addons.cfg; Prowl and Twitter must be installed manually.

With the action addons, I tested… 
Prowl: pushNotification(String subject, String message), pushNotification(String apiKey, String subject, String message), pushNotification(String apiKey, String subject, String message, int priority)
Twitter: sendTweet(String tweetTxt)


kaikreuzer:
Thanks! I have added the actions with https://github.com/openhab/openhab/pull/4420.

The list of 1.x addons is now generated and resides at http://docs.openhab.org/addons/1xaddons.html, so I close this PR.

PR #216, opened 2016-05-26
Add hint how to connect to the openHAB shell using ssh in getting-started.md

joleuger:
Sometimes you want access to the openHAB shell even after registering openHAB as a system service. This shows novice users how.


kaikreuzer:
Thanks!

@splatch Didn't you want to create a PR that changes user/pwd from karaf to openhab?

splatch:
@kaikreuzer Yes I did, however Karaf 4 with no aries/blueprint makes it harder to add new realms. We can just change default username and password, however JAAS realm name will stay the same which may be missleading.

kaikreuzer:
@splatch Would be fine for me, I don't think many people get in direct contact with the JAAS realm name...

PR #240, opened 2016-06-27
added further config options

kaikreuzer:


maggu2810:
Are you using the non-default values by intention (auto approve and non-auto linkage)?

kaikreuzer:
Yes - because they are commented out. So the people that want to change to a non-default behavior only have to remove the comment on the line and do not have to wonder what value they have to fill in.

maggu2810:
> Yes - because they are commented out.

:blush: I am blind.

PR #271, opened 2016-09-02
Bring service description in sync with other services

ThomDietrich:
Also, add documentation links
(https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Documentation=)

Signed-off-by: Thomas Dietrich <Thomas.Dietrich@tu-ilmenau.de>

### DIFF
--- a/distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
+++ b/distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=openHAB 2 - empowering the smart home
-Documentation=https://docs.openhab.org
+Documentation=http://docs.openhab.org
 Documentation=https://community.openhab.org
 Wants=network-online.target
 After=network-online.target
--- a/distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
+++ b/distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
@@ -1,6 +1,7 @@
 [Unit]
-Description=Starts and stops the openHAB Home Automation Bus
-Documentation=http://www.openhab.org
+Description=openHAB 2 Home Automation Bus Solution
+Documentation=http://docs.openhab.org
+Documentation=http://community.openhab.org
 Wants=network-online.target
 After=network-online.target
 
--- a/distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
+++ b/distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
@@ -1,7 +1,7 @@
 [Unit]
-Description=openHAB 2 Home Automation Bus Solution
-Documentation=http://docs.openhab.org
-Documentation=http://community.openhab.org
+Description=openHAB 2 - empowering the smart home
+Documentation=https://docs.openhab.org
+Documentation=https://community.openhab.org
 Wants=network-online.target
 After=network-online.target
 


distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
-Description=Starts and stops the openHAB Home Automation Bus
-Documentation=http://www.openhab.org
+Description=openHAB 2 Home Automation Bus Solution
+Documentation=http://docs.openhab.org
```
kaikreuzer: "A space-separated list of URIs" - there should hence only be a single "Documentation=" entry here.
Btw, I prefer https://community.openhab.org/ (note the https), since the forum is SSL-enabled.

distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
-Description=Starts and stops the openHAB Home Automation Bus
-Documentation=http://www.openhab.org
+Description=openHAB 2 Home Automation Bus Solution
```
kaikreuzer: Do we need the "Solution"?
How about "openHAB 2 - empowering the smart home"?

ThomDietrich: This actually was inspired by you, telling me at some point that you like to call openHAB a "solution" :) I'm okay with your suggestion, it's the one used on the webpages. I feel however like this should follow the rule of uppercase in headlines
"openHAB 2 - Empowering the Smart Home" or
"openHAB 2 - Empowering the SmartHome"
 Yay or Nay?

distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
ThomDietrich: "A space-separated list of URIs... This option may be specified more than once, in which case the specified list of URIs is merged."
Other famous services like samba follow this strategy.

distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
kaikreuzer: In general yes, but here I would say it is a fixed term as this is [how it appears in the logo](http://www.openhab.org/assets/images/openhab-logo-top.png).

distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
kaikreuzer: ok, thanks, I didn't see that this s also possible.

kaikreuzer: But I still prefer https :-)

distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
ThomDietrich: Okay I'll adhere to it. Maybe you can consider this in the future.

distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
ThomDietrich: Me too ;)

ThomDietrich:
As discussed. Ready to merge

distributions/distribution-resources/src/main/resources/deb/systemd/openhab2.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
-Description=Starts and stops the openHAB Home Automation Bus
-Documentation=http://www.openhab.org
+Description=openHAB 2 - empowering the smart home
+Documentation=https://docs.openhab.org
```
kaikreuzer: Err, only the community is https, docs is hosted on github pages, where we cannot do ssl...

ThomDietrich: I'm able to reach https://docs.openhab.org but I just saw that it seems like I have added an exception to my browser to allow the mismatched certificate. fixing now.

PR #272, opened 2016-09-02
Add git files and vim modeline to demo resources

ThomDietrich:
Regarding the demo package: The following additions make tracking of a users configuration in a git/GitHub repository easier.

**.gitattributes**
* this file will tell GitHub which files have to be counted as which language in statistics: https://github.com/github/linguist#using-gitattributes

**.gitignore**
* providing a default gitignore with commented examples

**vim modeline**
* in accordance with https://github.com/github/linguist#using-emacs-or-vim-modelines
* By adding `// vim: syntax=Xtend` at the end of each file, GitHub is able to syntax highlight openHAB files.
* Compare [before](https://github.com/openhab/openhab-distro/blob/master/features/openhab-demo-resources/src/main/resources/rules/demo.rules) and [after](https://github.com/ThomDietrich/openhab-distro/blob/demo-syntax-hl/features/openhab-demo-resources/src/main/resources/rules/demo.rules)
* The extension `// vim: syntax=Xtend syntax=openhab` will additionally make this compatible with https://github.com/cyberkov/openhab-vim

### DIFF
--- a/features/openhab-demo-resources/src/main/resources/.gitattributes
+++ b/features/openhab-demo-resources/src/main/resources/.gitattributes
@@ -1,0 +1,6 @@
+*.items linguist-language=Xtend
+*.persist linguist-language=Xtend
+*.sitemap linguist-language=Xtend
+*.things linguist-language=Xtend
+*.map linguist-language=Xtend
+*.cfg linguist-language=INI
--- a/features/openhab-demo-resources/src/main/resources/.gitignore
+++ b/features/openhab-demo-resources/src/main/resources/.gitignore
@@ -1,0 +1,6 @@
+# This file is only relevant, if you decide to track your configuration in a git repository.
+# If you are going to upload your configuration to GitHub or another public place, be sure to
+# hide your sensitive data, examples may be:
+
+#services/pushover.cfg  # may contain your login credentials
+#things/astro.things    # may contain your physical location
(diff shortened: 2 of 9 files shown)


kaikreuzer:
> The extension // vim: syntax=Xtend syntax=openhab will additionally make this compatible with https://github.com/cyberkov/openhab-vim

Shall we move these files directly somewhere to the distro and add this option in the documentation?

ThomDietrich:
Thanks for merging! can you rephrase your question?

kaikreuzer:
What I meant is that https://github.com/cyberkov/openhab-vim mainly contains two simple files, which people can copy to their .vim folder. I wondered if we should put these files in the openhab-distro repo (and potentially even package them in the distro zip as well (although I do not have a good suggestion for the folder where to put them).

ThomDietrich:
It's arguable if patching vim is a task of the openHAB setup. There is the exact same file ready for nano, would we have to add this one too? I have btw. the open issue https://github.com/cyberkov/openhab-vim/issues/2 asking the developer to add "Xtend" as a syntax, which brings us to the next question: Should vim be patched for openHAB or for Xtend in the long run?

Both vim and nano syntax highlightings are by the way now also part of openHABian.

If you want to continue this discussion, let's create an issue for it.

kaikreuzer:
Ok, created https://github.com/cyberkov/openhab-vim/issues/3 for further discussion.

ThomDietrich:
The vim modeline part of this PR is now broken.
GitHub must have changed something in it's highlighting engine. `// vim: syntax=Xtend syntax=openhab` is not valid anymore, `// vim: syntax=Xtend` is working but is in conflict with "openhab-vim". Discussed here: https://github.com/cyberkov/openhab-vim/issues/2

PR #294, opened 2016-10-11
Add description to conf/html/index.html

ThomDietrich:
Minor detail. Describing the purpose of the static subpath. Adapt the phrasing as you like.

https://community.openhab.org/t/migration-from-oh-1-x-to-oh-2-tutorial/13947/45

### DIFF
--- a/distributions/distribution-resources/src/main/resources/conf/html/index.html
+++ b/distributions/distribution-resources/src/main/resources/conf/html/index.html
@@ -1,3 +1,5 @@
 <html>
-<h1>Hello World!</h1>
+<h1>openHAB user provided static content</h1>
+<p>Serve your own static html pages or resources from here. Files stored in this folder will be available through the HTTP server of openHAB, e.g. <a href="http://device-address:8080/static/image.png">http://device-address:8080/static/image.png</a>.</p>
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Thanks!

ThomDietrich:
2016-10-14 16_44_53-ssh pi 192 168 11 66

Oh, I didn't anticipate that... But I guess that's default behavior.

kaikreuzer:
This is the apt update, right? @theoweiss can this message be avoided somehow?

ThomDietrich:
Exactly.
I'm unsure however if we need to handle this in any way. In the example above my systems index.html was replaced by the new file as it should be and the new readme.txt was proposed to me as I did not have it anymore. So I *guess* apt is doing the right thing.

BClark09:
On the other hand, this would only happen if an example file is updated, and if the user has changed it (or removed the readme) then it would be safe to assume that they already know it's purpose? The halt caught me surprise because I usually use a no prompt option to update automatically.

`-o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"` options for apt-get should remove this prompt.

Issue #301, opened 2016-10-17
Installing openhab2 snapshot from repository onraspberry pi fails


mihalski:
Hi,

I am having a look at openhab due to it's zwave support and homebridge bindings (thus allowing motion sensor integration with homekit). Having followed the instructions on installing openhab2 on a raspbeery pi I encountered the following error:

```
pi@homepi:~ $ sudo apt-get install openhab2-online
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following NEW packages will be installed:
  openhab2-online
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 0 B/48.7 MB of archives.
After this operation, 54.8 MB of additional disk space will be used.
(Reading database ... 37653 files and directories currently installed.)
Preparing to unpack .../openhab2-online_2.0.0~20161017033125_all.deb ...
dpkg: error processing archive /var/cache/apt/archives/openhab2-online_2.0.0~20161017033125_all.deb (--unpack):
 subprocess new pre-installation script returned error exit status 1
Errors were encountered while processing:
 /var/cache/apt/archives/openhab2-online_2.0.0~20161017033125_all.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

It appears that there is a problem with the pre-installation script, but I am unable to identify it. Here is what is currently in the script:

```
#!/bin/sh

set -e

UD_CACHE=/var/lib/openhab2/cache
UD_TMP=/var/lib/openhab2/tmp

removeCache(){
    [ -d ${UD_CACHE} ] && rm -rf ${UD_CACHE}
    [ -d ${UD_TMP} ] && rm -rf ${UD_TMP}
}

case "$1" in
        install)
        removeCache
                OH_USER=openhab
                OH_GROUP=openhab
                if [ x"${USER_AND_GROUP}" != x ]; then
                        OH_USER=`echo ${USER_AND_GROUP} | cut -d ":" -f 1`
                        OH_GROUP=`echo ${USER_AND_GROUP} | cut -d ":" -f 2`
                fi
                if ! getent group "$OH_GROUP" > /dev/null 2>&1 ; then
                        addgroup --system "$OH_GROUP" --quiet
                fi
                if ! getent passwd "$OH_USER" > /dev/null 2>&1 ; then
                        adduser --quiet --system --ingroup "$OH_GROUP" --no-create-home \
                        --disabled-password --shell /bin/false \
[...]

PR #286, opened 2016-09-28
apt upgrade fixes

theoweiss:
should fix #284 and #283
Merge should wait for feedback from @ollie-dev .

PR #302, opened 2016-10-17
ensure removeCache returns true

theoweiss:
This is the fix for #301 which is a regression introduced with #286.
New installations will fail because "set -e" is enabled and /var/lib/openhab2/tmp don't exist. Upgrades are not affected.
It will be good to have this merged soon because new snapshot installations are impossible.

### DIFF
--- a/distributions/distribution-resources/src/main/resources/deb/control-runtime/postrm
+++ b/distributions/distribution-resources/src/main/resources/deb/control-runtime/postrm
@@ -16,6 +16,7 @@ removeOpenHABInit() {
 removeCache(){
     rm -rf /var/lib/openhab2/cache
     rm -rf /var/lib/openhab2/tmp
+	true
 }
 
 case "$1" in
--- a/distributions/distribution-resources/src/main/resources/deb/control-runtime/preinst
+++ b/distributions/distribution-resources/src/main/resources/deb/control-runtime/preinst
@@ -8,6 +8,7 @@ UD_TMP=/var/lib/openhab2/tmp
 removeCache(){
     [ -d ${UD_CACHE} ] && rm -rf ${UD_CACHE}
     [ -d ${UD_TMP} ] && rm -rf ${UD_TMP}
+	true
 }
 
 case "$1" in


peripal:
Yeah. Its a biggie. :) Quick work tho!

ThomDietrich:
lgtm... I'm not a big fan of the extra lines though. jm2c:

```
rm -rf /var/lib/openhab2/cache || true
rm -rf /var/lib/openhab2/tmp || true
```
Reads "delete, otherwise return true"

ThomDietrich:
@kaikreuzer this is critical. installations are failing right now...

kaikreuzer:
will merge as soon as @theoweiss has updated the PR.

kaikreuzer:
or shall I merge right away and @theoweiss can do a follow up PR?

maggu2810:
I don't get the point what is exactly the problem.
Shouldn't the usage of "-f" ignore non-existing stuff?
For me it succeeded.

```bash
$ rm -rf non-existing-dirent; echo $?
0
```

But IMHO Ignore failures on deletion is not safe.
Why not check for existence and remove if it exists without ignoring failures?

maggu2810:
Okay, got it.
It is the test that results in the error code.

`[ -d ${...} ]` caused the error

maggu2810:
You should use something similar to that one:
```text
removeCache() {
	if [ -e "${UD_CACHE}" ]; then
		rm -r "${UD_CACHE}"
	fi
	if [ -e "${UD_TMP}" ]; then
		rm -r "${UD_TMP}"
	fi
}
```

First think about if you care if the directory entry is a file or a directory or something similar.
If you would like to fail if it is a file and not a directory, you have to change the test.
If you don't care about the type, you can only check for existence.

You don't also need to ignore non-existing stuff because you check for existence before.

The usage of && respect the exit code of the test and if that signals "failure" the result is failure.
You should use a explicit test that evaluates the expression.

ThomDietrich:
@maggu2810 you are totally right. My suggestion was not fully thought through.
@kaikreuzer I would opt for "merge now and create another PR later"

kaikreuzer:
ok, merged!

ThomDietrich:
Soooo is somebody interested in reworking this?

PR #317, opened 2016-11-05
Demo dynamic image URLs

watou:
See https://github.com/eclipse/smarthome/pull/2412

Signed-off-by: John Cocula <john@cocula.com>

### DIFF
--- a/features/openhab-demo-resources/src/main/resources/items/demo.items
+++ b/features/openhab-demo-resources/src/main/resources/items/demo.items
@@ -121,5 +121,6 @@ Dimmer Volume                   "Volume [%.1f %%]"
 Number Temperature_Setpoint     "Temperature [%.1f °C]" <temperature>
 Location DemoLocation           "Brandenburg Gate Berlin"
 Number Wifi_Level               "Wifi Level [%d/4]"     <signal>        (FF_Office)
+String ImageURL                 "Image URL [%s]"
 
 // vim: syntax=Xtend syntax=openhab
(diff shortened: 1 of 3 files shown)


ThomDietrich:
Funny URLs :) But do we really need all 8 of them in the demo file?
Also according to the referenced PR the url parameter serves as a fallback image. Didn't get clear from this example...

Now that you already experimented with this new feature, would you add another PR here?
https://github.com/openhab/openhab-docs/blob/gh-pages/features/sitemap.md

watou:
Thanks for your input.  I've shortened the list a bit.  I don't know a good way to make the default URL feature more prominent in the demo but would appreciate suggestions.  I will definitely submit a PR to sitemap.md.  

I think this is a good enhancement for sitemap-based UIs, but as we were discussing [here](https://github.com/eclipse/smarthome/issues/1617#issuecomment-258637368), something in addition should be provided for non-sitemap-based UIs.

kaikreuzer:
I am actually not sure whether we should add this feature to the demo files at all.
It is nothing I would want to promote as a feature that should be used - it is rather something for the ticks&tricks section how to solve certain situations, if nothing else helps...

But we should probably first finish the discussion on the implementation itself.

kaikreuzer:
Ok, having had a closer look at the implementation in https://github.com/eclipse/smarthome/pull/2412, I merged it and I am also fine with adding this example to the demo files - will merge as soon as the next ESH stable is flowing to OH!

PR #334, opened 2016-11-24
adapted packaging to new mechanism

kaikreuzer:
* changed addons.cfg to new packaging types
* refactored demo packaging to use addons configuration instead of direct Karaf features
* removed experimental features

See https://github.com/openhab/openhab-core/pull/81


distributions/distribution-resources/src/main/resources/conf/services/addons.cfg
```diff
@@ -1,10 +1,12 @@
 # The base installation package of this openHAB instance (default is "standard")
 # Valid options:
 #   - minimal  : Installation only with dashboard, but no UIs or other addons
-#   - standard : Typical installation with all standards UIs
+#   - simple   : Setup for using openHAB purely through UIs - you need to expect MANY constraints in functionality!
+#   - standard : Default setup for normal users, best for textual setup
+#   - expert   : Setup for expoert users, especially for people migrating from openHAB 1.x
```
ThomDietrich: "expoert"

kaikreuzer: Thanks, this was actually a typo in the original file (fixed it now here https://github.com/openhab/openhab-distro/pull/336)

kaikreuzer:
> "MANY" sounds a bit scary in this context.

That's exactly the idea behind it - people should not use it and expect to be able to use openHAB purely through UIs with all features. It is rather a PoC and gives a glance of what is there to come in future.

PR #347, opened 2016-12-09
do not set a default package, but leave it to the user on the first start

kaikreuzer:


distributions/distribution-resources/src/main/resources/conf/services/addons.cfg
```diff
@@ -1,12 +1,15 @@
 # The base installation package of this openHAB instance (default is "standard")
```
ThomDietrich: default is "standard"

PR #354, opened 2016-12-11
reverted change of aggregate-xml

kaikreuzer:


kaikreuzer:
This should solve upgrade problems like https://community.openhab.org/t/build-637-does-not-start-any-bundles-anymore/18048.
But in any case, we still have a major issue with the upgrade process here that needs to be solved...

Issue #299, opened 2016-10-15
Upgrade mechanism


kaikreuzer:
We need to have a stable way to upgrade from release to release as well as from snapshot to snapshot.
First this needs to be solved for the distro, then for apt, docker, spk, etc.

My idea for the upgrade would be the following:
- completely replace the runtime folder by the new version
- delete userdata/cache and userdata/tmp folders
- keep conf and the rest of userdata fully in place

With this, I would assume Karaf to reconstruct the OSGi framework with all the latest versions of the bundles and using the new "openhab-base", while using all statically (conf) and dynamically (userdata) created data as before.

kaikreuzer:
I did some analysis here. A first major problem is https://github.com/openhab/openhab-distro/issues/315 - this causes configuration to be lost.

kaikreuzer:
With https://github.com/openhab/openhab-distro/pull/318, the configuration is moved to userdata/etc, so if the upgrade mechanism leaves this untouched, the "lost configuration issue" should be solved with it.

Next up: If add-ons are installed through Paper UI, these are lost if the cache folder is deleted, see https://github.com/openhab/openhab-distro/issues/319.

kaikreuzer:
My nice theory from above does not really work out - by moving the etc folder from runtime to userdata, we now also have some system files in userdata that might need to be updated on an upgrade. Likewise, we might need to offer new versions of certain conf or userdata files that might have been edited by the user and in case we do folder structure refactorings, we might even have to move stuff around.

@theoweiss explained [here](https://github.com/openhab/openhab-snap/issues/2#issuecomment-2652871859) how this is handled in Debian packages. Unfortunately, we require a mechanism that will work across all our packaging types (Debian, Docker, Synology, QNAP, Snap, etc.) and even on Windows 😦 .

Looking for good suggestions how to deal with that!

mhilbush:
> we now also have some system files in userdata that might need to be updated on an upgrade.

One thing I'll mention here, since I stumbled upon this just yesterday.  I was not seeing in PaperUI any bindings that were newly added to the distro.  It seems I got caught in the middle of this and a couple earlier changes.

https://github.com/openhab/openhab-distro/pull/354

I discovered that userdata/config/profile.cfg was still referencing this:
```
repository.mvn:org.openhab.distro/openhab-features/2.0.0-SNAPSHOT/xml/features="mvn:org.openhab.distro/openhab-features/2.0.0-SNAPSHOT/xml/features"
```

[...]

PR #356, opened 2016-12-21
upgraded to Karaf 4.0.8

kaikreuzer:
implements https://github.com/openhab/openhab-distro/issues/18

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <maven.compiler.compilerVersion>1.8</maven.compiler.compilerVersion>
 
-        <karaf.version>4.0.4</karaf.version>
+        <karaf.version>4.0.8</karaf.version>
         <maven.compiler.plugin.version>3.3</maven.compiler.plugin.version>
         <build.helper.maven.plugin.version>1.9.1</build.helper.maven.plugin.version>
 


BClark09:
image

The branding seems to have lost its way.

kaikreuzer:
Only, if you login via ssh - if it is started on command line, the correct splash screen is shown. Will have to find out where it has the karaf one stored...

@BClark09 There is a critical issue through the update, though: After an apt upgrade, the runtime will fail to start. We need to update a couple of files in userdata/etc.

What would need to be copied from a new package to the users folder is:
```
userdata/etc/*.properties
userdata/etc/org.apache.karaf.*.*
```

When these files are updated by the apt upgrade, it all should work. Could you help doing this change (here on the old packaging)? I unfortunately do not know, where to have a look...

kaikreuzer:
> Only, if you login via ssh - if it is started on command line, the correct splash screen is shown. Will have to find out where it has the karaf one stored...

Ok, this is due to [this change](https://github.com/apache/karaf/pull/210/files) and fixed by https://github.com/openhab/openhab-distro/pull/357

BClark09:
@kaikreuzer I upgraded using apt and was prompted if I wanted to replace the following files to the distributor version:

```
startup.properties
config.properties
distribution.info
org.apache.management.cfg
jre.properties
org.apache.karaf.features.cfg
org.apache.karaf.features.repos.cfg
```

After specifying "yes" to all prompts, I am running a working openHAB (online version).

The default option if the user hits enter on the prompt is to keep the users version, which I suspect is the problem. Do you think that the above files are okay to be overwritten each upgrade? If so then we'll need to extend [these lines](https://github.com/openhab/openhab-distro/blob/master/distributions/pom.xml#L146-L158) so that the files above are set as `<conffile>false</conffile>`

ThomDietrich:
@BClark09 is it possible to define yes as default? That would be the preferable solution, right?

kaikreuzer:
> Do you think that the above files are okay to be overwritten each upgrade? 

Yes, imho all those files can be considered "system" files that the user should not touch. So it is ok to overwrite them.

ThomDietrich:
Okay, then my question is invalid. ;)

BClark09:
#358 should do the trick.

@ThomDietrich it is a very valid question. It would solve many problems like this in the future. I don't know of the answer yet though.

kaikreuzer:
@BClark09 Thanks for helping out. I think we should be able to define which file is a system and which a config file by going through them one by one. But I'd actually like to better separate them as originally intended (system files should really be in "runtime", not "userdata"). I think I'll be able to do some improvements wrt Karaf, but I doubt that I will be able to fully solve it. See also my https://github.com/openhab/openhab-distro/issues/299#issuecomment-268253707.

PR #358, opened 2016-12-22
Removed conffile status of some system files.

BClark09:
Also added openhab2.service as a conf file so users can run openhab as root (if needed)


kaikreuzer:
Many thanks!

PR #359, opened 2016-12-22
Removed conf status of all org.apache.karaf*

BClark09:
...just in case.

Formatting changes for debian pom also included. Particularly I've removed the extra "/" which was causing some of the paths to be written as `/var/lib/openhab2//etc/*` @kaikreuzer, This may have been the reason behind some users problems, but I'm not sure why it didn't effect any of the tests I threw at it. Apologies for missing this the first time.

### DIFF
--- a/distributions/pom.xml
+++ b/distributions/pom.xml
@@ -149,20 +149,19 @@
                                         <type>archive</type>
                                         <includes>userdata/**</includes>
                                         <excludes>
-                                                  userdata/etc/startup.properties,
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Thanks again, @BClark09!

ThomDietrich:
In cases like this I love this addon. In case you didn't know ;) https://github.com/jerone/UserScripts/tree/master/Github_Commit_Whitespace#readme

PR #364, opened 2016-12-30
moved some system files away from userdata/etc

kaikreuzer:


BClark09:
Just making sure, would you expect custom.properties, system.properties and org.ops4j.pax.web.cfg, to be changed by the user at any point? If not, you may want to add them to the list of files to be excluded and included later just as in PR #364.

kaikreuzer:
> you may want to add them to the list of files to be excluded 

@BClark09 What list do you mean exactly?
No, I consider many more files in userdata/etc as system files that should not be touched by the user - the ones you mention are such cases. Unfortunately, Karaf expects them in the etc folder and there is no easy way to move them out. I am still experimenting with some solutions for this.

This PR here just does some simple changes that work well, so I wanted to get that out asap, while working on the rest.

BClark09:
Sorry I was speaking about the list of configuration files for the Debian packages. If anything has happened to the three files I mentioned above it might prompt the user to keep the existing files on the upgrade. I think it should be fine to merge and make a new PR that does what I did in #364 for `custom.properties`, `system.properties` and `org.ops4j.pax.web.cfg`

kaikreuzer:
The files I moved are now in runtime and not userdata anymore, so they will anyhow be completely overwritten (and should never be touched by a user). As mentioned in https://github.com/openhab/openhab-distro/issues/299#issuecomment-268253707, we should get towards a solution that is not Debian specific, so my goal is to really strictly separate system from user files.

PR #367, opened 2017-01-02
Recreate /var/lib/openhab2/tmp on upgrade

### LINKED ISSUE #366: /var/lib/openhab2/tmp folder is missing when doing an apt install/upgrade
The openHAB distro contains the folder userdata/tmp, which contains a file README.
This seems to be lacking when doing the install through apt, see [these reports](https://community.openhab.org/t/oh2-not-starting-upgrade-20170101125906/19239), which causes Karaf to not start correctly.

BClark09:
the removeCache function removes the tmp folder on each upgrade, but for some reason openhab is unable to create the folder again on install.

Workaround: Run a simple script to make the directory and set the appropriate owner before running openhab.

Fixes #366 

### DIFF
--- a/distributions/distribution-resources/src/main/resources/deb/control-runtime/postinst
+++ b/distributions/distribution-resources/src/main/resources/deb/control-runtime/postinst
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+UD_TMP=/var/lib/openhab2/tmp
+
 set -e
 if [ x"${OH_DOCKER_INSTALLATION}" != x ]; then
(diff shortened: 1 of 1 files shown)


kaikreuzer:
> but for some reason openhab is unable to create the folder again on install.

Ah, I missed that. Yes, clearly, our upgrade strategy is to "delete the tmp" folder. So I assume the upgrade just did that and did not even try to recreate it afterwards again, right? So the new correct strategy would be to "delete the content of the tmp folder or recreate it from the package".

Thanks for your PR!

PR #384, opened 2017-01-15
Switch karaf remote console key to PEM format

ThomDietrich:
In case https://github.com/openhab/openhab-linuxpkg/issues/15 gets implemented at a later point, this change would be needed. As the karaf key of users gets changed, triggered by this PR, I want to propose this for 2.0 RC/GA. The key format switch doesn't seem to change any behavior as we know it.

Signed-off-by: Thomas Dietrich <thomas.dietrich@tu-ilmenau.de> (github: ThomDietrich)


kaikreuzer:
Sorry, saw this too late for RC, but fine for GA.

ThomDietrich:
That's still okay. ;) Thanks

PR #386, opened 2017-01-16
Renaming org.apache.karaf.jaas.cfg

### LINKED ISSUE #385: " org.apache.karaf.jaas.cfg" has space at front of filename
This caught me off guard when trying to change the conf files. Is this supposed to be like this?

BClark09:
github is being a bit weird about the space diff, hopefully this fixes the issue.

Fixes #385 


BClark09:
AFAIK, the previous PR had no actual effect on this distro. So you may want to test the change with the correct filename.

PR #387, opened 2017-01-16
improved demo configuration

kaikreuzer:
- fixed broken icons
- fixed broken video widget
- fixed rules to not throw any exceptions anymore
- added rules for radio and volume control
- added tags to items
- added homekit and hueemulation add-ons
- added HABPanel demo configuration


ThomDietrich:
Would comments in the rules be wise?

Please change "// vim: syntax=Xtend syntax=openhab" to "// vim: syntax=Xtend" in all files. GitHub changed it's way how to evaluate this line and while this "hack" worked before, it's now broken. Sorry I didn't report that yet. (If you are now thinking about the vim addon, I'll take care of that differently.)

ThomDietrich:
@DocFraggle any items or rules you want to include here to be applicable in your tutorial (or the other ay around)??

DocFraggle:
@ThomDietrich we could add the network and zwave binding, of course only if that isn't too much. If we do, we could also add the items and rules from my tutorial

kaikreuzer:
> Would comments in the rules be wise?

Feel free to add some. I think at least the ones I added are pretty straight forward as they are.

> Please change "// vim: syntax=Xtend syntax=openhab" to "// vim: syntax=Xtend" in all files

Done.

> we could add the network and zwave binding, of course only if that isn't too much.

You mean here in the demo package? No, I wouldn't do that. Z-Wave is very complex and requires specific hardware setup and network is very specific. Note that the demo package is mainly about showing how the DSL files are constructed. I only added a few additional bindings of stuff that people might have floating around in their networks already, so that there is a nice "oh, it has found that" effect.

PR #412, opened 2017-01-30
Consider OPENHAB_* on startup. Closes #408

### LINKED ISSUE #408: Allow to set OPENHAB_CONF and other OPENHAB_* environment variables
I ran into an issue using the openhab2 docker image as a base image for my specific docker image which includes my specify configuration directory:

It is not possible to start openhab2 with a different configuration directory location.
Looking at [oh2_dir_layout](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/oh2_dir_layout) and [setenv ](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/setenv)only OPENHAB_HTTP_PORT and OPENHAB_HTTPS_PORT are used if set before start.

Could you please add OPENHAB_CONF to the list and maybe it's a good idea to add most of OPENHAB_* environment variables (OPENHAB_CONF, OPENHAB_USERDATA, OPENHAB_LOGDIR).

e.g. by adding something like this to oh2_dir_layout:
if [ -z ${OPENHAB_CONF } ]; then
    export OPENHAB_CONF="${OPENHAB_HOME}/conf"
fi

heziegl:
if  OPENHAB_CONF, OPENHAB_RUNTIME, OPENHAB_USERDATA or OPENHAB_LOGDIR are set upon start their values are now used instead of defaults.

Closes #408

### DIFF
--- a/distributions/openhab/src/main/resources/bin/oh2_dir_layout
+++ b/distributions/openhab/src/main/resources/bin/oh2_dir_layout
@@ -4,13 +4,25 @@
 (unset CDPATH) >/dev/null 2>&1 && unset CDPATH
 
 export OPENHAB_HOME=`cd "$DIRNAME/../.."; pwd`
-export OPENHAB_CONF="${OPENHAB_HOME}/conf"
-export OPENHAB_RUNTIME="${OPENHAB_HOME}/runtime"
-export OPENHAB_USERDATA="${OPENHAB_HOME}/userdata"
-export OPENHAB_LOGDIR="${OPENHAB_USERDATA}/logs"
+
+if [ -z ${OPENHAB_CONF} ]; then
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Thanks. Did you test whether it all still works correctly if you point the different variables to completely different folders?

@BClark09 How do the apt start scripts actually define the folders? Aren't they using the oh2_dir_layout script?

BClark09:
The apt versions have their own version:

https://github.com/openhab/openhab-linuxpkg/blob/master/resources/deb/bin/oh2_dir_layout

kaikreuzer:
Ok, thanks - so this shows that it is pretty safe to define custom folders :-)

Issue #417, opened 2017-02-05
barking.mp3 and Sonos


lolodomo:
A new file named barking.mp3  was recently added in conf/sounds.
This file is playing well on my PC using foobar; I can see that it is a MP3 with 1 channel, a bitrate of 64kbps and a sampling rate at 11025 Hz.
When I try to play it with:
```
smarthome:audio play sonos:PLAY5:chambre barking.mp3
```
nothing is heard. Maybe this kind of file is not compatible with Sonos ?
doorbell.mp3 with 2 channels, a bitrate of 128kpbs and a sampling rate at 44100 Hz is working well.
Maybe we could change barking.mp3 ?

lolodomo:
I tried to add this file in my music library in Sonos and I can confirm that this file cannot be rendered by Sonos.

yphyph01:
I confirm, Sonos can't read 11025Hz MP3 file

kaikreuzer:
Stupid Sonos...

> Maybe we could change barking.mp3 ?

Feel free to change it to 44.1kHz (I'd assume that 22kHz won't work either...).

PR #418, opened 2017-02-06
Resample barking.mp3 for compatibility with Sonos

lolodomo:
File was resampled from 11025 Hz to 44100 Hz without changing its size.

Signed-off-by: Laurent Garnier <lg.hc@free.fr>


lolodomo:
Fix #417

PR #438, opened 2017-03-20
enabling snap builds directly from distro repository

kubiko:
Adding snap craft recipe to openhab-distro.
This enabled easy way to build snap packages directly from distro repository and adds option to enable automatic builds in Launchpad


snap/snapcraft.yaml
```diff
@@ -0,0 +1,82 @@
...
+          - -start*
+          - -runtime/bin/*.bat
+        prime:
+          - -start*
+          - -runtime/bin/*.bat
+
+    # java run time
+    jre:
+        plugin: zulu-jre
+        # zulu supported arcghitectures(armhf, amd64) are already handled by plugin itself
```
tavalin: architectures spelt incorrectly

ThomDietrich:
Is this repository the right place for snap related changes? Linux packages were moved to their own repository not too long ago: https://github.com/openhab/openhab-linuxpkg

kaikreuzer:
@kubiko I would agree that snap specific stuff should be only in https://github.com/openhab/openhab-snap.
What is the reason that you want to put this here?

kubiko:
@kaikreuzer @ThomDietrich this is to enable continuous integration if we choose to.
While building snap packages from [openhab-snap](https://github.com/openhab/openhab-snap) works just fine, it's tricky to trigger build there with every change committed to distro repository. If we drop snap package recipe here, we simply setup git mirroring in launchpad and whenever there is new commit to [openhab-distro](https://github.com/openhab/openhab-distro) it will trigger new snapshot build (once mirror is synced to launchpad and commit is detected there), new snap package can be then auto uploaded to edge channel. Like this we can have in edge channel always latest snapshot build from master, without any manual work around it. And if user has installed openHAB from edge channel, then he will be automatically updated to newest snapshot.
This would make also snapshot build lot more accessible to wider audience, of course if that is desired.

kaikreuzer:
@kubiko Getting hold of commits actually won't help you much - we often run the distro build without ANY changes to the git repo, because what it does is actually to aggregate build results from OTHER repos. So my suggestion would be to set up a "snap" build job, which is automatically triggered when https://openhab.ci.cloudbees.com/job/openHAB-Distribution/ finishes (the very same is e.g. done for https://openhab.ci.cloudbees.com/job/openhab-linuxpkg/, which builds and publishes the apt packages).

kubiko:
@kaikreuzer ok that makes lot of sense. Then indeed we should keep all  snap related stuff in separate  [openhab-snap](https://github.com/openhab/openhab-snap) repo, and wire launchpad build to cloudbees

PR #443, opened 2017-04-08
Add support for setting the HTTP(S) listen address

mgbowman:
Signed-off-by: Matthew Bowman <mgbowman@users.noreply.github.com> (github: mgbowman)

### DIFF
--- a/distributions/openhab/src/main/resources/bin/setenv
+++ b/distributions/openhab/src/main/resources/bin/setenv
@@ -52,6 +52,15 @@
 . "$DIRNAME/oh2_dir_layout"
 
 #
+# set listen address for HTTP(S) server
+#
+if [ ! -z ${OPENHAB_HTTP_ADDRESS} ]; then
+    HTTP_ADDRESS=${OPENHAB_HTTP_ADDRESS}
+else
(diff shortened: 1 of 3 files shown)


maggu2810:
Wouldn't it be better to set HTTP_PORT, HTTPS_PORT and HTTP_ADDRESS using the configuration admin instead of system properties?

mgbowman:
@maggu2810 That would definitely be more "user-friendly" but I had an issue today where I couldn't even start OH because 8080 and 8443 were already in use. Once I figured out how to define the listen address (I wanted to use the default ports on a different IP), it seemed like a simple enough PR to submit.

If I had the option of setting this via the configuration admin, I could of (temporarily) suspended the competing service, fired up OH to set the the listen address, restarted OH (for the changes to be applied) and restored the competing service.

ThomDietrich:
> Could you check if you find a good place in the documentation where this feature could be mentioned?

@mgbowman maybe I can help with that. First you should add the new variable to /etc/default/openhab2. Please see here: https://github.com/openhab/openhab-linuxpkg/pull/49 for a recent improvement on this file. Second you'll want to document. The topic is already discussed in this thread: https://github.com/openhab/openhab-docs/issues/393

PR #471, opened 2017-06-26
Added update script for UNIX and MacOS systems

BClark09:
Related: #299 


BClark09:
@kaikreuzer, when this gets merged will this file appear in /runtime/bin for new zip files? If so, I think it would be good to test the self update feature of the script before announcing it.

kaikreuzer:
Thanks! I'll trigger a new distro build right away, so we can test its self-upgrading.

PR #472, opened 2017-06-26
Added function to update openHABs addon files

BClark09:
Related: #299 


BClark09:
Cheers @kaikreuzer, I've queued another build of the distro. Could you do another test with an addons file inside the directory for the Mac, just in case?

kaikreuzer:
Tested it, works nicely!

Issue #455, opened 2017-05-24
Upgrade to Karaf 4.1.x


kaikreuzer:
We should try to stay up-to-date with latest Karaf releases.
A switch to Karaf 4.1.x will bring us a more recent Jetty (9.3) version, a cleaned up etc folder and the possibility to move to the much more powerful log4j2.

Current release is 4.1.1, but 4.1.2 is expected anytime soon.

martinvw:
They do no seem to reach their plans:

4.1.2	eta 08/05/17

Or is it some non-european notation referring to august...

kaikreuzer:
Hehe, they so far hardly ever reached their plan and they even fail to update the ETAs ;-)

I am actually wondering whether it is safer to go for an openHAB 2.1 release on the current Karaf and do the upgrade directly afterwards - this will leave us more time to get everything stable again. Wdyt?

maggu2810:
> Hehe, they so far hardly ever reached their plan and they even fail to update the ETAs ;-)

We should use a respectful treatment with the work of other communities!

martinvw:
Agree on that, I did not intend any harm, I was seriously doubting the date format. With dashes is is always clear that it is European, in this case I'm in doubt.

And yes our own ETA's are also hardly ever reached :-)

kaikreuzer:
I am deeply respectful that they dare to provide ETAs, which I completely fail to do for openHAB :-)
So "failing to reach the ETA" is already one quality level higher than my own project management. I hope this proofs that I definitely didn't mean to blame anyone 😎

martinvw:
@kaikreuzer do you agree that it is getting a little late to switch to this version, if it gets released. Maybe for the next version...?

kaikreuzer:
Yes, see my https://github.com/openhab/openhab-distro/issues/455#issuecomment-305723011.

kaikreuzer:
@martinvw A first try using 4.1.1 (4.1.2 is still not available...) is here: #484

PR #484, opened 2017-07-15
Upgrading to Karaf 4.1.2

kaikreuzer:
Implements #455


kaikreuzer:
This is work in progress - it seems to work quite well already, but I still have a few issues that need to be addressed:

- There is an exception on logout, see [here](https://issues.apache.org/jira/browse/KARAF-5111?focusedCommentId=16088505&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16088505).
- Karaf creates userdata/log/karaf.log with an empty file. Apparently this happens before our own logging configuration is processed (which uses other files). Not sure how to avoid this.
- Karaf creates runtime/instances, although our runtime folder is supposed to be read-only and thus nothing should be written in it.

@dvanherbergen If you happen to like spending a few hours on openHAB again, feel free to help on this :-) Everybody else is very welcome to test and help as well, of course!

dvanherbergen:
Unfortunately I already have too many things on my todo list, so I don't have much time to help. But I did notice that the instances location is determined in the [karaf shell script](https://github.com/apache/karaf/blob/8c1e8218f9c6185b51792d44fe317453408d3ae6/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf#L282).
If you update all references in this script from $KARAF_HOME/instances to $KARAF_BASE/instances, that should relocate the folder.

kaikreuzer:
Thanks @dvanherbergen, it was worth a try to ask you ;-)

Good observation regarding the karaf script - I actually just noticed that the very same issue already exists in the current 4.0.8-based openHAB distro, so this is nothing new here...

dvanherbergen:
No problem. If you can find a way to add the property
```karaf.bootstrap.log = ${karaf.base}/logs/openhab.log```
to /userdata/etc/config.properties, that will redirect the karaf.log to the same log file as openhab.
In the karaf script, you will also need to remove the [creation of the default log dir](https://github.com/apache/karaf/blob/8c1e8218f9c6185b51792d44fe317453408d3ae6/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf#L261).

**Comment by kaikreuzer** [comment]
Wonderful, you're the best! I'd never heard of `karaf.bootstrap.log`, but this indeed solves the issue!

> you will also need to remove the creation of the default log dir.

Already done [here](https://github.com/openhab/openhab-distro/pull/484/commits/b816eb7c9f64266ab81fe1fb560f2de8cf4b130d#diff-9001988e213bff2abe70b494dd96eb80R245) :-)

**Comment by kaikreuzer** [comment]
Log issue is solved and I have just tested Karaf 4.1.2-SNAPSHOT, which also resolves the exception at startup. I'll hence wait for the 4.1.2 release and I think we should then be fine for the upgrade!

**Comment by kaikreuzer** [comment]
Karaf 4.1.2 has been released this week.
I have updated this PR and all looks pretty good.

@BClark09 & @ThomDietrich There is one question regarding the update process: Karaf 4.1 now uses log4j2 instead of log4j - which is in general good news. The bad news is that the syntax of `userdata/etc/org.ops4j.pax.logging.cfg` has changed. As users might have customized that file, we do not overwrite it in our update script. What is your suggestion how we should deal with it?

**Comment by kaikreuzer** [comment]
Assuming that people have a backup, I would say that overwriting this file should be ok - so how about changing the update script accordingly?

I have just found that log4j2 now also [supports multiple config files](https://stackoverflow.com/a/38408283), so that would be ideal to have the default config in future in runtime/etc and the user specific changes in userdata/etc. But this will take a bit longer investigation how to exactly do this in the context of pax-web and Karaf...

**Comment by BClark09** [comment]
That's cool, if you can have additional files then will this be safe to overwrite each upgrade? In that case we just have to add that file to the list in the upgrade scripts. 

For linuxpkg, by default it will warn people there's a change in the configuration, but if they ignore this warning they'll have the old version. I assume you'd prefer to always overwrite?

**Comment by kaikreuzer** [comment]
>  if you can have additional files then will this be safe to overwrite each upgrade? 

Yes, I would even say that we probably should not have anything in userdata/etc anymore about logging - the standard stuff should be in runtime/etc and the user specific one somewhere in conf... But that's a separate issue.

> I assume you'd prefer to always overwrite?

Is the "alternative" written to disk (i.e. a backup of the user version in case it is overwritten resp. the new version if the user chooses to keep his one)? If so, overwriting would imho be best as otherwise the user won't at all get the new version - and his old one won't work, because it is the wrong syntax.

**Comment by BClark09** [comment]
I'd have to think about the best way of doing that, it'd need a preinst script to save the old version first.

>  I would even say that we probably should not have anything in userdata/etc anymore about logging - the standard stuff should be in runtime/etc 

Yes, the sooner we can remove static files out of userdata the better. And then when we can get configuration files out of userdata too that'd be perfect!

**Comment by kaikreuzer** [comment]
> I'd have to think about the best way of doing that, it'd need a preinst script to save the old version first.

Is apt using the update script by now, so that we are sharing the same logic?
If so, we could simply add a line in the update script that creates a copy of the user version and then overwrites it with the new one.

> And then when we can get configuration files out of userdata too that'd be perfect!

That's my goal since long, but Karaf is really hard to pursuade this way...

**Comment by kaikreuzer** [comment]
@BClark09 Could you have a look at https://github.com/openhab/openhab-distro/pull/484/commits/417878b793a65c115c7634f5c664b2f9a3f3f403? Would that be ok as a temporary solution so that we can roll out Karaf 4.1.2 now? I will post the info in the community as well, once it is in the build.

**Comment by BClark09** [comment]
Yes it looks about right for the update scripts. But I'd have to change the preinst script on linuxpkg to do the same. 

The update script can't be directly used by apt or yum, as they'd be trying to mess with the same files. This is why we were after a csv type transition script right? So that they could be used by both?

**Comment by BClark09** [comment]
If this were to be merged now, it would let the updater know if there's a change in the file, if it's never been edited by them it will be replaced anyway.

**Comment by kaikreuzer** [comment]
> This is why we were after a csv type transition script right?

Absolutely :-)

So if I get you right, I have green light to merge?

**Comment by BClark09** [comment]
I believe so, when you make the announcement though please advise apt users to respond with Y. Otherwise, they'll have to copy org.ops4j.pax.logging.cfg.dpkg-dist to org.ops4j.pax.logging.cfg.

If they do make the change, but they want to copy over their old stuff, their version will be called org.ops4j.pax.logging.cfg.dpkg-old

File: distributions/openhab/src/main/resources/bin/karaf [hunk 1]
```diff
@@ -0,0 +1,327 @@
...
+}
+
+REALNAME=$(realpath "$0")
+DIRNAME=$(dirname "${REALNAME}")
+PROGNAME=$(basename "${REALNAME}")
+
+#
+# Load common functions
+#
+. "${DIRNAME}/inc"
```
BClark09: Is this included? It doesn't seem to be with the rest of the PR?

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -112,6 +112,10 @@ Function Update-openHAB {
         Remove-Item ($userdata + '\cache') -Recurse -ErrorAction SilentlyContinue
         Remove-Item ($userdata + '\tmp') -Recurse -ErrorAction SilentlyContinue
 
+        # We need to keep a backup in case the user modified that file
+        Copy-Item ($userdata + '\etc\org.ops4j.pax.url.mvn.cfg') -Destination ($userdata + '\etc\org.ops4j.pax.url.mvn.cfg.bak') -Force
+        Remove-Item ($userdata + '\etc\org.ops4j.pax.url.mvn.cfg') -ErrorAction SilentlyContinue
```
BClark09: Mentioned in a separate comment, but thought i'd post as a review too: This is a different file to the logging file used in bin/update

kaikreuzer: Good catch! Indeed a copy&paste error - it was already late last night ;-)

distributions/openhab/src/main/resources/bin/karaf
kaikreuzer: No, but I hope it does not harm either.
This file is the standard file from Karaf, which I didn't change, but for [some fixes](https://github.com/openhab/openhab-distro/pull/484/files/417878b793a65c115c7634f5c664b2f9a3f3f403#diff-9001988e213bff2abe70b494dd96eb80R271) regarding the log folder.

kaikreuzer:
> Yes, the sooner we can remove static files out of userdata the better. And then when we can get configuration files out of userdata too that'd be perfect!

FTR, I have created an issue about that: https://github.com/openhab/openhab-distro/issues/516

distributions/openhab/src/main/resources/userdata/etc/shell.init.script
```diff
@@ -0,0 +1,56 @@
...
+help = { *:help $args | more } ;
+man = { help $args } ;
+log:list = { log:get ALL } ;
+service:get = { $.context getService ($.context getServiceReference $args) };
+
+env = { shell:set $args }
+edit = { shell:nano $args }
+more = { shell:less $args }
+
+//prompt = "[1m${USER}[0m@${APPLICATION}(${SUBSHELL})> "
```
martinvw: @kaikreuzer any reason this was commented, I'm debuging some problems with docker: https://community.openhab.org/t/docker-compose-vs-docker-error-in-initialization-script/34025

kaikreuzer: Afair, it messed up our console prompt (we do not want user/application/subshell information).
If you find that we must not comment it out, but instead define "our" prompt here, we might change it to [the prompt from here](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/filtered-resources/userdata/etc/branding-ssh.properties#L15).

martinvw: Indeed I changed it in an attempt to fix but did not solve the problem but it did make the prompt ugly 😉

pyjamasam:
Just going to chime in here to hopefully ping the people that need to be (@kaikreuzer).  The bot picked up the forum post though and it has the details in them so I won't re-iterate them.

kaikreuzer:
Hi @pyjamasam, It would be much better if you could open a new issue / bug report here rather than commenting on an old and merged PR!

kaikreuzer:
And please DO re-iterate (copy&paste) the relevant content to the issue, so that it is immediately clear what the problem is.

pyjamasam:
Sounds good.  I only posted here as this PR was the one that broke it.

New issue inbound...

PR #496, opened 2017-07-26
Add die message function to setenv

BClark09:
`runtime/bin/setenv` is called in `/runtime/bin/karaf` before the `die` function is defined. Since `setenv` uses `die`, we should define it here.

The reasons for the `&& /bin/true` workaround are at https://github.com/openhab/openhab-linuxpkg/pull/72#issuecomment-315424714 I'd understand if this isn't acceptable though.

### DIFF
--- a/distributions/openhab/src/main/resources/bin/setenv
+++ b/distributions/openhab/src/main/resources/bin/setenv
@@ -108,6 +108,12 @@ export EXTRA_JAVA_OPTS="${EXTRA_JAVA_OPTS_COMMON} ${EXTRA_JAVA_OPTS_ARCH} ${EXTR
 # The functions below are copied from the karaf script to set JAVA_HOME to a default value.
 # This avoids Karaf printing a warning message at startup.
 
+die() {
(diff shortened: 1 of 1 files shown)


BClark09:
CC @ThomDietrich

distributions/openhab/src/main/resources/bin/setenv
```diff
@@ -108,6 +108,12 @@ export EXTRA_JAVA_OPTS="${EXTRA_JAVA_OPTS_COMMON} ${EXTRA_JAVA_OPTS_ARCH} ${EXTR
 # The functions below are copied from the karaf script to set JAVA_HOME to a default value.
 # This avoids Karaf printing a warning message at startup.
 
+die() {
+  # echo and exit, && /bin/true delays the exit for systemd logging (workaround).
+  echo "$*" && /bin/true
+  exit 1;
```
ThomDietrich: ;

BClark09: ;)

PR #498, opened 2017-07-27
Add Require statement

nwalk7800:
Expand-Archive requires PowerShell v5.
The Require statement will prevent the script from running on older versions and causing broken updates.


kaikreuzer:
Thanks!

Issue #499, opened 2017-07-28
Add backup/restore scripts


BClark09:
Continuing on from #299...

Separate files that backup and restore the openHAB configuration should be made. Users should be able to specify a file for the archive to be created/extracted depending on the script.

---
### General
 - Assume that the scripts are being called from openHAB's root directory or work backwards from the script path (`../../`).
 - If no filepath is specified, save to `[openHAB]/backups`.
 - If no filename is specified, use a timestamp.
 - Backups should not contain `userdata/cache` , `userdata/tmp` and any of the following files:
    - `/userdata/etc/all.policy`
    - `userdata/etc/branding.properties`
    - `userdata/etc/branding-ssh.properties`
    - `userdata/etc/config.properties`
    - `userdata/etc/custom.properties`
    - `userdata/etc/version.properties`
    - `userdata/etc/distribution.info`
    - `userdata/etc/jre.properties`
    - `userdata/etc/profile.cfg`
    - `userdata/etc/startup.properties`
    - `userdata/etc/org.apache.karaf*`
    - `userdata/etc/org.ops4j.pax.url.mvn.cfg`

---

### Specific to Linux/Mac OS 
  - Use $OPENHAB_CONF and $OPENHAB_USERDATA first if they exist.
  - Files should be `backup` and `restore` with no extension and executable rights set (755).

@whopperg, you have already done most of this, care to finish it off and send a PR?

---

### Specific to Windows
  - Use %OPENHAB_CONF% and %OPENHAB_USERDATA% first if they exist.
  - Files should be `backup.bat` and `restore.bat`, there should be no need to use powershell here.

@bdleedy and @ghys, I'm not sure if anything exists here yet, care to take a look?

ghys:
I'm not aware of anything existing for backup and restore on Windows.
I have the same opinion about PowerShell vs batch files as in #500.

BClark09:
Another thought, backups should be platform independent. It should be perfectly reasonable for a user to move their windows installation to a Linux server, or visa versa. For this reason, we should use zip.

Some methods of compressing with zip will remove previous file ownership. I'd suggest using a file inside the backup to determine a couple of things so that they can be restored later:

backup.properties
```
version=2.1.0
user=openhab
group=openhab
```

Secondly, if no path is specified where should the zip file be stored? For manual installations this can be `openhab2/backup` but for apt/rpm? @ThomDietrich, do you have any ideas?

BClark09:
[...]

PR #507, opened 2017-08-04
Add backup and restore scripts for Linux/MacOS

BClark09:
Related to #499 

### DIFF
--- a/distributions/openhab/src/main/resources/bin/restore
+++ b/distributions/openhab/src/main/resources/bin/restore
@@ -1,0 +1,189 @@
+#!/bin/sh
+
+setup(){
+  if [ -z "$1" ] || [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
+    echo "Usage: restore filename"
+    echo ""
+    echo "  e.g. ./restore myBackup.zip  << Restores config from myBackup.zip"
+    echo ""
+    echo "Use this script to restore an openHAB configuration that was previously made with"
+    echo "the openHAB 'backup' script."
+    echo ""
+    exit 0
+  fi
+
+  ## Ask to run as root to prevent us from running sudo in this script.
+  if [ "$(id -u)" -ne 0 ]; then
+    echo "Please run this script as root! (e.g. use sudo)" >&2
+    exit 1
+  fi
+
+  command -v unzip >/dev/null 2>&1 || {
+    echo "'unzip' program was not found, please install it first." >&2
+    exit 1
+  }
+
+  ## Check to see if processes are running before restoring
+  if [ ! -z "$(pgrep -f "openhab2.*java")" ]; then
+    echo "openHAB is running! Please stop the process before restoring." >&2
+    exit 1
+  fi
+
+  WorkingDir="$(cd "$(dirname "$0")" && cd ../.. && pwd -P)"
+
+  ## Set path variables
+  if [ -r /etc/profile.d/openhab2.sh ]; then
+    . /etc/profile.d/openhab2.sh
+  elif [ -r /etc/default/openhab2 ]; then
+    . /etc/default/openhab2
+  fi
+  if [ -z "$OPENHAB_CONF" ];       then OPENHAB_CONF="$WorkingDir/conf"; fi
+  if [ -z "$OPENHAB_USERDATA" ];   then OPENHAB_USERDATA="$WorkingDir/userdata"; fi
+
+  echo "Using '$OPENHAB_CONF' as conf folder..."
+  echo "Using '$OPENHAB_USERDATA' as userdata folder..."
+
+  ## Check two of the standard openHAB folders to make sure we're updating the right thing.
+  if [ ! -d "$OPENHAB_USERDATA" ] || [ ! -d "$OPENHAB_CONF" ]; then
+    echo "Configuration paths are invalid..." >&2
+    echo "Try setting OPENHAB_USERDATA and OPENHAB_CONF environment variables." >&2
+    exit 1
+  fi
+
+  CurrentVersion="$(awk '/openhab-distro/{print $3}' "$OPENHAB_USERDATA/etc/version.properties")"
+
+  ## Store anything in temporary folders
+  TempDir="/tmp/openhab2/restore"
+  ## Clear older stuff if it exists
+  rm -rf "${TempDir:?}"
+  echo "Making Temporary Directory"
+  mkdir -p "$TempDir" || {
+    echo "Failed to make temporary directory: $TempDir" >&2
+    exit 1
+  }
+}
+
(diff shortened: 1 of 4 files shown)


BClark09:
That's my best stab at it for now, would appreciate reviews, suggestions and testing @ThomDietrich and @whopperg :)

BClark09:
image

thopiekar:
Like it! I previously also added something like a backup mechanism for QNAP NASs. As soon as this is ready I'll adopt it.

BClark09:
Thanks @thopiekar, looking at the folder structure it for QNAP NASs it should work fine, have you already tested it for it?

@kaikreuzer, could you test these scripts on MacOS?

ThomDietrich:
Hey @BClark09, I went through the code and everything looks great.

The only thing I still don't quite like is "Usage: ./runtime/bin/backup [filename]". Feels unnatural to demand a relative path for execution of a script. As a user I'd like to run the script with an absolute path, from inside the bin folder.

Furthermore I'm wondering if we should add symlinks to these scripts (and others) to `/usr/bin` and `/usr/sbin` in the https://github.com/openhab/openhab-linuxpkg project.

thopiekar:
@BClark09 No, I haven't but implemented a simple function to my service script to make backups of the whole distribution (only working when openHAB is offline, of course.).
So yeah, nice to have a commonized script/function now!

BClark09:
@ThomDietrich ,

> The only thing I still don't quite like is "Usage: ./runtime/bin/backup [filename]". Feels unnatural to demand a relative path for execution of a script/program

This is only necessary if the OPENHAB_* environment paths are unknown. It'll work anywhere otherwise. I'd rather not try to work out where the appropriate path is from the file location, since this starts to break down with symlinks.

> Furthermore I'm wondering if we should add symlinks to these scripts to /usr/bin and /usr/sbin in the https://github.com/openhab/openhab-linuxpkg project.

Actually, I was considering adding /usr/bin/openhab soon to add the commands:

```
openhab console
openhab backup $@
openhab restore $@
openhab showlogs
```

thoughts?

thopiekar:
I agree with @ThomDietrich. It would be nice, if the filename would be optional. This way people can decide whether they want to create a archive inside the distro or eg. outside.

kaikreuzer:
> @kaikreuzer, could you test these scripts on MacOS?

Tested, not working. I get the message (both on backup and restore)
```
cp: the -R and -r options may not be specified together.
cp: the -R and -r options may not be specified together.
```
and the backup only contains empty conf and userdata folders.

Regarding the message on restore:

> Any existing files with the same name will be replaced, but this will NOT delete existing files that have no replacement.

Why not? If I restore a backup, I would want the setup of that moment and not some mix of current stuff and restored stuff. Or do I misinterpret the notice?

thopiekar:
@kaikreuzer That's why I make a full copy of the whole distro and then if something goes wrong I remove everything and restore all from the backup.
Just like I would expect it from a 'snapshot' of a known working state.

BClark09:
Alright will rework. Not sure about that -R, -r error though. It only uses one.

kaikreuzer:
Checking the man page it says:

```
     -a    Same as -pPR options. Preserves structure and attributes of files but not directory structure.
```
so by specifying `-a` you implicitly have a `-R` as it seems.

kaikreuzer:
... and indeed, when I remove the `r`, the backup script seems to work nicely 👍

BClark09:
Alright, made a couple of changes that should mean that the script can be called from anywhere, and now should work on MacOS (Thanks @kaikreuzer!)

> I would want the setup of that moment and not some mix of current stuff and restored stuff. Or do I misinterpret the notice?

You didn't misinterpret, but I understand this is probably the desired choice. How best to proceed with that then? Delete all but the specific `userdata/etc/` files and then the backup on top of the somewhat empty directory?

BClark09:
Hi all, I've made some adjustments so that it completely replaces userdata and conf, getting rid of anything that wasn't in the backup.

Please could you check and test these new changes, as we need to eliminate anything that could be potentially dangerous.

bdleedy:
I'm writing the Windows version of this and I'm noticing that you put the backup in runtime\bin\backups. If you would then use upgrade that backup would then be deleted. Is this intended? Is there a plan to modify upgrade to handle retaining the backup?

BClark09:
Hi @bdleedy, the backup should be stored in openhab_home/backups by default. Is that not the case here?  The line:

```
 WorkingDir="$(cd "$(dirname "$0")" && cd ../.. && pwd -P)"
```

should point two directories up from where the backup script is.

bdleedy:
NM. Duh. Apparently too much code floating around in my head yesterday. Thanks.

BClark09:
Alright, I think I've finished with this. @ThomDietrich, @kaikreuzer and @thopiekar, I've taken all comments on board. Please let me know how you get on with testing and reviewing the two scripts. I've tested on a debian virtual machine, but please try to break the scripts as much as possible.

distributions/openhab/src/main/resources/bin/restore
```diff
@@ -0,0 +1,191 @@
...
+  }
+fi
+
+echo "Deleting old conf folder..."
+rm -rf "${OPENHAB_CONF:?}/"*
+
+
+## Restore configuration
+echo "Restoring openHAB with backup configuration..."
+command cp -af "$TempDir/conf/"*     "${OPENHAB_CONF:?}/" || {
```
ThomDietrich: what's the `command` needed for?

BClark09: Some distributions use an alias for cp which includes the -i flag. 

https://serverfault.com/questions/119869/turning-off-cp-copy-commands-interactive-mode-cp-overwrite

kaikreuzer:
@BClark09 I have tested it on Mac, seems to run smoothly! 👍

kaikreuzer:
Any further comments/tests by the others? Or do you think we are good to merge?

ThomDietrich:
I did not get around doing any tests (my time is a bit limited atm) but I went through the whole code and it LGTM.

BClark09:
Hi @ThomDietrich and @kaikreuzer, would you be able to test this once more? I cleaned up and used the scripts on my own home today so these should be fine to use. I've tried to harden the script as much as possible but please try to break it if you can :)

ThomDietrich:
LGTM. Great work @BClark09 !

whopperg:
TOP!!! @BClark09

kaikreuzer:
Thanks @BClark09, successfully tested it on Mac as well!

BClark09:
Awesome, thanks everyone! Shall I put a post up on the community forums about this?

Issue #500, opened 2017-07-28
Add update.ps1 script functionality


BClark09:
Following from #299

The windows script needs a few modifications to bring it up to the level of functionality that the Linux/Mac OS equivalent has:

 - It does not currently handle the update of addon files if old ones exist.
 - Ideally, the script should run it's own newer version from the downloaded archive. Otherwise, specific processes will always be one update too stale.

Additionally, the script is only compatible with PowerShell v5, this ought to be lowered to suit more installations. Ideally, as a `.bat` script instead of `.ps1`.

@ghys, @bdleedy any thoughts on this?

ThomDietrich:
I have no experience with PowerShell but judging from the information found here: https://en.wikipedia.org/wiki/PowerShell#PowerShell_4.0 Version 4 can be found on all Windows 7 and up PCs and would hence be acceptable. Going back to Batch is imho too much...

BClark09:
PowerShell 2 is found on all Windows 7 machines, although 4.0 can be downloaded and installed on any SP1 version, and it is fairly likely (but not certian) to have WMF 4.0 (which includes PS4.0) installed nowadays.

ThomDietrich:
![grafik](https://user-images.githubusercontent.com/2870104/28771115-df9ff158-75e1-11e7-8911-94eeacde2ce7.png)

Confirmed

ghys:
I'm reluctant about rewriting the current PowerShell script as a batch file too. we can expect PowerShell v2 to be installed on all Windows platforms.
Moreover, it would be a better fit to include it in https://github.com/openhab/openhab-docs/issues/447 which is PowerShell-based.

bdleedy:
Adding the 4.0 (or 2.0, do we seriously think that someone running openHAB WOULDN'T have their Windows machine updated??) requirement to the script would be good. I was going to suggest that. I haven't had time at all since the script was added. Next week the wife is out of town for business so I'll have all week to work on this. Since Windows 10 1703, shift-right click only produces an option to open PowerShell in the current location. It's becoming the defacto shell and is way easier and more powerful for this kind of application.

bdleedy:
Whoops, apparently Thom is one of those people.

bdleedy:
I'm starting to merge my updates into the script and also looking at what would need to be downgraded for PS2.0. It makes everything MUCH harder. I'm hoping that we can decide to go with 4.0.

bdleedy:
[...]

PR #509, opened 2017-08-06
Update functionality for Issue #500

bdleedy:
- Check if we're in the correct directory before stoping and deleting current service
- Added conf directory check
- Added version check
- Check for service LIKE openHAB%
- Use system TEMP for .zip and unzipped files
- Added Download function and changed all downloads to WebClient for consistency
- Added all userdata deletes from documentation
- Changed unzip to use PS 4.0 friendly class
- Added Addon and LegacyAddon updates

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -58,6 +58,14 @@
           value="enabled"/>
       <setupTask
           xsi:type="setup:PreferenceTask"
+          key="/instance/org.eclipse.jdt.core/org.eclipse.jdt.core.compiler.annotation.nonnull"
+          value="javax.annotation.Nonnull"/>
+      <setupTask
+          xsi:type="setup:PreferenceTask"
+          key="/instance/org.eclipse.jdt.core/org.eclipse.jdt.core.compiler.annotation.nullable"
+          value="javax.annotation.Nullable"/>
+      <setupTask
+          xsi:type="setup:PreferenceTask"
           key="/instance/org.eclipse.jdt.core/org.eclipse.jdt.core.compiler.annotation.nullanalysis"
           value="enabled"/>
       <setupTask


ghys:
Cool, the only thing missing now is a check to detect if the script itself has changed after downloading the new distribution, and run the new version as appropriate.

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -1,4 +1,27 @@
...
+        [string]$DownloadSource,
+        [Parameter(Mandatory=$true)]
+        [string]$Outputfile
+    )
+    $client = New-Object System.Net.WebClient
+    try {
+        $client.DownloadFileAsync($DownloadLocation,$Outputfile)
+        while ($client.IsBusy)
+        {
+            Write-Host "Downloading file..."
```
ghys: The reason I used `Start-BitsTransfer` and `Invoke-WebRequest` (the former didn't work with Cloudbees for some SSL-related reason) is that they handle progress nicely themselves.
Could you maybe integrate https://stackoverflow.com/a/4927295 instead of reporting progress like this?

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -1,4 +1,27 @@
...
+            Sleep -Seconds 3
+        }
+    }
+    catch [Net.WebException]
+    {
+        Write-Host $_
```
ghys: I wouldn't simply catch the exception and let the script continue if there's an error with the download.

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -38,51 +60,57 @@ Function Update-openHAB {
...
         Write-Host -ForegroundColor Cyan "Checking whether a service exists..."
-        $service = (Get-WmiObject Win32_Service -filter "name='openHAB2'")
+        $service = (Get-WmiObject Win32_Service -filter "name LIKE 'openHAB%'")
         if ($service) {
             # Stop the service
             Write-Host -ForegroundColor Cyan "Stopping the service..."
-            Stop-Service openHAB2 -Force
-
+            Stop-Service $service.Name -Force
+
```
ghys: Extraneous whitespace

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -93,48 +121,64 @@ Function Update-openHAB {
```
ghys: My fault actually (sorry) but I just noticed: line 117, `mkdir backup` should be `mkdir $OHDirectory\backup`.
The update could be launched from another directory with the `-OHDirectory` parameter explicitely set.

distributions/openhab/src/main/resources/bin/update.ps1
bdleedy: I used BITS in mine as well (and had cool progress scripting) but like you said, the cloudbees has an issue with that. I think it may have something to do with http 1.1. Since I was looking into downgrading to PS 2.0 I stated looking here. I could revert to Invoke-WebRequest for both. (I wish BITS worked for CB)

distributions/openhab/src/main/resources/bin/update.ps1
bdleedy: Whoops. Should have put a Todo. I'll fix that.

bdleedy:
@ghys How do I use GitHub to make these changes?

distributions/openhab/src/main/resources/bin/update.ps1
ghys: `Invoke-WebRequest` has poor performance - I suppose it's not recommended for downloading large binaries.

Using `WebClient.DownloadFileAsync` is better, but you should really replace the while loop by the event handling/`Write-Progress` in the StackOverflow link above, and it would report progress in a standard way, much like `Start-BitsTransfer` (I hope).

ghys:
> How do I use GitHub to make these changes?

Just push more commits and they will appear here (btw, you should've created a dedicated branch for this PR, but never mind, it's too late now 😄).

ghys:
Also please check https://github.com/openhab/openhab-distro/blob/master/CONTRIBUTING.md
You need to add sign-off lines to your commits for the PR to be accepted.

bdleedy:
@ghys I think I committed correctly this time (signed off correctly?) and have a good download function now.

bdleedy:
@ghys Agh. I jacked this PR all up. I was trying to create a new PR for the backup script and was mashing the keyboard (sadfsadf) and it came into THIS PR. I don't have a local repo so I'm not finding any way to fix this. Any input?

ghys:
@bdleedy do a `git reset --hard 88ac471` then `git push -f` and don't touch your master branch (commits/merges) until this PR is merged.
Create a separate PR with your `bleedy-backup-ps1` branch as a base for the backup script.

This PR as of https://github.com/openhab/openhab-distro/pull/509/commits/88ac471287822a2daf483515efe01bfa4f0e5638 LGTM.

PR #511, opened 2017-08-07
Update update.ps1

bdleedy:
-Checks if we're in the correct directory before stopping and deleting
current service
- Added conf directory check
- Added version check
- Check for service LIKE openHAB%
- Use system TEMP for .zip and unzipped files
- Added download function with progress and changed all downloads to WebClient for
consitancy
- Added additional userdata deletes from documentation
- Changed unzip to use PS 4.0 friendly class
- Added Addon and LegacyAddon updates
- Fixed backup directory creation

Signed-off-by: Brian Leedy <awsnap@gmail.com>


bdleedy:
For #500 
Removed original PR to create with different branch to allow second PR for backup script.

kaikreuzer:
Thanks for this update, I somehow completely missed this PR, sorry!
Could we have any Windows user test this before merging? How about @MHerbst?

tmrobert8:
@kaikreuzer @bdleedy 
Mind if I make some comments on this?  I just tried it (win10, 64-bit):

1. Check for version.properties - if it's not found, you should silently ignore.  Right now it throws off three exceptions if version.properties doesn't exist (Get-Content: cannot fine path, then two null messages when trying to use the resulting null $VersionLine
2. I'd blow away the $TempDir right before you create it - if it already exists, it could be filled with junk stuff that could interfere with future runs.  
3. Copy-Item $OHDirectory/* - should really check to see if the directories exist prior to running them.  If the script failed at some point after you removed the directories later in the script (which mine did fail after that), a rerun of the script will have those copy-items fail with file not found messages.
4. ExtractToDirectory needs "$TempDir\" before "openhab-$OHVersion.zip".  Without that - it was looking in the current directory (not $TempDir) for the downloaded file.
5.  Making the backup directory will always fail with a "already exists" message if you've run the script mulitple times.  Might want to check to see if the directory exists before making it.
6.  Download progress bar never finishes until the whole script is finished (after it's downloaded and processed everything).  Personally - I'd get rid of the progress bar and change to a non-async download for this script.

Otherwise - seems to work ;)

Tim

tmrobert8:
btw - also wrapped this into an 'update.bat' file that contained the following line.  Makes it much easier for someone to run

`powershell -command "& { . .\update.ps1; Update-openHAB -OHVersion 2.2.0 -Snapshot $true }"`

tmrobert8:
One more comment:
1.  When you copy the userdata files back - seems like the Join-Path statement is wrong.  Seems like it should be "$OHDirectory\userdata" rather than "$TempDir..." since you want to copy the item to the OH directory?

kaikreuzer:
@bdleedy Are you still around? As we are heading towards the 2.2 by the end of this week, is there any chance to finish this off?

bdleedy:
Yes on this too. Tuesday I’ll have time.

(Sent from my iPhone)

On Dec 10, 2017, at 3:18 PM, Kai Kreuzer <notifications@github.com> wrote:

@bdleedy Are you still around? As we are heading towards the 2.2 by the end of this week, is there any chance to finish this off?

bdleedy:
@tmrobert8 

1. When would version.properties not exist? Easy enough to fix though.
2. Deleting $TempDir if exists.
3. If Backup exists, give option to overwrite
4. Yup, I missed this
5. New workflow handles this better
6. The progress bar is a bit behind however I think the feedback is more helpful than the few seconds lost.
7. Copy back fixed

Can someone test this please?

tmrobert8:
@bdleedy 
Works MUCH better (although I still don't like the progress bar - script says it's done but the progress bar is still going - why not just close it or 100% it at the end?).

Only issue:
```
Copy-Item ($userdata + '\etc\org.ops4j.pax.logging.cfg') -Destination ($userdata + '\etc\org.ops4j.pax.logging.cfg.bak') -Force
        Remove-Item ($userdata + '\etc\org.ops4j.pax.logging.cfg') -ErrorAction SilentlyContinue
```

You didn't update those lines since you removed $userdata - so the script tries to copy "c:\etc\org...cfg" rather than use the openhab distribution directory

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -38,83 +65,107 @@ Function Update-openHAB {
...
+        Remove-Item ($OHDirectory + '\userdata\etc\org.ops4j.pax.url.mvn.cfg') -ErrorAction SilentlyContinue
+        Remove-Item ($OHDirectory + '\userdata\etc\profile.cfg') -ErrorAction SilentlyContinue
+        Remove-Item ($OHDirectory + '\userdata\etc\startup.properties') -ErrorAction SilentlyContinue
+        Remove-Item ($OHDirectory + '\userdata\etc\version.properties') -ErrorAction SilentlyContinue
+        Remove-Item ($OHDirectory + '\userdata\etc\org.apache.karaf*') -ErrorAction SilentlyContinue
+        Remove-Item ($OHDirectory + '\userdata\cache') -Recurse -ErrorAction SilentlyContinue
+        Remove-Item ($OHDirectory + '\userdata\tmp') -Recurse -ErrorAction SilentlyContinue
 
         # We need to keep a backup in case the user modified that file
         Copy-Item ($userdata + '\etc\org.ops4j.pax.logging.cfg') -Destination ($userdata + '\etc\org.ops4j.pax.logging.cfg.bak') -Force
```
BClark09: These two line also need the `$OHDirectory +` base directory. Points to 'C:\etc\org.ops4j.pax.logging.cfg'

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -1,4 +1,32 @@
-#Requires -Version 5.0
+﻿#Requires -Version 4.0
+function DownloadFiles {
```
BClark09: I got the message that the Download completed *after* the script finished. The function then hangs indefinitely with the progress bar on display.

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -38,83 +65,107 @@ Function Update-openHAB {
...
+            Write-Host -ForegroundColor Cyan "Downloading the openHAB $OHVersion distribution..."
+            DownloadFiles $DownloadLocation "$TempDir\openhab-$OHVersion.zip"
         }
 
         Write-Host -ForegroundColor Cyan "Making a backup in $OHDirectory\backup..."
-        mkdir backup | Out-Null
-        Copy-Item $OHDirectory\conf $OHDirectory\backup\conf -Recurse -Force
-        Copy-Item $OHDirectory\runtime $OHDirectory\backup\runtime -Recurse -Force
-        Copy-Item $OHDirectory\userdata $OHDirectory\backup\userdata -Recurse -Force
+        if (Test-Path "$OHDirectory\backup")
```
BClark09: Is this temporary? If not, should it use the backup script you've already implemented?

kaikreuzer:
@bdleedy Any chance to address the remaining comments by tomorrow?

distributions/openhab/src/main/resources/bin/update.ps1
bdleedy: Seeing this now... I'll look into this. This is NOT how it was working before.

bdleedy:
Working feverishly.

distributions/openhab/src/main/resources/bin/update.ps1
bdleedy: Completely reworked this

distributions/openhab/src/main/resources/bin/update.ps1
bdleedy: Completely reworked this to use new backup.ps1 file

distributions/openhab/src/main/resources/bin/update.ps1
bdleedy: This is backed up in the .zip file but I do now make a .bak copy in the correct location.

bdleedy:
I think this is solid now. Uses the backup.ps1 even. Also checks for a new update.ps1 in the download and uses it.

kaikreuzer:
Thanks @bdleedy!
@tmrobert8 Could you do a quick final test?

tmrobert8:
I'm having all sorts of errors in this and I'm not sure what directory I should be running this from.

1) If I run from "c:\openhab2.2" (the main branch)....
The script runs fine until the very end - then it says:
```
Finished Download
===================================
New Update.ps1 found. Using that...
===================================
Update-openHAB : A parameter cannot be found that matches parameter name 'SkipNew'.
At C:\openhab2.2\runtime\bin\update.ps1:53 char:96
+ ... $OHDirectory -OHVersion $OHVersion -Snapshot $Snapshot -SkipNew $true
+                                                            ~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Update-openHAB], ParameterBindingException
    + FullyQualifiedErrorId : NamedParameterNotFound,Update-openHAB
```

2)  If I run it from the "c:\openhab2.2\runtime\bin" directory - I get
```
C:\openhab2.2\runtime\bin>powershell -command "& { . .\update.ps1; Update-openHAB -OHVersion 2.2.0 -Snapshot $true }"
Checking the specified openHAB directory...
.\userdata doesn't exist! Make sure you are in the openHAB directory or specify the -OHDirectory parameter!
At C:\openhab2.2\runtime\bin\update.ps1:105 char:13
+             throw "$OHDirectory\userdata doesn't exist! Make sure you ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (.\userdata does...tory parameter!:String) [], RuntimeException
    + FullyQualifiedErrorId : .\userdata doesn't exist! Make sure you are in the openHAB directory or specify the -OHD
   irectory parameter!```

3)  If I then take it's advise and use the OHDirectory parameter - I get (notice 'runtime/bin' is repeated because it's using the (pwd) rather than $OHDirectory):
```
C:\openhab2.2\runtime\bin>powershell -command "& { . .\update.ps1; Update-openHAB -OHDirectory c:\openhab2.2 -OHVersion 2.2.0 -Snapshot $true }"
Checking the specified openHAB directory...
Checking whether a service exists...
Checking whether openHAB is running...

Backup script starting...
Making a backup in 'c:\openhab2.2\backups' ...
. : The term 'C:\openhab2.2\runtime\bin\runtime\bin\backup.ps1' is not recognized as the name of a cmdlet, function,
script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is
correct and try again.
At C:\openhab2.2\runtime\bin\update.ps1:150 char:15
+             . $BackupScript
+               ~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\openhab2.2\r...\bin\backup.ps1:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

Backup-openHAB : The term 'Backup-openHAB' is not recognized as the name of a cmdlet, function, script file, or
operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try
again.
At C:\openhab2.2\runtime\bin\update.ps1:151 char:13
+             Backup-openHAB -OHDirectory (pwd)
+             ~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Backup-openHAB:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

Backup script finished.

Downloading the openHAB 2.2.0-SNAPSHOT distribution...
Downloaded 67390K of 67390K
Finished Download
===================================
New Update.ps1 found. Using that...
===================================
Update-openHAB : A parameter cannot be found that matches parameter name 'SkipNew'.
At C:\openhab2.2\runtime\bin\update.ps1:53 char:96
+ ... $OHDirectory -OHVersion $OHVersion -Snapshot $Snapshot -SkipNew $true
+                                                            ~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Update-openHAB], ParameterBindingException
    + FullyQualifiedErrorId : NamedParameterNotFound,Update-openHAB
```

**Comment by bdleedy** [comment]
@tmrobert8 Yeah, so that's the issue with testing this right now is that it doesn't support backwards compatibility because it will try to run that old one now. I built it with the thought that this is what's included in 2.2 and so it updates 2.2 to something else.

Both the new update.ps1 and backup.ps1 should be in \runtime\bin. I added the 'check' for the new update.ps1 because that's how the linux version does it. I have tried to keep this script as the 'Windows version' of it. I tested it by adding a pause right after the download and before the check, added the update script to the zip file, then resumed it.

**Comment by kaikreuzer** [comment]
What's the summary for me now? Shall I merge? It is the last chance now this afternoon. From what I see, it definitely improves the current situation, so I assume it should be merged as is for 2.2?

**Comment by BClark09** [comment]
I was able to test this too. I came to the same problems as @tmrobert8 did regarding paths:

image

The script does work, but only if you strictly go to the openHAB root directory AND use the `-OHDirectory .` parameter. There is the capacity for the upgrade to go wrong otherwise. The example above uses `SkipNew` to prevent the case where it's trying to access a new feature in an older zipfile, but if we *must* merge, then the user would have to do something as specific as:

```ps
cd C:\openhab
. .\runtime\bin\update.ps1
Update-openHAB 2.2.0 -OHDirectory .
```

kaikreuzer:
@bdleedy What's you advice then? We are just about to run the final release build now...

kaikreuzer:
> but if we must merge, then the user would have to do something as specific as

Doesn't sound very straight forward :-(

@BClark09 So better NOT merge now and keep a proper Windows update support for 2.3 instead?

kaikreuzer:
Hm, I am feeling a bit lonely now with that decision 😕

BClark09:
Hi Kai, unfortunately I've not been able to test further to see if there is any problem. But my gut feeling is that it is okay, as long as those instructions are clear for the release notes. I think with a good release cycle, the patch for it will not be not too long down the road anyway?

kaikreuzer:
Thanks @BClark09, I'll merge then.
Could you be so kind to fill in the according section in the release notes?
Yes, I hope that @pfink's work will enable us to reduce the time between releases in future, so improvements will get to the users more quickly!

kaikreuzer:
And of course: Many thanks @bdleedy for your work on this!

BClark09:
Done! Thanks @bdleedy, this is awesome work.

bdleedy:
Thanks guys, just building on a great start. I'll test with the parameters and see if I can figure out the directory issue. In my testing I didn't use them, just the defaults.

@kaikreuzer When(ish) can I expect the next release? I can work on a fix for this and get the restore script finished.

bdleedy:
@tmrobert8 @BClark09 Oh! It doesn't work because you're blindly giving a first parameter (without the - identifier) of a version but in reality it's expecting directory there. You have to feed arguments in order OR define them. The correct command line would be...

`cd C:\openhab
. .\runtime\bin\update.ps1
Update-openHAB -OHVersion 2.2.0.`

martinvw:
@bdleedy I updated the wiki can you and maybe @BClark09 validate it?

https://github.com/openhab/openhab-distro/wiki/Draft-for-openHAB-2.2-Release-Notes

tmrobert8:
@bdleedy 
I didn't have the same issue that @BClark09 was having (with the parameters).  The issue I was having is that if I run it *outside* of the openhab directory and specify the -OHDirectory parameter - it would blow up on the backup step because it's using (pwd) rather than -OHDirectory...

bdleedy:
@tmrobert8 Ah OK! I'll fix that. I also never updated the default OHVersion to 2.2.0... Doh.

@martinvw Looks good to me.

Issue #533, opened 2017-09-09
deb-package: etc/org.apache.karaf.shell.cfg gets overwritten each update


euphi:
Each time I update openhab2 (nightly build), the Karaf Shell configuration /var/lib/openhab2/etc/org.apache.karaf.shell.cfg gets overwritten.

As it contains the `sshHost` that is set to localhost by default, this stops external login.

(Luckily, sshd key configuration is kept, so after changing `sshHost` back to `0.0.0.0` everything is fine again.

BClark09:
Whilst this specific problem belongs at [https://github.com/openhab/openhab-linuxpkg](openhab/openhab-linuxpkg) `org.apache.karaf.shell.cfg` also gets replaced with the distro's upgrade scripts. @kaikreuzer, is this file safe to keep?

kaikreuzer:
Well, it is safe as long as we do not do any upgrade of Karaf, which changes the content of that file...
I am wondering if we should disallow manual editing of the karaf*.cfg files and rather ask users to put their settings into `services/runtime.cfg`.
@euphi Could you try to add
```
org.apache.karaf.shell:sshHost = <yourhost>
```
to that file and see if that does the trick?

euphi:
Yes, that did the trick. Many thanks for the hint.

I leave the issue open, because I suggest to update http://docs.openhab.org/administration/console.html accordingly.

ThomDietrich:
Sounds like an excellent solution. In the docs article we are suggesting to change sshHost and sshPort. Let's add commented settings lines to [runtime.cfg](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/conf/services/runtime.cfg) and change the docs article accordingly. Everyone aboard?

@euphi Would you like to add the PR for both?

euphi:
Its not that easy for the documentation, because
* I don't know the exact term for the "openHAB configuration directory". So I used "openHAB configuration directory, e.g. `/etc/openhab2/`"
* Other parameters are affected by the same issue (they may get overwritten, too).
* However, passwords/ssh-keys in the `user.properties` file seem to get not overwritten
  * The file name makes may assume that this is intented, however then this is inconsistent. **There should be only one base directory where users change their configuration.**

ThomDietrich:
Hey @euphi,

[...]

PR #537, opened 2017-09-17
#533: Add sshHost parameter to services/runtime.cfg

euphi:
As suggested in #533 I added the org.apache.karaf.shell:sshHost parameter to services/runtime.cfg with an explanation.


ThomDietrich:
Excellent writeup!

PR #543, opened 2017-09-25
Automatically calculate online.repo+pax url suffix+build number depending whether there is a releasebuild or not

pfink:
@kaikreuzer: Unfortunately, for those properties the existing property updating mechanism cannot be used as it just works for (already released) versions. To not have multiple property updating mechanisms, I'd suggest to automatically calculate those properties. "Release properties" will be used when providing `-Drelease` while building.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
         </snapshotRepository>
     </distributionManagement>
 
-    <properties>        
+    <properties>
         <online.repo.version>${parsedVersion.majorVersion}.${parsedVersion.minorVersion}</online.repo.version>        
 
         <ohc.version>2.2.0-SNAPSHOT</ohc.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -206,6 +213,32 @@
...
+        </profile>
+        <profile>
+            <id>releasebuild</id>
+            <activation>
+                <property>
+                    <name>release</name>
+                </property>
+            </activation>
+            <properties>
+                <build.number>- release build -</build.number>
```
kaikreuzer: @ThomDietrich What was our final decision on this String here? We wanted to use "Release Build" in capitals and without dashes, correct?
@pfink FYI, the current naming with dashes caused issues on openHABian, so @ThomDietrich asked me to change this in future - so this is now the right place to define it :-)

ThomDietrich: Release Build, correct.
@pfink the reason is simple, openHABian will show the version details on ssh login and the dashes were "ugly" while not having any purpose. Currently there is a simple search and replace logic in place:

grafik

pfink: Your wish is my command. Done ;-)

PR #569, opened 2017-11-18
Closes #568 Check if owner of start.sh matches owner of instance.prop…

### LINKED ISSUE #568: Check permissions on startup
A manual Java installation lead to some minor troubleshooting. For that I startet openHAP as root. Some files and directories are created on fist start so these got owned by root and the normal start of the service with user openhab failed.
Google showed some openHAB users struggeling with permissions and I guessed that starting openHAB with a wrong user could be a common pitfall if a package installation for Java AND openHAB is not possible or not working.

I suggest a simple check in start_debug.sh.

- Is this failure worth a check?
- Is the start skript the right place for this check?
- Is stderr the right target or should the error go somewhere else?
- Should the skript exit on failure? Exit code?
- Should this be ckecked in start.sh, too?

axxelG:
Check if owner of start.sh matches owner of instance.properties to verify that the first start of openHAB was done with the right user.

### DIFF
--- a/distributions/openhab/src/main/resources/home/start_debug.sh
+++ b/distributions/openhab/src/main/resources/home/start_debug.sh
@@ -1,4 +1,22 @@
 #!/bin/sh
 
 DIRNAME=`dirname "$0"`
+FILE_1=`basename "$0"`
+OWNER_1=$(ls -ld $FILE_1 | awk '{print $3}')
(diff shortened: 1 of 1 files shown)


kaikreuzer:
See my https://github.com/openhab/openhab-distro/issues/568#issuecomment-345475671

PR #594, opened 2017-12-03
increase thread pool sizes

kaikreuzer:
Seeing [errors like this](https://community.openhab.org/t/error-freezes-openhab/36839/9), I very much assume that our safe caller thread pool is far too small - increasing the pool size should hopefully resolve such issues.


kaikreuzer:
Something is weird here - why does it try to resolve org.openhab.binding.seneye? This does not even exist in our openhab2 repo...

I guess the PR can be safely merged anyhow.

Issue #562, opened 2017-11-01
Windows .bat files stroed in Unix format in ...\openhab-2.1.0\runtime\bin


WaldGrisBang:
The following *.bat are stored in unix format (LF only instead of CR/LF). Makes it a bit difficult to track down startup problems on a windows box.

C:\openHAB2\openhab-2.1.0\runtime\bin\setenv.bat
C:\openHAB2\openhab-2.1.0\runtime\bin\status.bat
C:\openHAB2\openhab-2.1.0\runtime\bin\stop.bat

bdleedy:
I'll see if I can get this fixed tonight.

kaikreuzer:
I briefly looked into it - most files are not from our repo, but created by the karaf packaging.

So what is needed is an adaption of the Maven assembly descriptor - just like we already [do it for the start.bat](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/descriptors/archive.xml#L80).

kaikreuzer:
#607 might solve this, but I am not sure, though (see my comment on it).

kaikreuzer:
Closing it as outdated and assuming it was fixed by #607.

PR #607, opened 2017-12-14
Use crlf line endings for bat files

kaikreuzer:
addresses #562

I would expect that this fixes the issue, but for some reason, trying it locally, I only see those endings on start.bat, but e.g. not on stop.bat. Maybe Jenkins or another Maven version is more successful, though and the change should do no harm.


kaikreuzer:
@martinvw Let me merge that even without review to test whether the Jenkins build result looks better!

PR #612, opened 2017-12-15
fix string in shell.init.script

sjka:
I really have no idea what I'm doing and how it is intended to be like, but counting the (unescaped) quotation marks I think karaf is completely right to complain about some string indices etc... 

fixes #570
fixes openhab/openhab-docker#128
Signed-off-by: Simon Kaufmann <simon.kfm@googlemail.com>

### DIFF
--- a/distributions/openhab/src/main/resources/userdata/etc/shell.init.script
+++ b/distributions/openhab/src/main/resources/userdata/etc/shell.init.script
@@ -43,10 +43,6 @@ __load_class = {
 // as they do occur if the session is headless / non-interactive
 jlineReader = $.reader
 if { %(jlineReader != null) } {
-
-  # On 256 colors terminal, add a right prompt
-  max_colors = ($.jline.terminal getNumericCapability max_colors)
-
   setopt group
   setopt auto-fresh-line
   unsetopt insert-tab
--- a/distributions/openhab/src/main/resources/userdata/etc/shell.init.script
+++ b/distributions/openhab/src/main/resources/userdata/etc/shell.init.script
@@ -47,7 +47,7 @@ if { %(jlineReader != null) } {
   # On 256 colors terminal, add a right prompt
   max_colors = ($.jline.terminal getNumericCapability max_colors)
   if { %(max_colors >= 256) } {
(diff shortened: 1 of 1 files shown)


kaikreuzer:
What do you mean by it fixes #570?
Note that the content of that file came directly from Karaf - so unless you also submit a fix there, we will see a regression after the next Karaf upgrade (which isn't far). Unfortunately, I didn't find that file anywhere in their repo, not sure where it comes from...

martinvw:
Does this literally come from karaf? As far as I remember it does, see https://github.com/apache/karaf/blob/master/assemblies/features/base/src/main/resources/resources/etc/shell.init.script

Should we create an upstream issue / pr and does it work better now? Does it also occur on a independant karaf?

kaikreuzer:
I've just checked: The latest Karaf 4.2.0.M1 still contains the very same string... (/cc @splatch)

kaikreuzer:
Ok, @martinvw was quicker :-)

kaikreuzer:
@sjka I am not sure that I like that change. Doing it on my installation results in:
```
Launching the openHAB runtime...

                          __  _____    ____      
  ____  ____  ___  ____  / / / /   |  / __ )     
 / __ \/ __ \/ _ \/ __ \/ /_/ / /| | / __  | 
/ /_/ / /_/ /  __/ / / / __  / ___ |/ /_/ /      
\____/ .___/\___/_/ /_/_/ /_/_/  |_/_____/     
    /_/                        2.2.0-SNAPSHOT
                               Build #1139   

Hit '<tab>' for a list of available commands
and '[cmd] --help' for help on a specific command.
Hit '<ctrl-d>' or type 'system:shutdown' or 'logout' to shutdown openHAB.

Error in initialization script: /Users/kai/Downloads/oh/userdata/etc/shell.init.script: null
```

sjka:
Didn't I mention that I have no idea what I'm doing here...? 

You don't need to like this change. If you know how to fix it please to so. In any case, this stupid like makes the docker instances go nuts. How about deleting these lines completely as the important ones down below are commented out anyway?

kaikreuzer:
That indeed works better for me.

sjka:
I removed the entire block, as it's useless anyway.

kaikreuzer:
Thanks! Please also file an issue upstream, so that it won't come back in future versions.

martinvw:
FTR: it was introduced in karaf in this commit

Issue #516, opened 2017-08-13
Split log configuration into system and user entries


kaikreuzer:
Currently, the logging configuration is stored in `userdata/etc/org.ops4j.pax.logging.cfg`. This file either gets overwritten by updates or not - but we do not have a chance to do adjustments in the system configuration without wiping any changes the user might have done.

As we have moved to log4j2 now, there is chance to address this: log4j2 supports [multiple configuration files](https://stackoverflow.com/a/38408283).
It would therefore be ideal to have the system config stored in `runtime/etc`, while the user config could reside in `conf` or `userdata/etc` (in case it is possible to edit the file e.g. through logging changes in the console).

cweitkamp:
@kaikreuzer Can you give any hint where to start looking to implement this feature?

kaikreuzer:
Not really. I guess 80% of the work for this issue is finding the right place and way how to implement it :-)
The best starting point for the investigation would imho be the [new feature in log4j2 itself](https://issues.apache.org/jira/browse/LOG4J2-494) and the [Karaf logging documentation](http://karaf.apache.org/manual/latest/#_log4j_v2_support) - this content about log4j2 wasn't there yet last week, so it must be pretty fresh :-)

We might think about moving from property files to XML files for the configuration as well (as it is also used in the docs now) - I somewhere read that XML includes should also work, so this might be another option for split files.

One thing to also keep in mind is how to deal with logging configuration changes through the Karaf shell - the `log:set` commands update the logging configuration files, which can be quite neat - but we must also ensure that nothing is written to conf and runtime as these folders are considered to be read-only.

Hope this helps to get an idea about the issue and where to look at!

cweitkamp:
I did some research. At the first try I got totally lost and gave up. A few days ago I coincidentally saw your comment in openhab/openhab-docs#476 containing the link to the new released documentation of [Karaf logging](http://karaf.apache.org/manual/latest/#_log). This manual is much better than the old version. Here are some facts I discovered so far. Now we have to put them together:

- The initial log configuration is always loaded from `etc/org.ops4j.pax.logging.cfg`
- You can add a log4j v2 XML-based configuration file by adding this line to the above file
```
org.ops4j.pax.logging.log4j2.config.file=${karaf.etc}/log4j2.xml
```
- XML configuration files can include other files with [XInclude](http://logging.apache.org/log4j/2.x/manual/configuration.html#XInclude)
[...]

PR #625, opened 2017-12-20
Update to Karaf 4.1.4

splatch:
This commit just lift up karaf version and removes overrides which are integrated into Karaf codebase.

### DIFF
--- a/features/overrides/src/main/feature/feature.xml
+++ b/features/overrides/src/main/feature/feature.xml
@@ -2,7 +2,6 @@
 <features name="${project.artifactId}-${project.version}" xmlns="http://karaf.apache.org/xmlns/features/v1.4.0">
 
   <feature name="openhab-overrides" description="force to bundle packages" version="${project.version}" hidden="true">
-    <bundle dependency="true">mvn:org.apache.karaf.deployer/deployer/4.1.3-sp1</bundle>
     <bundle dependency="true">mvn:org.jline/jline/3.5.1-sp1</bundle>
   </feature>
 
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
 
         <jackson.version>1.9.13</jackson.version>
 
-        <karaf.version>4.1.3</karaf.version>
+        <karaf.version>4.1.4</karaf.version>
         <jetty.version>9.3.22</jetty.version>
 
         <!-- provided by Karaf -->


kaikreuzer:
Did you check whether Karaf 4.1.4 brings any other changes in cfg or start files that would need to be sync'ed here as well?
Also note that the Karaf version update needs to be done in sync with https://github.com/openhab/openhab-core/blob/master/poms/pom.xml#L63.

Did you already do extensive testing to see whether there are any regressions?

Btw, as we are again have some time until a next release, I rather planned to directly move to Karaf 4.2. Do you see any reason against that?

splatch:
Coming back to this after a while. There are no major changes in Karaf 4.1.4 beside bugfixes, most important ones are confirmed to be solved. I haven't done any major testing of this.

On Karaf 4.2 will definitely bring some new semantics. There is major rework of features internals which will support feature level overrides. This will likely cause hiccups during update. Safest option would be to stay on 4.1.x untill 4.2 line gets first patch release.

kaikreuzer:
Ok, we could update to 4.1.4 to resolve some smaller issues, but what you write rather reassures me that we should asap do testing of 4.2.0.M2, so that any major regressions could actually be addressed in an 4.2.0 release. Waiting for a 4.2.1 just hoping that it'll be all fine and stable is probably pretty risky, don't you think?

kaikreuzer:
FTR, the reason why I would like to move to 4.2 soon is https://github.com/openhab/openhab-distro/issues/516, which is imho a pretty important feature for the future update process.

splatch:
Following upstream is best way to integrate early. If you are ok with some bumpy builds, that's the way to go.

Karaf 4.1.4 is already on Log4j2 via pax-logging 1.10.1 - meaning it shares same version of logging backend as 4.2 line and #516 can be tested even now - with 4.1.3 used in assembly.

kaikreuzer:
>  If you are ok with some bumpy builds, that's the way to go.

Better now than close to a release...

Note that #516 requires [KARAF-5354](https://issues.apache.org/jira/browse/KARAF-5354), which isn't part of Karaf 4.1.4, if I am not mistaken.

DanielMalmgren:
Just so that it's mentioned here as well: Quite many of us would be happy to get [this](https://issues.apache.org/jira/browse/KARAF-5473) fix in, which should happen in 4.1.5 and 4.2.0.

kaikreuzer:
Superseded by https://github.com/openhab/openhab-distro/pull/655.

PR #628, opened 2017-12-22
update scripts: replace org.apache.felix config files

ghys:
This includes userdata/etc/org.apache.felix* to the list of files not
preserved on update.

In particular, this makes sure #601 is applied when updating.

Signed-off-by: Yannick Schaus <habpanel@schaus.net>


kaikreuzer:
Sorry, never really noticed that PR - thanks for it!

PR #633, opened 2018-01-03
Windows start.bat enhancements

bcurran3:
I want start.bat to check the JAVA_HOME environmental variable and give an error if it's not found instead of just flashing a DOS box and going away.

I'm open to comments if it's the right place to do it or not. karaf.bat checks, but if JAVA_HOME is set incorrectly, I'm not sure it makes it that far in the batch files and it doesn't report an error.

I hope I did the PR correctly.

### DIFF
--- a/distributions/openhab-addons-legacy/pom.xml
+++ b/distributions/openhab-addons-legacy/pom.xml
@@ -47,13 +47,6 @@
                     </archive>
                 </configuration>
             </plugin>
-            <plugin>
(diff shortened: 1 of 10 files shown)


kaikreuzer:
Thanks @bcurran3, sounds like a useful enhancement.

> I hope I did the PR correctly.

Unfortunately not, there is no commit from yourself in this PR.

bcurran3:
will try to step through this again correctly later

PR #634, opened 2018-01-04
Add unicode file encoding jvm property

OzZyHH:
I just wan't to add a JVM property for loading config files with unicode encoding. This works well for me and some of my friends. You can see an example here: [Click](https://community.openhab.org/t/encoding-problem-solution-on-apt-repo-installation/34448)

I am not sure if it is a good idea for everyone. Maybe there are particular reasons to not set this property I am not aware off.

### DIFF
--- a/distributions/openhab/src/main/resources/bin/setenv
+++ b/distributions/openhab/src/main/resources/bin/setenv
@@ -96,7 +96,7 @@ export JAVA_OPTS="${JAVA_OPTS}
 # set JVM options
 #
 ARCH=`uname -m`
-EXTRA_JAVA_OPTS_COMMON="-Djava.awt.headless=true"
+EXTRA_JAVA_OPTS_COMMON="-Djava.awt.headless=true -Dfile.encoding=UTF-8"
 EXTRA_JAVA_OPTS_ARCH=""
 case "$ARCH" in
     *arm*) ;;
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Sorry for commenting so late.

> I am not sure if it is a good idea for everyone.

Me neither. According to https://github.com/eclipse/smarthome/pull/5206, ESH expects file to be in the "default platform encoding", so forcing it to UTF-8 might cause a bunch of problems for people that are e.g. on Windows (where UTF-8 is not the default).

I didn't fully understand what causes your issue. If I read it correctly, you say the manually installed tar version works, while an installation through apt doesn't. In both cases, you are using UTF-8 encoded files, right? You say

> On systemd service unit the terminal locale doesn’t matter.

which means that UTF-8 is not you default system locale, but just you user setting, right? If so, openHAB imho behaves correctly by using the system locale when being started as a system service. Imagine 3 users, each using a different local encoding for their files: How could openHAB deal with that? It stands no chance...

OzZyHH:
Hi Kai,
sorry for late reply. Yes its probably a bad idea  to change this in general. I changed my System locale with "localectl" command utiliy and it seems to this my issue (RPI3 with Raspbian Jessie). I am using the docker image for a few days now and the issue doesn't exist there too.

Closing this PR due to misconfiguration on user side.

PR #638, opened 2018-01-26
added -Djava.awt.headless=true to launch config

kaikreuzer:


kaikreuzer:
This should solve https://community.openhab.org/t/solution-found-ide-build-does-not-work-on-macbook-with-java-se-8-152-but-it-work-with-se9-but-it-works-on-windows-and-eclipse-home-also-works/37505/15

PR #678, opened 2018-04-07
Some improvements to demo sitemap and items

mueller-ma:
**Don't use groups in demo sitemap**

It still has the group demos as they are useless if they don't use
groups.

* Prevents people from changing the demo server.
* Change item labels to something usefull, e.g. "Bath" in subpage
    "Bath" to "Window" in subpage "Bath"
* Needed for localized sitemap

**Use wikipedia for webview**

* Better knows
* Not only in German
* No ads
* No tracker

**Use special icons for non-windows contacts** 
A door icon for a door is better suited than a window icon

**Use https where possible**

**Add label to all sitemap elements**

Needed for translation

### DIFF
--- a/features/distro-resources/src/main/resources/sitemaps/demo.sitemap
+++ b/features/distro-resources/src/main/resources/sitemaps/demo.sitemap
@@ -139,9 +139,9 @@ sitemap demo label="Main Menu"
 			}
 			Frame label="Multimedia Widgets" {
 				Image url="https://raw.github.com/wiki/openhab/openhab/images/features.png" label="openHAB" {
-					Text label="http://www.openHAB.org" icon="icon"
+					Text label="https://www.openHAB.org" icon="icon"
 				}
-				Video url="http://www.openhab.org/assets/smarthome.mp4"
+				Video url="https://www.openhab.org/assets/smarthome.mp4"
 				Webview url="https://en.m.wikipedia.org/w/index.php?title=Main_Page" height=8
 				Image url="http://fpoimg.com/320x240?text=openHAB" item=ImageURL
 			}
--- a/features/distro-resources/src/main/resources/sitemaps/demo.sitemap
+++ b/features/distro-resources/src/main/resources/sitemaps/demo.sitemap
@@ -1,91 +1,17 @@
 sitemap demo label="Main Menu"
 {
 	Frame {
-		Text label="First Floor" icon="firstfloor" {
-			Text label="Bathroom" icon="bath" {
-				Switch item=Light_FF_Bath_Ceiling label="Ceiling"
-				Switch item=Light_FF_Bath_Mirror label="Mirror"
-				Switch item=Heating_FF_Bath label="Heating"
-				Switch item=Shutter_FF_Bath label="Roller shutter"
-				Text item=Temperature_FF_Bath label="Temperature [%.1f °C]"
-				Text item=Window_FF_Bath label="Window [MAP(en.map):%s]"
-			}
-			Text label="Office" icon="office" {
-				Switch item=Light_FF_Office_Ceiling label="Ceiling"
-				Switch item=Heating_FF_Office label="Heating"
-				Switch item=Shutter_FF_Office_Window label="Window"
-				Switch item=Shutter_FF_Office_Door label="Door"
-				Text item=Temperature_FF_Office label="Temperature [%.1f °C]"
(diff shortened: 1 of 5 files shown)


features/distro-resources/src/main/resources/sitemaps/demo.sitemap
```diff
@@ -1,17 +1,91 @@
...
+				Switch item=Light_FF_Office_Ceiling label="Ceiling"
+				Switch item=Heating_FF_Office label="Heating"
+				Switch item=Shutter_FF_Office_Window label="Window"
+				Switch item=Shutter_FF_Office_Door label="Door"
+				Text item=Temperature_FF_Office label="Temperature [%.1f °C]"
+				Text item=Window_FF_Office_Window label="Window [MAP(en.map):%s]"
+				Text item=Window_FF_Office_Door label="Balcony Door [MAP(en.map):%s]"
+				Text item=Wifi_Level label="Wifi Level [%d/4]"
+			}
+			Text label="Child's Room" icon="boy_1" {
```
mueller-ma: I would like to change that to an actuall common English name and maybe add a daughters room, too.

kaikreuzer: sure, feel free.

martinvw:
> (1. serve as content for the demo server, 2. be an example how to design a sitemap for newbies, 3. cover varies features so that regressions are quickly noticed). Any suggestions welcome!

Maybe create a separate sitemap which we also host on the demo site but not as the default? Just focusing on every option available in small form.

mueller-ma:
> Maybe create a separate sitemap which we also host on the demo site but not as the default? Just focusing on every option available in small form.

That's also what I thought: I can create a new sitemap called "en_GB" with all the changes I made to the demo sitemap.

martinvw:
> That's also what I thought: I can create a new sitemap called "en_GB" with all the changes I made to the demo sitemap.

@kaikreuzer how do you feel about this solution, will this suffice in covering both scenario's?

kaikreuzer:
Yes, this solution should imho be ok.
@mueller-ma I assume that en_GB.sitemap is the starting point and we will soon see 100 further such translated sitemaps, right? Would it make sense to prefix them somehow, as I'd fear that otherwise the "demo.sitemap" will be drowning somewhere in the middle of the (alphabetically sorted) file system...

Are those international sitemaps only meant for the demo server or do you expect them to be installed with the "demo package" as well?

mueller-ma:
I can rename the demo sitemap to 0demo. And add some comments to en_gb that the same things can be achieved much simpler, e.g. with groups.

I'm not 100% sure since my last oh installation was some time ago, but I think there is already a german mappings file. These mapping files might be usefull for a normal user and I don't think it's worth the effort to filter the localiced sitemaps out.

mueller-ma:
I decided to add the prefix "lang_" to the localized sitemap.

kaikreuzer:
> I decided to add the prefix "lang_" to the localized sitemap.

👍. Better than "0demo"!

> These mapping files might be usefull for a normal user and I don't think it's worth the effort to filter the localiced sitemaps out.

Meaning you suggest to add the files to the demo package or not? I'd think it is better to leave them out (as it is now) as they'll rather lead to confusion for use case (2).

mueller-ma:
> I'd think it is better to leave them out (as it is now) as they'll rather lead to confusion for use case (2).

Well, that's true. So package the transform files and not the sitemap files?

kaikreuzer:
Yes. If you plan to localise the transform files, we might actually want to consider adding a bit more content and usage of them as well... But that can be done in a different PR.

yveshanoulle:
We probably need more experience with multi language, yet this solution still has the mindset of only one language per sitemap (so if you want to offer a multilangual version, you need to maintain two sitemaps.
Maybe that's enough, yet this means that everyone will have their own translation. In Belgium, we have 3 official languages. When using openhab in a company setting, we would need many sitemaps. I hope that European software will support more the easy swapping of languages.

kaikreuzer:
@yveshanoulle Sitemaps were always considered to be created for specific persons, so localisation was not really a concern here.

I see your point wrt bigger setups in companies or other places, though.
Note that the current sitemaps still come from the early openHAB 1 days - all the new features that are developed in ESH are usually designed to be easily localisable. You might be interested that the discussion and prototyping for a new sitemap concept has just started: https://github.com/eclipse/smarthome/issues/5337 - localisation options should be a requirement for it.

PR #682, opened 2018-04-16
Update addons.cfg

### LINKED ISSUE #680: Documentation improvement
The file addons.cfg contains some documentation that is confusing.

>'# A comma-separated list of bindings to install (e.g. "sonos,knx,zwave")
This is confusing, because it's not clear if the " are part of the list. 

A for me less confusing comment would be:

>'# A comma-separated list of bindings to install (e.g. "binding =  sonos,knx,zwave") 

(And similar for all lists in this file.) 

I don't know yet how to create PR for your project. I think we need an issue first.

martinvw:
Fixes #680 

As suggested by Yves Hanoulle (github: yveshanoulle)

Signed-off-by: Martin van Wingerden <martinvw@mtin.nl> (github: martinvw)

### DIFF
--- a/distributions/openhab/src/main/resources/conf/services/addons.cfg
+++ b/distributions/openhab/src/main/resources/conf/services/addons.cfg
@@ -28,23 +28,23 @@
 #
 #legacy = true
 
-# A comma-separated list of bindings to install (e.g. "sonos,knx,zwave")
+# A comma-separated list of bindings to install (e.g. "binding = sonos,knx,zwave")
 #binding = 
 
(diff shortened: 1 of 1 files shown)


yveshanoulle:
To test this, I "just" need to install the latest version? 

I assume this will overwrite my personal addons.cfg ?

yveshanoulle:
The commit looks like I would have written it.

yveshanoulle:
I just realized that Openhab uses nightly builds (or so I think) 
this means that tonight (let's say around 22:00 -CEST-) this won't be in a version yet.
Am I correct?

PR #683, opened 2018-04-17
Localize transform files via crowdin

mueller-ma:
The transform files can be directly used by crowdin. I removed the german translation as this should be managed via crowdin.

I added mappings for on and off as well. What else could we add?
The astro binding mappings from @ThomDietrich (https://community.openhab.org/t/astro-items-sitemap-map-binding-example/21155) might be a good idea.

Tested it here:
PR: https://github.com/mueller-ma-bot/openhab-distro/pull/1
Config (only the transform part): https://github.com/mueller-ma-bot/openhab-distro/blob/master/crowdin.yml
Crowdin: https://crowdin.com/translate/openhab-test/152/enus-de

### DIFF
--- a/features/distro-resources/pom.xml
+++ b/features/distro-resources/pom.xml
@@ -58,7 +58,6 @@
 								<artifact><file>src/main/resources/scripts/demo.script</file><type>cfg</type><classifier>scripts</classifier></artifact>
 								<artifact><file>src/main/resources/persistence/rrd4j.persist</file><type>cfg</type><classifier>persistence-rrd4j</classifier></artifact>
 								<artifact><file>src/main/resources/transform/en.map</file><type>cfg</type><classifier>transform-en</classifier></artifact>
-								<artifact><file>src/main/resources/transform/de.map</file><type>cfg</type><classifier>transform-de</classifier></artifact>
 								<artifact><file>src/main/resources/services/logging.cfg</file><type>cfg</type><classifier>services-logging</classifier></artifact>
 								<artifact><file>src/main/resources/services/basicui.cfg</file><type>cfg</type><classifier>services-basicui</classifier></artifact>
 								<artifact><file>src/main/resources/services/classicui.cfg</file><type>cfg</type><classifier>services-classicui</classifier></artifact>
--- a/features/distro-resources/src/main/resources/transform/de.map
+++ b/features/distro-resources/src/main/resources/transform/de.map
@@ -1,3 +1,0 @@
-CLOSED=zu
-OPEN=offen
-NULL=undefiniert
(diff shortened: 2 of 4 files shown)


martinvw:
This kills the build, when will you add the one translated by crowdin?

> 18:38:37.383  |  [ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:2.4:install (default-install) on project distro-resources: Failed to install artifact org.openhab.distro:distro-resources:cfg:transform-de:2.3.0-SNAPSHOT: /scratch/jenkins/workspace/PR-openHAB-Distribution/features/distro-resources/src/main/resources/transform/de.map (No such file or directory) -> [Help 1]

martinvw:
Can it be combined in this PR or can you have the other one ready so I can merge them more or less together?

mueller-ma:
Where can I see which scripts are executed by Jenkins? If it includes the german file, the script should be adapted to include all transform files before merging this.

martinvw:
https://github.com/openhab/openhab-distro/search?utf8=%E2%9C%93&q=%22de.map%22&type=

mueller-ma:
Is it possible to use wildcards in these xml files?

martinvw:
@pfink do you have some advice in this case?

kaikreuzer:
> Is it possible to use wildcards in these xml files?

Afaik, no. I don't have any good idea how to dynamically include files in the packaging...
I am also not too convinced anymore that we should push people to make use of MAP transform a lot - we have meanwhile a lot of other i18n features in ESH and with the Home Builder people can build a localised version of their files as a starting point.

martinvw:
>I am also not too convinced anymore that we should push people to make use of MAP transform a lot - we have meanwhile a lot of other i18n features in ESH and with the Home Builder people can build a localised version of their files as a starting point.

I do not understand or agree ;-)

I use the the MAP transform for two reasons:

* Having dutch labels (=i18n)
* But also mapping states of different types, I have a mapping for motion, coffeemachine, alarm, doorbell etc. So OFF can mean Unreachable, Ok or OFF And ON can mean OK, movement, ALARM!! and ON

mueller-ma:
> people to make use of MAP transform a lot

Map transforms are the only way I know to change "CLOSED" to "closed" on Text items representing a contact.

So rather remove the german transform file? Anyway we need those files to fully localize the demo sitemaps.

kaikreuzer:
> I do not understand or agree ;-)
> I use the the MAP transform for two reasons: ...

I didn't say that they aren't useful (at least as of today). I said that I am not "convinced anymore that we should push people to make use of" it. The reason being that MAP transform service is an optional component that needs to be installed and ALL of the transformation services are a very code-centric piece of "add-on", which does not fit anywhere in the architecture between bindings and UIs. There imho must be better ways to localise states than to use those MAP files.

At the moment, there are merely 3 entries in those demo files, which should show that it isn't an integral concept that everyone is expected to use - and I therefore would want to avoid that we start adding hundreds of strings in there and have them translated to 100 languages just to figure out that this isn't actually the best approach for doing state localization...

Issue #685, opened 2018-04-22
Include feature dependency to ESH USB-Serial discovery for ZigBee binding


hsudbrock:
Eclipse SmartHome got support for discovery of USB devices with a serial port (see https://github.com/eclipse/smarthome/pull/5315 and https://github.com/eclipse/smarthome/pull/5400). This feature is contained in the Karaf feature `esh-config-discovery-usbserial`.

The ZigBee binding wants to use this feature to discover ZigBee USB dongles (cf. https://github.com/openhab/org.openhab.binding.zigbee/pull/179), so the Karaf feature `openhab-binding-zigbee` needs a dependency to the new Eclipse SmartHome Karaf feature `esh-config-discovery-usbserial` (or to some feature depending on that feature).

PR #686, opened 2018-04-22
[685] Add dependency to feature esh-config-discovery-usbserial for ZigBee binding

hsudbrock:
This resolves https://github.com/openhab/openhab-distro/issues/685.

I was not completely sure whether the dependency to the Eclipse SmartHome feature should be directly added to the ZigBee binding feature, or whether this dependency should be added to some openhab feature (like, e.g., `openhab-runtime-base` or a new `openhab-usbserial-discovery` feature) in https://github.com/openhab/openhab-core/blob/master/features/openhab-core/src/main/feature/feature.xml.

I decided to add the direct dependency to the ESH feature here; if the other way would be the preferred one, I would be happy for that feedback and open a PR in the `openhab-core` repository for that.

### DIFF
--- a/features/addons/src/main/feature/feature.xml
+++ b/features/addons/src/main/feature/feature.xml
@@ -6,6 +6,7 @@
     <feature name="openhab-binding-zigbee" description="ZigBee Binding" version="${project.version}">
         <feature>openhab-runtime-base</feature>
         <feature>openhab-transport-serial</feature>
+        <feature>esh-config-discovery-usbserial</feature>
         <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee/1.0.8</bundle>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Thanks @hsudbrock!

> I was not completely sure whether the dependency to the Eclipse SmartHome feature should be directly added to the ZigBee binding feature, or whether this dependency should be added to some openhab feature

That's a very good remark. I guess my quick pointer to the Zigbee feature was actually a bad lead.
Contemplating about it, I came up with https://github.com/openhab/openhab-core/pull/337 instead - this solution only requires from bindings to state that they require serial-transport and they will have all features around serial support automatically available then. If no binding with serial access is installed, the `openhab-runtime-base` is still as small as possible and does not include any balast. I guess this is probably the best solution.

hsudbrock:
Thanks @kaikreuzer for the fix!

PR #693, opened 2018-05-01
Adding upgrade notes on compatibility break for PR (3264)

zarusz:
Updating notes to correspond with PR [3264](https://github.com/openhab/openhab2-addons/pull/3264)

Signed-off-by: Tomasz Maruszak <maruszaktomasz@gmail.com>

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -5,6 +5,7 @@ NOTE;*.rules files are now validated upon startup. Files with errors will from n
 
 [2.3.0]
 ALERT;Oceanic Binding: The 'softener' Thing Type no longer exists and is replaced by the 'serial' and 'ethernet' Thing Types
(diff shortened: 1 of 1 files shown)


martinvw:
@BClark09 is this the way it is intended or is this text maybe too long?

BClark09:
Ideally, the messages should be as short and as specific as possible. But I think we should handle the case for long messages anyway, it may be needed to give the user a good impression on what needs to be done.

Large terminals look okay IMO:
image

Narrow terminals start to get a bit hard to read:
image

I can also change the script to align the message if you believe this looks better:
image

martinvw:
Shall we leave away some details, what would you think of:

```Changed configuration keys. For main Thing keys changed from `HOST` to `host`, `PORT` to `port`. For zone Thing keys changed from `ZONE` to `zone`. ```
 
I think it is (or at least should) always the case that the documentation contains up-to-date examples. The (technical) reason why might be less relevant.

BClark09:
I think that would be better :)

zarusz:
I just pushed the changed message, which is what @martinvw suggested :-)

martinvw:
Thanks @zarusz

PR #717, opened 2018-05-24
updated links to documentation

kaikreuzer:


kaikreuzer:
You cannot be to quick, the new website just went live in case you haven't noticed :-)

martinvw:
> You cannot be to quick, the new website just went live in case you haven't noticed :-)

I indeed ended up there by accident, looks great!

PR #733, opened 2018-06-29
Add openHAB ZigBee binding to openHAB's oomph project setup

### LINKED ISSUE #732: Add openHAB ZigBee binding to oomph project setup file
Currently, when installing the Eclipse IDE via the oomph installer using the openhab project setup (defined [here](https://github.com/openhab/openhab-distro/blob/master/launch/openHAB2.setup)), many openHAB bindings can be included in the IDE, but the openHAB ZigBee binding cannot.

The reason is that the openHAB ZigBee binding is not in the [openHAB2 addons repository](https://github.com/openhab/openhab2-addonsl), but in its own repository.

I think that it would be nice to include the project for the openHAB ZigBee binding in the project setup file as well.

hsudbrock:
Resolves #732

This PR adds a project for the openHAB ZigBee binding to the project setup file `openHAB2.setup` (in the same way that it is done for Eclipse Smarthome, openhab2 addons, openhab1 addons, etc.).

The PR also contains a small fix for the dynamic working set for the openhab1 addons, which contained a wrong (probably outdated) path regex.

I tested this with an installation of the new Eclipse Photon release with openhab projects selected in the install process, and it looked good to me.

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2402,7 +2402,7 @@
           name="OH1 Add-ons">
         <predicate
             xsi:type="predicates:LocationPredicate"
-            pattern=".*/openhab/bundles/.*"/>
+            pattern=".*/openhab1-addons/bundles/.*"/>
       </workingSet>
     </setupTask>
     <setupTask
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2430,37 +2430,92 @@
     <stream
         name="master"/>
   </project>
-  <project
-        name="4_zigbee"
-        label="openHAB ZigBee binding">
-        <setupTask
-          xsi:type="git:GitCloneTask"
-          id="git.clone.zigbeebinding"
-          remoteURI="https://github.com/openhab/org.openhab.binding.zigbee.git"
-          pushURI="https://github.com/${github.user.id}/org.openhab.binding.zigbee.git">
-        <description>openHAB ZigBee binding</description>
-      </setupTask>
-      <setupTask
-          xsi:type="setup.workingsets:WorkingSetTask">
-        <workingSet
-            name="openHAB ZigBee binding"
-            id="">
-          <predicate
-              xsi:type="predicates:LocationPredicate"
-              pattern=".*/org.openhab.binding.zigbee/.*"/>
-        </workingSet>
-      </setupTask>
-      <setupTask
-          xsi:type="projects:ProjectsImportTask">
-        <sourceLocator
-            rootFolder="${git.clone.zigbeebinding.location}"
-            locateNestedProjects="true"/>
-      </setupTask>
-      <stream
-          name="master"
-          label=""/>
+  <project name="4_zigbee"
+      label="openHAB ZigBee binding">
+    <setupTask
+        xsi:type="git:GitCloneTask"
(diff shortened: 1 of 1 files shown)


kaikreuzer:
> Resolves #732

No, it doesn't resolve https://github.com/openhab/openhab-distro/issues/732#issuecomment-401377512 :-)

hsudbrock:
Hey ;)

I'll have a look!

hsudbrock:
I have integrated the zwave binding as well as habpanel and habmin into the oomph setup model.

For https://github.com/openhab/org.openhab.binding.bacnet/, there is no eclipse project file available, and using a maven project import in Eclipse IDE does not yield an Eclipse project structure that I am satisfied with... I will try to get a `.project` file in the https://github.com/openhab/org.openhab.binding.bacnet/ repo first...

kaikreuzer:
Cool, many thanks!
As https://github.com/openhab/org.openhab.binding.bacnet/pull/1 is merged now, I will wait for your update on this PR before merging.

hsudbrock:
I have added the BACNet binding now as well.

The result looks as follows on my machine:

In the oomph installer:

project-list-in-oomph-installer

In the package explorer in the created workspace:

dynamic-workingsets

I'd like to mention two things, where the changes made in this PR somewhat differ from the previous oomph setup model:

1. Before the changes from this PR, the oomph  setup model provided only options for adding comparatively large sets of projects to the Eclipse workspace, like, e.g., "all openHAB 1 addons", "all openHAB 2 addons", "all ESH extensions", or "all ESH and OH2 core bundles". With this PR, there are new options that add only few new projects, like, e.g. "the single Eclipse project for the Z-Wave binding", or "the 5 Eclipse projects for the ZigBee binding".

2. The same difference is also reflected in the dynamic working sets in the generated Eclipse workspace (i.e., only working sets containing many projects were added before this PR; working sets containing few projects are introduced by this PR).

This "difference" occurs, because the git repositories of the newly included openHAB projects contain only few Eclipse projects, and I decided to go along with using one checkbox in the installer / one dynamic working set per git repository.

If there is a better structure that I just do not see, this should be easy to integrate.

launch/openHAB2.setup
```diff
@@ -2430,7 +2430,157 @@
     <stream
         name="master"/>
   </project>
-  <project name="4_esh"
+  <project name="4_zigbee"
+      label="openHAB ZigBee binding">
```
kaikreuzer: Could you change all occurrences of "binding" to "binding" to be in sync with the other texts (where we also have e.g. "Add-ons")?

kaikreuzer:
@hsudbrock Can you tell what error the HABmin project shows in the UI? @cdjackson might want to fix it :-)

cdjackson:
It’s missing the optional null annotations import which are used by some dependancies. I’ll update...

launch/openHAB2.setup
hsudbrock: From "binding" to "binding" would be a no-op ;) I guess you mean the capitalization, as this is the inconsistency I could spot; updated the PR by capitalizing "binding" for both installer labels and working set labels.

kaikreuzer: You are certainly very right 🤣

Issue #716, opened 2018-05-23
New Update.ps1 found. Using that... double -SNAPSHOT appended


eusblh:
I'm running windows. openhab-2.2.0 stable
I wanted to update to openhab-2.3.0-SNAPSHOT using update.ps1

Update-openHAB -OHVersion 2.3.0 -Snapshot $true
(excerpts from logfile included) 
Downloading the openHAB 2.3.0-SNAPSHOT distribution...
Downloaded 63603K of 63603K
Finished Download
New Update.ps1 found. Using that...

Here's where the trouble starts. It appears that on the second run of update.ps1, it tries to download the snapshot again, but it duplicates/appends the -SNAPSHOT onto the end again:
Downloading the openHAB 2.3.0-SNAPSHOT-SNAPSHOT distribution...
Exception calling "GetResponse" with "0" argument(s): "The remote server returned an error: (404) Not Found."

also interesting, there seems to be no error catching. It thinks everything went well, although I now don't even have a runtime folder:

openHAB updated to version 2.3.0-SNAPSHOT-SNAPSHOT!

as a workaround, this worked:
Update-openHAB -OHVersion 2.3.0 -Snapshot $true -SkipNew $true
 2.3.0-SNAPSHOT Build #1285

wborn:
Was this issue also fixed in #809 @tmrobert8?

wborn:
I see it had it's own fix in https://github.com/openhab/openhab-distro/pull/736 :wink:

PR #736, opened 2018-07-17
Fix for : New Update.ps1 found. Using that... double -SNAPSHOT append…

oliviervanroey:
…ed #716

When the boolean parameter $snapshot was set to true, it will check whether the version contains the suffix "-SNAPSHOT" . It will only be added if true...

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.ps1
+++ b/distributions/openhab/src/main/resources/bin/update.ps1
@@ -167,7 +167,7 @@ Function Update-openHAB {
         # Download the selected openHAB version
         # Choose bintray for releases, cloudbees for snapshot.
         if ($Snapshot) {
-            $OHVersion = "$OHVersion-SNAPSHOT"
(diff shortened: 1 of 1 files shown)


martinvw:
@bdleedy could you a quick review, its seems sane to me :-)

martinvw:
@BClark09 maybe you?

BClark09:
I'm not able to test the line, but the logic and reasoning looks good to me! :)

eusblh:
@bdleedy see reference above to "Fix for : New Update.ps1 found. Using that... double -SNAPSHOT appended"

eusblh:
https://github.com/openhab/openhab-distro/issues/716

oliviervanroey:
it's not really a bug but it will increase the usability of the code... it will correct a snapshot version when only the version is given. otherwise you could opt for removing the boolean $snapshot parameter and define it by the naming convention...

tmrobert8:
FYI - I just tested this change on my system and it works.

PR #743, opened 2018-07-27
Created a demo.py to demonstrate a cron trigger jython rule

openhab-5iver:
This adds a demo Jython rule.

Signed-off-by: Scott Rushworth <openhab@5iver.com> (github: openahb-5iver)


kaikreuzer:
Hm, this will fail, if Jython isn't installed, won't it? Or is it simply ignored in that case?

openhab-5iver:
I just tested with a clean manual install, Rule Engine (Experimental), and both demo scripts. The JS ran but the Jython did not. No errors in the log. So, I don't think it will hurt, but iit's not going to help much either!

PR #747, opened 2018-08-16
Define astro items with dimension

kaikreuzer:


kaikreuzer:
Related to https://github.com/eclipse/smarthome/issues/6055

Issue #755, opened 2018-08-29
Add option for smaller backups using openhab-cli


DanielMalmgren:
Somewhere between 2.2 and 2.3 the default behaviour of "openhab-cli backup" was changed as to include tmp and cache directories, which makes the backup file a multitude bigger (mine went from 300kB to 90MB). I suggest the behaviour is changed back to the original with an optional argument for those who also want the tmp and cache directories.

I have added this on my installation, attaching diff files for openhab-cli and backup. This adds a --full argument to the commands.

Also see discussion here: [https://community.openhab.org/t/openhab-cli-backup-from-700kb-to-85mb/48266/1](url)

[backup.diff.txt](https://github.com/openhab/openhab-distro/files/2332084/backup.diff.txt)
[openhab-cli.diff.txt](https://github.com/openhab/openhab-distro/files/2332085/openhab-cli.diff.txt)

BClark09:
Just a note for this issue, the `openhab-cli` command is for linuxpkg installs only. But the issue exists in the `$OPENHAB_HOME/bin/backup` script, which will effect Mac and manual installs too.

DanielMalmgren:
Ok, I didn't know that. The changes I've done to openhab-cli is mostly aesthetic.

I'd also like to note that my current implementation is far from optimal. It still copies the entire userdata folder to the temporary directory and then deletes tmp and cache from there which obviously is loads of megabytes unnecessarily copied. The best thing would be to not include them from the beginning, but cp doesn't have any way of excluding files. I guess that can be solved by replacing the cp (on line 118) with something else. I'll get back if I find anything good.

Dan1001:
rsync has an exclude option, but would have the undesirable effect of adding another dependency.

DanielMalmgren:
Yep, we like to stick with things that we know are always there. I made a version now that fixes this with help of find, sed and xargs, I guess those are standard commands? Attaching a new diff 😄 

[backup.diff.txt](https://github.com/openhab/openhab-distro/files/2335065/backup.diff.txt)

DanielMalmgren:
...except that I forgot to clean up my test stuff. There obviously shouldn't be a -v on the cp command on line 119. I will keep that as an exercise for the reader to remove...

BClark09:
When running the command I got the directories and files of `/userdata/config` inside `$TempDir/userdata`, but looking at the line, I'm not sure how this is possible?

BClark09:
[...]

PR #756, opened 2018-09-04
Make backup without tmp and cache

DanielMalmgren:
Changes the backup script so that it doesn't include tmp and cache directories. Add optional --full argument for those who want to include it.

This makes the default behaviour go back to how it worked in 2.2. For further details see https://community.openhab.org/t/openhab-cli-backup-from-700kb-to-85mb/48266 and https://github.com/openhab/openhab-distro/issues/755

### DIFF
--- a/distributions/openhab/src/main/resources/bin/backup
+++ b/distributions/openhab/src/main/resources/bin/backup
@@ -16,7 +16,7 @@ setup(){
     echo ""
     echo "  e.g. ./backup                     << Makes a file with a timed filename"
     echo "       ./backup myBackup.zip        << Makes a file called myBackup.zip"
-    echo "  e.g. ./backup --full              << Makes a full backup file with a timed filename"
+    echo "       ./backup --full              << Makes a full backup file with a timed filename"
     echo "       ./backup --full myBackup.zip << Makes a full backup file called myBackup.zip"
     echo ""
     echo "Use this script to backup your openHAB configuration, you can use the 'restore' script"
@@ -68,7 +68,7 @@ setup(){
 
   timestamp=$(date +"%y_%m_%d-%H_%M_%S")
   ## Set the filename
-  if [ -z "$1" ] || [ "$1" = "--full" ]; then
+  if [ -z "$1" ]; then
     echo "Using '$OPENHAB_BACKUPS' as backup folder..."
     OutputFile="$OPENHAB_BACKUPS/openhab2-backup-$timestamp.zip"
   else
--- a/distributions/openhab/src/main/resources/bin/backup
+++ b/distributions/openhab/src/main/resources/bin/backup
@@ -12,13 +12,17 @@ getFullPath() {
 
 setup(){
   if [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
-    echo "Usage: backup [filename]"
+    echo "Usage: backup [--full] [filename]"
     echo ""
-    echo "  e.g. ./backup                 << Makes a file with a timed filename"
-    echo "       ./backup myBackup.zip    << Makes a file called myBackup.zip"
+    echo "  e.g. ./backup                     << Makes a file with a timed filename"
+    echo "       ./backup myBackup.zip        << Makes a file called myBackup.zip"
+    echo "  e.g. ./backup --full              << Makes a full backup file with a timed filename"
+    echo "       ./backup --full myBackup.zip << Makes a full backup file called myBackup.zip"
     echo ""
     echo "Use this script to backup your openHAB configuration, you can use the 'restore' script"
     echo "from any machine to transfer your configuration across to another instance."
+    echo ""
+    echo "A full backup includes the tmp and cache directories that are normally excluded."
     echo ""
     echo "Set $OPENHAB_BACKUPS to change the default backup directory."
     exit 0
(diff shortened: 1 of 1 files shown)


DanielMalmgren:
Ok, my first PR ever. I've always thought git is kinda confusing (always liked svn better), so bare with me if this isn't properly made...

distributions/openhab/src/main/resources/bin/backup
```diff
@@ -55,9 +55,15 @@ setup(){
 
   fileList="$OPENHAB_RUNTIME/bin/userdata_sysfiles.lst"
 
+  ## Parse arguments
+  if [ "$1" = "--full" ] || [ "$2" = "--full" ]; then
```
BClark09: I think it would be safer to request that the "--full" parameter is done before the listed directory. This way, `--full` will always be stored in `$1`. Inside this function, you can add the shift command, now `$1` thereafter will always contain the full path as expected by the rest of the script.

distributions/openhab/src/main/resources/bin/backup
```diff
@@ -55,9 +55,15 @@ setup(){
```
BClark09: Don't forget to add your new feature in the `--help` documentation above (Lines 17-18) ;)

distributions/openhab/src/main/resources/bin/backup
```diff
@@ -55,9 +55,15 @@ setup(){
...
+  ## Parse arguments
+  if [ "$1" = "--full" ] || [ "$2" = "--full" ]; then
+    echo "including cache"
+    INCLUDE_CACHE="true"
+  fi
+
   timestamp=$(date +"%y_%m_%d-%H_%M_%S")
   ## Set the filename
-  if [ -z "$1" ]; then
+  if [ -z "$1" ] || [ "$1" = "--full" ]; then
```
BClark09: This can stay as it was if the above comment is taken onboard. Otherwise, `./runtime/bin/backup --full /my/directory` will not do as a user expects.

distributions/openhab/src/main/resources/bin/backup
DanielMalmgren: Hmmm... I actually put it as the first argument first... But then I looked at the other arguments to openhab-cli and noticed that the only other command that has a "--"-argument (start --debug) has it last so that felt more consistent. But I can change it back again ;-)

distributions/openhab/src/main/resources/bin/backup
DanielMalmgren: Oh. Never noticed there was a built in documentation. Fixing.

DanielMalmgren:
Oh crap. Now I just saved my changes in my repo and suddenly they became a new commit in this repo. Why did that happen? Anyway I guess I was supposed to put the sign-off-stuff in the comment which I didn't do now. Apart from that, the comments above should now be fixed...

distributions/openhab/src/main/resources/bin/backup
```diff
@@ -55,9 +59,16 @@ setup(){
...
+  if [ "$1" = "--full" ]; then
+    echo "including cache"
+    INCLUDE_CACHE="true"
+    shift
+  fi
+
   timestamp=$(date +"%y_%m_%d-%H_%M_%S")
   ## Set the filename
-  if [ -z "$1" ]; then
+  if [ -z "$1" ] || [ "$1" = "--full" ]; then
```
BClark09: With the latest changes, the addition here doesn't do anything. Probably best to keep it as:

```
if [ -z "$1" ]; then
```

Do you agree?

distributions/openhab/src/main/resources/bin/backup
```diff
@@ -12,14 +12,18 @@ getFullPath() {
...
 setup(){
   if [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
-    echo "Usage: backup [filename]"
+    echo "Usage: backup [--full] [filename]"
     echo ""
-    echo "  e.g. ./backup                 << Makes a file with a timed filename"
-    echo "       ./backup myBackup.zip    << Makes a file called myBackup.zip"
+    echo "  e.g. ./backup                     << Makes a file with a timed filename"
+    echo "       ./backup myBackup.zip        << Makes a file called myBackup.zip"
+    echo "  e.g. ./backup --full              << Makes a full backup file with a timed filename"
```
BClark09: Formatting comment: need for the 2nd "e.g." here.

distributions/openhab/src/main/resources/bin/backup
DanielMalmgren: Thought I changed this. How strange...

distributions/openhab/src/main/resources/bin/backup
DanielMalmgren: Yep, well spotted :-)

DanielMalmgren:
Any chance on getting this approved and committed before next milestone? @kaikreuzer ?

kaikreuzer:
> @kaikreuzer would you be able to do this?

Yes, works as expected and makes a lot of sense! 👍 

> Any chance on getting this approved and committed before next milestone?

Yes, sorry for my very late response, let's get that in M5!

PR #758, opened 2018-09-07
Fix wrong variable name: `OPENHAB_BACKUP`

reitermarkus:
Signed-off-by: Markus Reiter <me@reitermark.us>


cweitkamp:
Nice catch, thanks.

PR #763, opened 2018-09-12
Override default console highlighting colors

BClark09:
## PuTTY

image

## Powershell.exe

image

## Debian terminal

image

## Debian terminal (Black on white scheme)

image

### DIFF
--- a/distributions/openhab/src/main/resources/userdata/etc/scripts/shell.colors.script
+++ b/distributions/openhab/src/main/resources/userdata/etc/scripts/shell.colors.script
@@ -1,0 +1,40 @@
+// This script overrides the default colors for the Karaf console.
+// All colors are set using the original ANSI escape code specification.
+
+// HIGHLIGHTER_COLORS overrides the colors related to console input.
+// From the Apache Felix documentation:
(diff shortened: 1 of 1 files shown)


kaikreuzer:
I like it - let's wait for some other opinions, though :-)

BClark09:
Just for info, this is what PuTTY looks like without the change:

image

A couple of thoughts,

1. variable names should be in the system default font color, it's a bit hard to read on the light background.

2. The original colors had strings and numbers the same color, we could probably keep this as an option.

WDYT?

martinvw:
I only think that the dark green '0' in the example is not that readable

kaikreuzer:
What is the difference between "123" and "0" so that the first is light and the second dark green?

BClark09:
I believe the '0' at the end is something that Karaf adds at the end as a suggestion to close off the command. It is set using the 're' key. (for Repair). It will also add a quote as a suggestion to end an open quote. I can change this to yellow?

CWempe:
Looks good to me.

I think we should not change to much.

martinvw:
Okay thanks guys!

PR #767, opened 2018-09-17
Update for ZWave breaking changes

cdjackson:


distributions/openhab/src/main/resources/bin/update.lst
```diff
@@ -12,6 +12,7 @@ ALERT;Yamaha Receiver Binding: The configuration parameter names now use lower c
 ALERT;Astro Binding: The 'kilometer' and 'miles' channels have been replaced by a new 'distance' channel
 ALERT;Jeelink Binding: The 'currentWatt' and 'maxWatt' channels have been replaced with 'currentPower' and 'maxPower' channels
 ALERT;WeatherUnderground Binding: A bridge has been added on top of the current things, you need to add a bridge containing your api-key.
+ALERT;ZWave Binding: Major changes have been merged to support features such as security. All things must be deleted and re-added.
```
BClark09: If I understood your post on the forum correctly, should this be "All things (except the controller) must be deleted..."?

martinvw: Maybe we could change it to something like: "To be on the safe side, all things should be deleted and re-added" Or would that be to vague?

BClark09:
It'll have to be one line on the `update.lst` file, but any terminal should wrap the text if it's any size.

image

Would linking to your community thread be a good idea? e.g. what you already have, but with "see https://community.openhab.org/t/zwave-binding-migration-to-development-version for more details."

PR #771, opened 2018-09-22
invalid configuration for pool 'safeCall'

martinvw:
When running locally I get an error:

`Ignoring invalid configuration for pool 'safeCall': 3,10 - value must be an integer`

This fixes it, do you also see this error, is this new / expected?

### DIFF
--- a/launch/home/etc/services.cfg
+++ b/launch/home/etc/services.cfg
@@ -34,6 +34,6 @@ org.eclipse.smarthome.threadpool:thingHandler=3
 org.eclipse.smarthome.threadpool:discovery=3
 
 # Non-scheduled thread pools can also provide a max size
-org.eclipse.smarthome.threadpool:safeCall=3,10
+org.eclipse.smarthome.threadpool:safeCall=3
 
 org.eclipse.smarthome.autoupdate:sendOptimisticUpdates=true
--- a/launch/home/etc/services.cfg
+++ b/launch/home/etc/services.cfg
@@ -30,10 +30,10 @@ com.eclipsesource.jaxrs.swagger.config:swagger.basePath=/rest
 com.eclipsesource.jaxrs.swagger.config:swagger.info.title=openHAB REST API
 
 # Configuration of scheduled thread pool sizes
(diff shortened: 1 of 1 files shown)


launch/home/etc/services.cfg
```diff
@@ -34,6 +34,6 @@ org.eclipse.smarthome.threadpool:thingHandler=3
 org.eclipse.smarthome.threadpool:discovery=3
 
 # Non-scheduled thread pools can also provide a max size
-org.eclipse.smarthome.threadpool:safeCall=3,10
+org.eclipse.smarthome.threadpool:safeCall=3
```
wborn: I think 10 would be a better value because that's what is also used when normally launching openHAB:

https://github.com/openhab/openhab-distro/blob/df94f5b3d6bf5bd09dbaeade55ee8e5cea29a2cc/distributions/openhab/src/main/resources/runtime/services.cfg#L14

wborn:
> This fixes it, do you also see this error, is this new / expected?

I think it's not new. Just a left over from https://github.com/eclipse/smarthome/pull/1744 that has gone unnoticed so far. :-)

launch/home/etc/services.cfg
kaikreuzer: I agree. `3,10` was an old format to say "at least 3, max 10". Now the value is simply the max value, so it should be 10.

wborn: While we're at it we might also want to sync the other threadpool values from 3 to 5?

https://github.com/openhab/openhab-distro/blob/df94f5b3d6bf5bd09dbaeade55ee8e5cea29a2cc/distributions/openhab/src/main/resources/runtime/services.cfg#L11-L14

kaikreuzer: why not...

martinvw:
I have busy dat/week ahead feel free to amend to my PR 👍

Op zo 23 sep. 2018 om 10:43 schreef Kai Kreuzer <notifications@github.com>

kaikreuzer:
Updated it - @wborn for you to merge then :-)

PR #775, opened 2018-09-23
Bump ZigBee libraries to 1.1.2

cdjackson:
I hope I've got this right this time @kaikreuzer ;)

### DIFF
--- a/features/addons/src/main/feature/feature.xml
+++ b/features/addons/src/main/feature/feature.xml
@@ -6,11 +6,11 @@
     <feature name="openhab-binding-zigbee" description="ZigBee Binding" version="${project.version}">
         <feature>openhab-runtime-base</feature>
         <feature>openhab-transport-serial</feature>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee/1.1.1</bundle>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.dongle.cc2531/1.1.1</bundle>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.dongle.ember/1.1.1</bundle>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.dongle.telegesis/1.1.1</bundle>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.dongle.xbee/1.1.1</bundle>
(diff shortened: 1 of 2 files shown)


kaikreuzer:
> I hope I've got this right this time @kaikreuzer ;)

Almost, what you have missed was https://github.com/openhab/openhab-core/pull/406.

I must say that I am a bit puzzled, though. At the time when we agreed to add those libs to the target platform, we did that under the assumption that those are stable (meaning there might be one or two updated required per year).
It now seems as if the update frequency heavily increases and the library is not as stable as it seemed, but rather under active development - and with more activity on the binding, it is likely that it continues that way. We are now roughly at an update per month, for which the target platform mechanism isn't really suitable anymore. I think we should really re-discuss, how to handle this code. I still think that https://github.com/openhab/zigbee was an excellent compromise for it, at least for the time while there are many changes that are driven by the binding and which need to go hand in hand. Once the code is at Eclipse, it would anyhow be a different situation.

kaikreuzer:
retriggering Jenkins...

cdjackson:
Thanks - sorry for missing this.

As said, moving the repository to an openHAB repository is not something that is suitable and can't be accepted as the libraries are used by many people. This will not be something that is possible - sorry.If it's a problem to update the TP, then I'm happy to look at alternatives if you want to propose something.

kaikreuzer:
> moving the repository to an openHAB repository is not something that is suitable and can't be accepted as the libraries are used by many people

Where's the exact difference to moving it to Eclipse? That it is better recognised as a separate project?

tomhoefer:
Actually I can´t remember how often we have discussed this topic. Today I finally realized that we will never achieve a common agreement.

cdjackson:
I’m not sure I understand your point Thomas - we agreed about 6 months ago that the libraries can not be put under openHAB.

Regarding the frequency of update - when we discussed this back in April time, I think we said it would be updated every 2 or 3 months. I have since then performed 3 updates in 6 months (July, and September) - yes, I agree that the last 2 are only 2 weeks apart and that’s unfortunate but was required in order to correct a bug that was introduced. However I really don’t think that we are doing anything too different than anticipated, and I don’t think you can really say that the update rate has “heavily increased” ;)

kaikreuzer:
> I’m not sure I understand your point Thomas - we agreed about 6 months ago that the libraries can not be put under openHAB.

I do not understand it either - yes, we agreed to keep it for the moment where it is and work with the current setup and I think we are all aligned with the goal to have the code on a legally sound basis by moving it to Eclipse. It is imho really time to get that started as legal clarifications tend to last looooong... 😟

tomhoefer:
Nevermind ... we have to find a suitable solution

PR #780, opened 2018-10-05
Enable backup to set temporary directory via env

martament:
Default tmp is on /tmp. /tmp is a tmpfs in RAM and if your backup is too big for /tmp (> 450MB) you can't backup your files. With this enhancement it is possible to set /tmp to another location and the backup will work.
Example: export TempDir=/opt/temp

### DIFF
--- a/distributions/openhab/src/main/resources/bin/backup
+++ b/distributions/openhab/src/main/resources/bin/backup
@@ -21,6 +21,7 @@ setup(){
     echo "from any machine to transfer your configuration across to another instance."
     echo ""
     echo "Set $OPENHAB_BACKUPS to change the default backup directory."
+    echo "Set $OPENHAB_BACKUPS_TEMP to change the default backup temporary directory."
     exit 0
   fi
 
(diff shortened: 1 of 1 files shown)


maggu2810:
Environment variables are (mostly) all uppercase (AFAIK).

martament:
You are right.

I can use:
OPENHAB_BACKUP_TEMP for environment and map it internal to $TempDir

distributions/openhab/src/main/resources/bin/backup
```diff
@@ -79,12 +80,20 @@ setup(){
   CurrentVersion="$(awk '/openhab-distro/{print $3}' "$OPENHAB_USERDATA/etc/version.properties")"
 
   ## Store anything in temporary folders
-  TempDir="/tmp/openhab2/backup"
-  echo "Making Temporary Directory"
-  mkdir -p "$TempDir" || {
+  # Check if we should use TempDir from env $OPENHAB_BACKUPS_TEMP
+  if [ ${OPENHAB_BACKUPS_TEMP} ];  then
```
BClark09: This would be better as 

```
if [ -n "${OPENHAB_BACKUPS_TEMP}" ]; then
```

martament: I changed the patch as you suggested. Do i have to do something else?

kaikreuzer:
@martament One wish: Could you please squash your commits into one and make sure that the resulting commit has a [signed-off-by footer](https://github.com/openhab/openhab-distro/blob/master/CONTRIBUTING.md#sign-your-work)? Thanks!

martament:
@kaikreuzer 
I've done as you suggested. Hope it's alright.

kaikreuzer:
All perfect, many thanks!

PR #781, opened 2018-10-09
added new ESH MQTT binding to distro

kaikreuzer:

### DIFF
--- a/features/addons-esh/src/main/feature/feature.xml
+++ b/features/addons-esh/src/main/feature/feature.xml
@@ -47,6 +47,11 @@
         <feature>esh-binding-lirc</feature>
     </feature>
 
+    <feature name="openhab-binding-mqtt" description="MQTT Binding" version="${project.version}">
+        <feature>esh-binding-mqtt</feature>
+        <feature>esh-binding-mqtt-generic</feature>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
@davidgraeff FYI. I assume that is all that needs to be done to make it available in openHAB?

features/addons-esh/src/main/feature/feature.xml
```diff
@@ -47,6 +47,11 @@
         <feature>esh-binding-lirc</feature>
     </feature>
 
+    <feature name="openhab-binding-mqtt" description="MQTT Binding" version="${project.version}">
```
cweitkamp: Do we need dependency on `<feature>openhab-transport-mqtt</feature>`?

davidgraeff: Definitely not. Esh mqtt transport is the new revamped one with asynchronous API and the openhab mqtt transport is the one from openHAB 1.x days.

davidgraeff:
There are 2 mqtt PRs queuing up on the Esh side that will make the experience a lot smoother. But all in all yes I'm happy that we finally have this and I approve.

PR #782, opened 2018-10-09
enable authentication by default

kaikreuzer:
Related to https://github.com/openhab/openhab-core/pull/412

Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #785, opened 2018-10-10
Revert "enable authentication by default"

kaikreuzer:
Reverts openhab/openhab-distro#782

### DIFF
--- a/distributions/openhab/src/main/resources/conf/services/runtime.cfg
+++ b/distributions/openhab/src/main/resources/conf/services/runtime.cfg
@@ -1,9 +1,3 @@
-################# AUTHENTICATION ################
-# Whether or not authentication should be active.
-# By default, authentication is enabled, so you need to login with a user that is specified
-# in OPENHAB_USERDATA/etc/users.properties.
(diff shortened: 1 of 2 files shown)


mstormi:
That revert didn't (fully?) revert, snapshot #1386 is still asking for auth ...

kaikreuzer:
@splatch What is missing?

mstormi:
FWIW, using #1383 now, that still does not work. Not asking for auth but it's redirecting the Basic UI URL to some .../login?.... which then is a 404.
So not sure how far back you need to go.

kaikreuzer:
@splatch Latest build is completely broken as it asks for auth, but does not accept the jaas user (openhab/habopen). What does one have to do to disable it? Or is the only chance to also revert the openhab-core PR, which added the bundles?

kaikreuzer:
@splatch I was talking about what is missing fir DEACTIVATING the feature for now. Does this require to fully remove the bundles again, i.e. also revert the openhab-core PR?

splatch:
I'm on it.

PR #786, opened 2018-10-11
breaking changes HTTP auth in UI and [Rr]uleEngine both in runtime.cfg

mstormi:
See https://github.com/eclipse/smarthome/pull/6310

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -12,9 +12,11 @@ ALERT;Yamaha Receiver Binding: The configuration parameter names now use lower c
 ALERT;Astro Binding: The 'kilometer' and 'miles' channels have been replaced by a new 'distance' channel
 ALERT;Jeelink Binding: The 'currentWatt' and 'maxWatt' channels have been replaced with 'currentPower' and 'maxPower' channels.
 ALERT;WeatherUnderground Binding: A bridge has been added on top of the current things, you need to add a bridge containing your api-key.
-ALERT;ZWave Binding: Major changes have been merged to support features such as security. All things must be deleted and re-added. Refer to https://community.openhab.org/t/zwave-binding-updates/51080 for further information.
+ALERT;ZWave Binding: Major changes have been merged to support features such as security. All things must be deleted and re-added.
+ALERT;Refer to https://community.openhab.org/t/zwave-binding-updates/51080 for further information.
 ALERT;Synop Binding is now using UoM. 'wind-speed-ms' and 'wind-speed-knots' channels have been replaced by a single 'wind-speed' channel.
 ALERT;RuleEngine renamed to ruleEngine in runtime.cfg.
+ALERT;HTTP(S) authentication on UI access enabled, disable using 'org.eclipse.smarthome.auth:authentication.enabled=false' in runtime.cfg on problems.
 
 
 [[PRE]]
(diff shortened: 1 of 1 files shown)


distributions/openhab/src/main/resources/bin/update.lst
```diff
@@ -10,10 +10,13 @@ ALERT;Yamaha Receiver Binding: The configuration parameter names now use lower c
...
+ALERT;Jeelink Binding: The 'currentWatt' and 'maxWatt' channels have been replaced with 'currentPower' and 'maxPower' channels.
 ALERT;WeatherUnderground Binding: A bridge has been added on top of the current things, you need to add a bridge containing your api-key.
-ALERT;ZWave Binding: Major changes have been merged to support features such as security. All things must be deleted and re-added. Refer to https://community.openhab.org/t/zwave-binding-updates/51080 for further information.
+ALERT;ZWave Binding: Major changes have been merged to support features such as security. All things must be deleted and re-added.
+ALERT;Refer to https://community.openhab.org/t/zwave-binding-updates/51080 for further information.
 ALERT;Synop Binding is now using UoM. 'wind-speed-ms' and 'wind-speed-knots' channels have been replaced by a single 'wind-speed' channel.
+ALERT;RuleEngine renamed to ruleEngine in runtime.cfg.
+ALERT;HTTP(S) authentication on UI access enabled, disable using 'org.eclipse.smarthome.auth:authentication.enabled=false' in runtime.cfg on problems.
```
kaikreuzer: Not anymore, see https://github.com/eclipse/smarthome/pull/6351

distributions/openhab/src/main/resources/bin/update.lst
```diff
@@ -10,10 +10,13 @@ ALERT;Yamaha Receiver Binding: The configuration parameter names now use lower c
 
 [2.4.0]
 ALERT;Astro Binding: The 'kilometer' and 'miles' channels have been replaced by a new 'distance' channel
-ALERT;Jeelink Binding: The 'currentWatt' and 'maxWatt' channels have been replaced with 'currentPower' and 'maxPower' channels
+ALERT;Jeelink Binding: The 'currentWatt' and 'maxWatt' channels have been replaced with 'currentPower' and 'maxPower' channels.
 ALERT;WeatherUnderground Binding: A bridge has been added on top of the current things, you need to add a bridge containing your api-key.
-ALERT;ZWave Binding: Major changes have been merged to support features such as security. All things must be deleted and re-added. Refer to https://community.openhab.org/t/zwave-binding-updates/51080 for further information.
+ALERT;ZWave Binding: Major changes have been merged to support features such as security. All things must be deleted and re-added.
+ALERT;Refer to https://community.openhab.org/t/zwave-binding-updates/51080 for further information.
```
kaikreuzer: This should not be an additional line.

distributions/openhab/src/main/resources/bin/update.lst
```diff
@@ -10,10 +10,13 @@ ALERT;Yamaha Receiver Binding: The configuration parameter names now use lower c
...
 ALERT;Synop Binding is now using UoM. 'wind-speed-ms' and 'wind-speed-knots' channels have been replaced by a single 'wind-speed' channel.
+ALERT;RuleEngine renamed to ruleEngine in runtime.cfg.
```
kaikreuzer: There is no entry "RuleEngine" in runtime.cfg, so this line is pretty misleading. This parameter is in fact undocumented anyhow, so I am not sure we should mention an undocumented so prominently here.

mstormi: Well it is in use by a number of people that's in fact how I came to open this PR (right after I got hit by that Karaf auth fiasco). I believe it's always better to warn upfront rather than have people run into it (even if it does not affect many users). Changed wording to clarify it's an optional parameter.

mstormi:
Ok then ... let's leave it out of the README so we don't confuse and animate the majority of people.
But word is spread so it's in use, and there's applications to it (particularly on slow machines like RPi,
e.g. when people start many cron jobs at the same time).
So still I strongly suggest to make it a supported, meant-to-be-used, documented parameter.
But I close the PR as that warning to be appear would have been the only remaining change.

PR #804, opened 2018-11-06
Demo items: Wrong semantic tag for basement

ghys:
The correct tag in the ESH ontology is "Basement", not "Cellar".


martinvw:
Thanks!

PR #809, opened 2018-11-15
Updates to powershell scripts (fixes, added logic, etc)

tmrobert8:
New PR with the following changes:
1. Implemented common.psm1 for common modules between all the powershell scripts
2. Added .bat scripts to ease running of powershell
3. Updated backup.ps1 to be compatible with the linux version
4. Created a restore.ps1 (wasn't implemented yet) - made compatible with linux version
5. Rewrote update.ps1 to fix a few bugs, add error handling and added 'stuff' from linux version (update.lst, etc)

Signed-off-by: Tim Roberts <troberts@bigfoot.com>

### DIFF
--- a/distributions/openhab/src/main/resources/bin/backup.bat
+++ b/distributions/openhab/src/main/resources/bin/backup.bat
@@ -1,0 +1,19 @@
+@ECHO off
+SETLOCAL
+
+powershell -command "& { . .\backup.ps1; Backup-openHAB -OHDirectory %~1 -OHBackups %~2 -FileName %~3 }"
+EXIT /B 0
+
+:printArgs
+ECHO Usage: backup.bat {OHDirectory} {OHBackups} {FileName}
+ECHO OHDirectory (optional) - The openHAB distribution directory
+ECHO OHBackups   (optional) - The directory where backups are stored
+ECHO FileName    (optional) - The backup file name (found in OHBackups)
+ECHO.
+ECHO Example to backup openHAB to the default locations
+ECHO    backup.bat
+ECHO.
+ECHO Example to backup openHAB to "backup.zip" in "c:\openhab2\backups":
+ECHO    update.bat "c:\openhab2" "c:\openhab2\backups" "backup.zip" 
+ECHO.
+EXIT /B -1
--- a/distributions/openhab/src/main/resources/bin/backup.bat
+++ b/distributions/openhab/src/main/resources/bin/backup.bat
@@ -1,11 +1,30 @@
 @ECHO off
 SETLOCAL
 
-powershell -command "& { . .\backup.ps1; Backup-openHAB -OHDirectory %~1 -OHBackups %~2 -FileName %~3 }"
+IF "%1"=="?" GOTO printArgs
+IF "%1"=="\?" GOTO printArgs
+IF "%1"=="/?" GOTO printArgs
+
+SET bargs=
+
+IF NOT [%1]==[] ( SET bargs=%bargs% -MaxFiles %~1% )
+IF NOT [%2]==[] ( SET bargs=%bargs% -OHDirectory %~2% )
+IF NOT [%3]==[] ( SET bargs=%bargs% -OHBackups %~3% )
+IF NOT [%4]==[] ( SET bargs=%bargs% -FileName %~4% )
+
+CD %~dp0
+powershell -ExecutionPolicy Bypass -command "&{. .\backup.ps1; Backup-openHAB %bargs% }" 
+SET LEVEL=%ERRORLEVEL%
+
+if %LEVEL% LSS 0 (
+    PAUSE
+    EXIT /B %LEVEL%
+)
 EXIT /B 0
 
 :printArgs
-ECHO Usage: backup.bat {OHDirectory} {OHBackups} {FileName}
+ECHO Usage: backup.bat {MaxFiles} {OHDirectory} {OHBackups} {FileName}
+ECHO MaxFiles    (optional) - The maximum number of backups to keep (specify 0 for no maximum)
 ECHO OHDirectory (optional) - The openHAB distribution directory
 ECHO OHBackups   (optional) - The directory where backups are stored
 ECHO FileName    (optional) - The backup file name (found in OHBackups)
@@ -13,7 +32,7 @@ ECHO.
 ECHO Example to backup openHAB to the default locations
 ECHO    backup.bat
 ECHO.
-ECHO Example to backup openHAB to "backup.zip" in "c:\openhab2\backups":
-ECHO    update.bat "c:\openhab2" "c:\openhab2\backups" "backup.zip" 
+ECHO Example to backup openHAB to "backup.zip" in "c:\openhab2\backups" keeping at most 10 backups:
+ECHO    backup.bat 10 "c:\openhab2" "c:\openhab2\backups" "backup.zip" 
 ECHO.
 EXIT /B -1
--- a/distributions/openhab/src/main/resources/bin/backup.ps1
+++ b/distributions/openhab/src/main/resources/bin/backup.ps1
@@ -44,7 +44,8 @@ Function Backup-openHAB {
         BoxMessage "openHAB 2.x.x backup script" Magenta
         Write-Host ""
 
-        CheckForAdmin
+        # Check for admin (commented out - don't think we need it)
+        # CheckForAdmin
 
         Write-Host -ForegroundColor Cyan "Checking the specified openHAB directory"
         $OHDirectory = GetOpenHABRoot $OHDirectory
--- a/distributions/openhab/src/main/resources/bin/backup.ps1
+++ b/distributions/openhab/src/main/resources/bin/backup.ps1
@@ -1,4 +1,7 @@
-﻿Function Backup-openHAB {
+﻿#Requires -Version 5.0
+Set-StrictMode -Version Latest
+
+Function Backup-openHAB {
     <#
     .SYNOPSIS
     Backsup openHAB files.
@@ -8,60 +11,57 @@
     The directory where openHAB is installed (default: current directory).
     .PARAMETER OHBackups
     The directory to backup the files to.
-    .PARAMETER ZipFileOut
+    .PARAMETER FileName
     The name of the zip file to create
     .EXAMPLE
     Backup an openHAB instance to a zip file
     Backup-openHAB
     .EXAMPLE
     Backup the openHAB distribution in the C:\openHAB2 directory to c:\openHAB2-backup\backup.zip
-    Backup-openHAB -OHDirectory C:\openHAB2 -OHBackups c:\openHAB2-backup -ZipFileOut backup.zip
+    Backup-openHAB -OHDirectory C:\openHAB2 -OHBackups c:\openHAB2-backup -FileName backup.zip
     #>
 
     [CmdletBinding()]
     param(
-        [Parameter(ValueFromPipeline=$True)]
+        [Parameter(ValueFromPipeline = $True)]
         [string]$OHDirectory = ".",
-        [Parameter(ValueFromPipeline=$True)]
+        [Parameter(ValueFromPipeline = $True)]
         [string]$OHBackups,
-        [Parameter(ValueFromPipeline=$True)]
-        [string]$ZipFileOut
+        [Parameter(ValueFromPipeline = $True)]
+        [string]$FileName
     )
 
     begin {}
     process {
 
-        if (!([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()
-            ).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)) {
-            throw "This script must be run as an Administrator. Start PowerShell with the Run as Administrator option"
+        Import-Module $PSScriptRoot\common.psm1 -Force
+
+        Write-Host ""
+        BoxMessage "openHAB 2.x.x backup script" Magenta
+        Write-Host ""
+
+        CheckForAdmin
+
+        Write-Host -ForegroundColor Cyan "Checking the specified openHAB directory"
+        $OHDirectory = GetOpenHABRoot $OHDirectory
+        if ($OHDirectory -eq "") {
+            return PrintAndReturn "Could not find the userdata directory! Make sure you are in the openHAB directory or specify the -OHDirectory parameter!"
+        }
+    
+        $OHConf = "$OHDirectory\conf"
+        $OHUserdata = "$OHDirectory\userdata"
+
+        if ([string]::IsNullOrEmpty($OHBackups)) {
+            $OHBackups = "$OHDirectory\backups"
         }
 
-        Write-Host -ForegroundColor Cyan "Checking the specified openHAB directory..."
-        if (!(Test-Path "$OHDirectory\userdata") -And !(Test-Path -Path "$OHDirectory\conf")) {
-            throw "$OHDirectory\userdata doesn't exist! Make sure you are in the " +
-                "openHAB directory or specify the -OHDirectory parameter!"
-        }
-
-        if ($OHDirectory -eq '.') {$OHDirectory = pwd }
-
-        if ([Environment]::GetEnvironmentVariable("OPENHAB_CONF", "Machine")) {
-            $OHConf = [Environment]::GetEnvironmentVariable("OPENHAB_CONF", "Machine")
-        } else {
-            $OHConf = "$OHDirectory\conf"
-        }
-        if ([Environment]::GetEnvironmentVariable("OPENHAB_USERDATA", "Machine")) {
-            $OHUserdata = [Environment]::GetEnvironmentVariable("OPENHAB_USERDATA", "Machine")
-        } else {
-            $OHUserdata = "$OHDirectory\userdata"
-        }
-        if (!($OHBackups)) {
-            if ([Environment]::GetEnvironmentVariable("OPENHAB_BACKUPS", "Machine")) {
-                $OHBackups = [Environment]::GetEnvironmentVariable("OPENHAB_BACKUPS", "Machine")
-            } else {
-                $OHBackups = "$OHDirectory\backups"
-                if (!(Test-Path "$OHDirectory\backups")){
-                    mkdir "$OHBackups" | Out-Null
-                }
+        if (!(Test-Path $OHBackups -PathType Container)) {
+            try {
+                Write-Host -ForegroundColor Cyan "Creating backup directory $OHBackups"
+                CreateDirectory $OHBackups
+            }
+            catch {
+                return PrintAndReturn "Error creating backup directory $OHBackups - exiting" $_
             }
         }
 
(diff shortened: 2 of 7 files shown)


tmrobert8:
@bdleedy 
Would you mind testing the scripts (backup.ps1, restore.ps1 and update.ps1) and post any issues you have.  Please note you'll need to copy all three of these PLUS the common.psm1 to your runtime/bin directory.

"backup-openhab" will simply backup everything to the 'backups' directory
"restore-openhab" will restore it then (if you give it no parameters, restores the latest one)
"update-openhab" does all the updates.  Note: specify the secret "-KeepUpdateScript $true" for it to persist the update.ps1/common.psm1 files over the update (allows you to test without losing those files).

I'm particularly interested in the error handling of them.  If the update-openhab fails after it get's to the step where it starts modifying thing - it should automatically restore all the files to the prior version (keeping you a working copy of openhab prior to trying the update)

bdleedy:
Looking at this now.

bdleedy:
How do you intend the .bat files to be run? Double click? Right-click Run as Admin?

distributions/openhab/src/main/resources/bin/backup.bat
```diff
@@ -0,0 +1,19 @@
...
+ECHO Usage: backup.bat {OHDirectory} {OHBackups} {FileName}
+ECHO OHDirectory (optional) - The openHAB distribution directory
+ECHO OHBackups   (optional) - The directory where backups are stored
+ECHO FileName    (optional) - The backup file name (found in OHBackups)
+ECHO.
+ECHO Example to backup openHAB to the default locations
+ECHO    backup.bat
+ECHO.
+ECHO Example to backup openHAB to "backup.zip" in "c:\openhab2\backups":
+ECHO    update.bat "c:\openhab2" "c:\openhab2\backups" "backup.zip"
```
bdleedy: Should read "backup.bat"

bdleedy:
I'm not completely sure that calling dot sourced scripts and then running functions is going to work. I cannot find anyway for this to work for me...

distributions/openhab/src/main/resources/bin/backup.bat
```diff
@@ -0,0 +1,19 @@
+@ECHO off
+SETLOCAL
+
```
bdleedy: If you're planning on users running this from the BIN folder then maybe something like this. Especially if you want them to be able to Right-click - Run as admin.

cd %~dp0
cd ..\..

distributions/openhab/src/main/resources/bin/backup.bat
```diff
@@ -0,0 +1,19 @@
...
+powershell -command "& { . .\backup.ps1; Backup-openHAB -OHDirectory %~1 -OHBackups %~2 -FileName %~3 }"
```
bdleedy: To allow for RC - RAA. You can pretty up the arguments with variables if you want. Checks if arguments are empty. If you don't do this, PS throws an error about OHDirectory.

if [%1]==[] ( powershell -command "&{. .\runtime\bin\backup.ps1; Backup-openHAB }" ) else ( powershell -command "&{. .\runtime\bin\backup.ps1; Backup-openHAB -OHDirectory %~1 -OHBackups %~2 -FileName %~3 }" )

distributions/openhab/src/main/resources/bin/backup.ps1
```diff
@@ -8,103 +11,152 @@
...
-            $OHConf = [Environment]::GetEnvironmentVariable("OPENHAB_CONF", "Machine")
-        } else {
-            $OHConf = "$OHDirectory\conf"
+        Write-Host -ForegroundColor Cyan "Checking the specified openHAB directory"
+        $OHDirectory = GetOpenHABRoot $OHDirectory
+        if ($OHDirectory -eq "") {
+            return PrintAndReturn "Could not find the userdata directory! Make sure you are in the openHAB directory or specify the -OHDirectory parameter!"
         }
-        if ([Environment]::GetEnvironmentVariable("OPENHAB_USERDATA", "Machine")) {
-            $OHUserdata = [Environment]::GetEnvironmentVariable("OPENHAB_USERDATA", "Machine")
```
bdleedy: I only put these checks in here because there was talk of checking for set environment variables. Is that no longer a thing?

distributions/openhab/src/main/resources/bin/restore.bat
```diff
@@ -0,0 +1,24 @@
...
+ECHO OHDirectory (optional) - The openHAB distribution directory
+ECHO OHBackups   (optional) - The directory where backups are stored
+ECHO FileName    (optional) - The backup file name (found in OHBackups)
+ECHO AutoConfrim (optional) - "true" to automatically confirm restoration, "false" otherwise
+ECHO.
+ECHO Example to restore openHAB from the latest backup in the default locations
+ECHO    restore.bat
+ECHO.
+ECHO Example to restore openHAB from the "backup.zip" found in "c:\openhab2\backups" with auto confirming:
+ECHO    update.bat "c:\openhab2" "c:\openhab2\backups" "backup.zip" true
```
bdleedy: "restore.bat"

distributions/openhab/src/main/resources/bin/restore.bat
```diff
@@ -0,0 +1,24 @@
+@ECHO off
+SETLOCAL
+
```
bdleedy: Same deal here so that it can be RC - RAA.

cd %~dp0
cd ..\..

distributions/openhab/src/main/resources/bin/restore.bat
```diff
@@ -0,0 +1,24 @@
...
+IF NOT [%~4]==[] IF NOT "%~4"=="true" IF NOT "%~4"=="false" GOTO printArgs
+
+SET autoConfirm=False
+IF "%~4"=="true" SET autoConfirm=True
+powershell -command "& { . .\restore.ps1; Restore-openHAB -OHDirectory %~1 -OHBackups %~2 -FileName %~3 -AutoConfirm $%autoConfirm% }"
```
bdleedy: and again adding....
. .**_\runtime\bin_**\restore.ps1
...so that RC - RAA works.

Also need to check for %1 again.
`if [%1]==[] (powershell -command "& { . .\runtime\bin\restore.ps1; Restore-openHAB }") else (powershell -command "& { . .\runtime\bin\restore.ps1; Restore-openHAB -OHDirectory %~1 -OHBackups %~2 -FileName %~3 -AutoConfirm $%autoConfirm% }")`

distributions/openhab/src/main/resources/bin/restore.ps1
```diff
@@ -0,0 +1,211 @@
...
+        }
+    
+        $OHConf = "$OHDirectory\conf"
+        $OHUserdata = "$OHDirectory\userdata"
+
+        if (([string]::IsNullOrEmpty($OHBackups))) {
+            $OHBackups = "$OHDirectory\backups"
+        }
+
+        if ([string]::IsNullOrEmpty($FileName)) {
```
bdleedy: It would be nice to have something here like:

`Write-Host -ForegroundColor Yellow "No backup file specified. Finding latest backup."`

distributions/openhab/src/main/resources/bin/update.bat
```diff
@@ -0,0 +1,24 @@
+@ECHO off
+SETLOCAL
+
```
bdleedy: You know the drill at this point...

distributions/openhab/src/main/resources/bin/update.bat
```diff
@@ -0,0 +1,24 @@
...
+IF [%~1]==[] GOTO printArgs
```
bdleedy: I don't like this here. The whole point of the PS script was to handle autoupdates to the most current...

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -69,190 +12,582 @@ Function Update-openHAB {
...
+        $request = [System.Net.HttpWebRequest]::Create($uri)
+        $request.set_Timeout(15000)
 
-        if (!([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()
-            ).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)) {
-            throw "This script must be run as an Administrator. Start PowerShell with the Run as Administrator option"
+        Invoke-WebRequest $uri -Outfile $Outputfile -ErrorAction Stop
+    }
+
+    function NormalizeVersionNumber() {
```
bdleedy: There's this cool new function but it's not actually used before we try to download files that don't exist... Like `upgrade.bat 2.3`

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -69,190 +12,582 @@ Function Update-openHAB {
...
+    function DownloadFiles() {
+        param(    
+            [Parameter(Mandatory = $True)]
+            [string] $DownloadSource, 
+            [Parameter(Mandatory = $True)]
+            [string] $OutputFile
+        ) 
+    
+        $uri = New-Object "System.Uri" "$DownloadSource"
+
```
bdleedy: I really hate that the fancy percentage code that (yes, I found) is gone...
It was nice to know how long this was going to take.

I've found that I have needed to add the following code to get it to work.
`[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12`

distributions/openhab/src/main/resources/bin/update.ps1
bdleedy: Also, going from 2.3.0 to 2.3.0 would be a great way to do a reinstall but we should do some validation/notification that there isn't really an 'upgrade' happening.

No error handling when file just doesn't exist. Like Stable 2.4.0.

bdleedy:
- I ran update-openhab from the root and it deleted all of my files/folders in that folder.
- Probably should put `-ExecutionPolicy Bypass` in the batch files also

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -69,190 +12,582 @@ Function Update-openHAB {
...
-            Write-Host ""
-            Write-Host -ForegroundColor Magenta "Backup script starting..."
-            Write-Host -ForegroundColor Cyan "Making a backup in '$OHDirectory\backups' ..."
-            $BackupScript = Join-Path (pwd) '\runtime\bin\backup.ps1'
-            . $BackupScript
-            Backup-openHAB -OHDirectory (pwd)
-            Write-Host -ForegroundColor Magenta "Backup script finished."
-            Write-Host ""
+    if ($Snapshot) {
+        Write-Host -ForegroundColor Yellow "The current version is $CurrentVersion-SNAPSHOT"
```
bdleedy: This doesn't make any sense. Just because I choose to upgrade to a snapshot, doesn't mean I have a snapshot.

distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -69,190 +12,582 @@ Function Update-openHAB {
...
+                . $newUpdate
+                Update-openHAB -OHDirectory $OHDirectory -OHVersion $rawOHVersion -Snapshot $Snapshot -AutoConfirm $AutoConfirm -SkipNew $true -KeepUpdateScript $KeepUpdateScript
+                return 2;
+            } catch {
+                return PrintAndReturn "Execution of new update.ps1 failed - please execute it yourself (found in $newUpdate)" $_
+            }
         }
+    }
 
+    # Backup openHAB only if not coming via new update script
```
bdleedy: I thought I was going crazy... This is all well and good....
...but you've forgotten to download the files! :-D

tmrobert8:
@bdleedy 
I'll start looking at these comments today but wanted to start a conversation on two points:
1. Run as admin - I know you copied this from the linux side but do you see any reason for us to ensure you run as admin?  99.9% of the people will likely be installing this using their own ID (which probably is already admin) rather than install under an admin ID then switch to a non-admin ID to run the backup/restore/update.  I'm tempted just to remove those checks in the code (or replace them with a file write check to see if permissions are required and throw an error only if required).  What do you think?

2. Environment variables - I know what your saying here but nothing has occurred towards this and I doubt it would be useful even if someone went with that (very unlikely that someone will separate up the directory structure like the linux version).  I could put those checks back in (really not that big of a deal) but what do you think?

bdleedy:
1. I'm not sure actually. That's a very good question. You are probably correct though, in that most users are logging on with an account that has admin rights. Although it's not just about admin rights, UAC could possibly come into play. That's a tough one. One of the discussions we've had in the past was whether to use batch files or not. Microsoft has said that it's going to die. In the latest releases you can't Shift + Right Click and open a CMD prompt anymore, only PowerShell (or Linux). We talked about the types of users we have and what knowledge is required of them to run oH. The idea was that anyone should be able to do it, even if it's help from the documentation or community. Limiting that need is the goal. Is it easier to make sure users RAA or to help people with permissions on files/folders?

2. I didn't think this was a thing that would ever catch on in the current state. Maybe if there was an installer, but not now.

3. Nope. I just made sure that the important things really stuck out from anything around it.

distributions/openhab/src/main/resources/bin/backup.bat
tmrobert8: I didn't bother with that in the batch files since the powerscript files do that (that way you get the same functionality regardless if you call it from the batch or powerscript files directly).

tmrobert8: ah - I see what you mean by that.  Will fix

distributions/openhab/src/main/resources/bin/update.bat
tmrobert8: Oversight - removing

distributions/openhab/src/main/resources/bin/update.ps1
tmrobert8: I like it (ie reinstalling).  But what do you think we should do with AutoConfirm=true if we are reinstalling?  

1. Do we assume the users intention was to reinstall and just give a warning message about reinstallation (rather than a confirmation since it's autoconfirm)?
2. Or do we consider this situation special enough that we ignore the autoconfirm on a reinstallation and ask the user?

distributions/openhab/src/main/resources/bin/update.ps1
tmrobert8: That 'fancy percentage code' was always messing up for me (seemed like it was doing new lines on every update or so - made the cursor jump around like crazy as well).  You should still see the progress (ie bytes downloaded) from the native invoke-webrequest progress meter.

Surprised on the tls12 flag - worked fine for me but I've added that anyway

tmrobert8:
"I ran update-openhab from the root and it deleted all of my files/folders in that folder."

Could you describe what you mean here? You ran it from the root (c:\openhab2 I'm assuming) and did something fail and it didn't restore the files/folders?  or did it work correctly but deleted extra files/folders?  or am I missing your point entirely!  :)

distributions/openhab/src/main/resources/bin/update.ps1
tmrobert8: Really not sure what you mean here about "you've forgotten to download the files"?  

If I *think* what you mean (which probably a bad assumption!), the files were downloaded, we found a new update script, executed the update script and use the files that were already downloaded.  So what am I missing?

distributions/openhab/src/main/resources/bin/update.ps1
bdleedy: Ugh... Well, if we just go ahead and do it, what's the worst thing that happens? They stay where they were. I say just go for it.

distributions/openhab/src/main/resources/bin/update.ps1
bdleedy: Nope. If you decide to `$SKIPNEW -eq $True`, you have never downloaded the new files. The only place they are downloaded is within the `if ($SkipNew -eq $False)` code. You handle the Addons and LegacyAddons but not the actual distro.

bdleedy:
I BELIEVE, I was using the PowerShell `. .\Update.ps1` from the bin folder, then I went to the root and ran `Update-openHAB....` Yes...why would you do that? I have no idea. Haha.

distributions/openhab/src/main/resources/bin/update.ps1
tmrobert8: Yeah - that's actually an issue I tracked down yesterday (don't have the issue link - but it's basically if you run in the console, the download progress meter slows it down drastically.  If you run in a powershell console, it flies.  I'll look into putting the old meter back in and trying to fix it for my stuff.

BTW - almost ready to commit a new update that is fairly big changes (went ahead and did the environment variables in all scripts - mainly because I'm also doing an installer and was thinking about allowing that via the installer as well)..

bdleedy: Nice!

Let me know what you need to for testing the downloading.

tmrobert8:
@bdleedy 
Just updated - pretty big changes:
1. Alot of the stuff you asked for
2. Progress meter is back (and fixed to work atleast on my system)
3. Lots of fixes that I noticed (and added some stuff I missed from the linux scripts [like backups could be under userdata])
4. Usage of environment variables for directory location - this was a pretty big change actually as I had made a bunch of assumptions on the structure.  If you want to fully test this - do the following:
```
SET OPENHAB_CONF=c:\openhabtest\bconf\conf
SET OPENHAB_USERDATA=c:\openhabtest\buserdata\userdata
SET OPENHAB_RUNTIME=c:\openhabtest\bruntime\runtime
SET OPENHAB_HOME=c:\openhabtest\bhome
SET OPENHAB_ADDONS=C:\openhabtest\baddons\addons
SET OPENHAB_BACKUPS=C:\openhabtest\bbackups\backups
```
Of course setting it to a directory structure of your own - but you'll see backup/restore/update will work with only those directories.  Note: if you do the above, the system won't actually start as there are issues in the other scripts (namely start.bat assumes runtime will be under the current directory).  But atleast you can see the three scripts will work correctly with the environment variable specifications

tmrobert8:
@bdleedy 
What do you think about removing the $Snapshot parameter?  update.sh just looks at the version that is passed to it to see if "-SNAPSHOT" is applied to it (update.sh 2.3.0-SNAPSHOT).  This becomes important because the latest version of update.sh (which I just looked at) handles milestone builds as well "2.3.0-m6" and switches to the milestone URLS based on that.

If we keep $Snapshot - then I'd need to add a $Milestone and we are starting to get into parameter explosion for each new type.  

An alternative would be to continue what we are doing - but replace $Snapshot with $Type that can be "snapshot" or "milestone" or whatever.

So our options are:
1. Switch to just detecting from the passed in version ("2.3.0", "2.3.0-SNAPSHOT", "2.3.0-M6")
2. Add a $Milestone along side the $Snapshot (don't like this option)
3. Replace $Snapshot with $Type (or something better named) that contains "snapshot", "milestone" or whatever future has

Tim

tmrobert8:
FYI - went with the first option (which matches how update.sh works).  So there is no longer a $Snapshot parameter - simply specify a version of "2.4.0-SNAPSHOT"

bdleedy:
OK, I'll test this out today hopefully. I see that Kai has published an expected release date so we've got a deadline.

kaikreuzer:
@bdleedy Any feedback from your testing? Should this be merged?

bdleedy:
I can't find any problems. Only one question.

If openHAB is running, the script throws an exception. Is that new? I thought we just killed it before... Maybe the check for a service needs to go before this then since we check for the service and then stop it (killing openHAB) and delete the service.

tmrobert8:
@schrej-zz 
The scripts do that exactly that (if OPENHAB_CONF is not defined, use OPENHAB_HOME/conf instead).  I think I confused you by my message - what that message above meant was that even if you (manually) defined each of the variables to separate directories (such as OPENHAB_CONF=%OPENHAB_HOME%\bconf\conf) - the scripts would still work as intended.

tmrobert8:
@bdleedy 
Yes - I changed that behavior to match how the linux script operates.  I was trying to keep the windows functionality close to the linux one in features and behavior - making it easier to maintain when changes need to happen.  Let me know if you think that was unwise and I can change it back (and then maybe raise an issue to have the linux one match the windows one!).

BTW - I will be shortly committed the following (final) changes to the scripts:
1. Remove the requirement for ADMIN.  I don't see any need for it and it complicated the windows install scripts as well (because those need to run as admin then since they can schedule a backup/upgrade script.
2. Restore but deprecate the SNAPSHOT parameter.  I realized this weekend that the existing script will detect the new update.ps1 and try to call it with a SNAPSHOT parameter - which would then fail since it's no longer defined.  I'll put it back in (and handle it correctly) but issue a big boxed message saying that it's deprecated.

Look for it later today..
Tim

tmrobert8:
@bdleedy @kaikreuzer 
Should be final version - just removed need for admin, added -snapshot back as deprecated and fixed issue with case sensitivity when looking for -snapshot

tmrobert8:
@kaikreuzer 
Will this be merged for 2.4 stable?  Trying to decide how to handle this in the windows installer

kaikreuzer:
> Will this be merged for 2.4 stable?

I don't know. As @bdleedy tested and commented it, I'd leave this decision to him - if he approves the PR, it can be merged.

kaikreuzer:
Perfect then, thank you both!

PR #833, opened 2018-12-28
Ensure that UTF-8 is used as default for all text files on Windows

MHerbst:
This makes sure that the behaviour on Windows is the same as on other platforms where UTF-8 is already the default encoding.

See https://github.com/eclipse/smarthome/issues/6630

Signed-off-by: Martin Herbst <develop@mherbst.de>


PatrikG8:
Please add te solution also to the sample of openhab-wrapper.conf on https://www.openhab.org/docs/installation/windows.html#starting-openhab-as-a-service so its also works as Windows-Service.

Issue #423, opened 2017-02-16
Wrong redirection when behind a reverse proxy server


robnielsen:
When smarthome is behind a reverse proxy server that only allows https, the redirection from https://host/ is improperly going to http://host/start/index. This is happening because the 302 redirect is ignoring the X-Forwarded-Proto header. See https://community.openhab.org/t/nginx-redirect-to-port-80/16760/6 for more details.

According to @kaikreuzer, the redirection is occurring at https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/runtime/etc/jetty.xml#L42-L50

kaikreuzer:
Did you manage to figure out what needs to be changed in the Jetty configuration to solve this problem?

jeffsf:
As noted, https://www.example.com/exposed/path/ sends the browser to http://www.example.com/start/index

From what I can tell, the source from openhab-master `launch/home/etc/jetty.xml` causes a redirect, rather than a rewrite

    <New class="org.eclipse.jetty.rewrite.handler.RedirectRegexRule">
      <Set name="regex">/$</Set>
      <Set name="replacement">/start/index</Set>
    </New>

Unfortunately, it also appears that this is "compiled in" and is not present in `/var/lib/openhab2/etc/jetty.xml`

kaikreuzer:
> Unfortunately, it also appears that this is "compiled in" and is not present in /var/lib/openhab2/etc/jetty.xml

No, this is exactly the file you can find in  https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/runtime/etc/jetty.xml.

If you have a solution, please provide a PR for it.

jeffsf:
I've banged at this for a while, and haven't found a quick solution. If someone else takes a stab at it, here's what I've found so far:

My nginx reverse proxy is properly inserting/modifying the headers

- Host
- X-Real-IP
- X-Forwarded-For
- X-Forwarded-Proto

In the default configuration, a `GET / HTTP/1.1` request results in openHAB returning
```
HTTP/1.1 302 Found
Date: Wed, 15 Mar 2017 01:13:47 GMT
Location: http://www.example.com/start/index
Connection: close
Server: Jetty(9.2.19.v20160908)
```
which, as noted is _not_ correct.

If the proxy is configured to redirect http to https, it _appears_ to work to the end user, but only because the redirect to http handled at the proxy.

[...]

PR #840, opened 2019-01-09
Use X-Forwarded-Proto for forwarded protocol as RFC7239 states

stefan-muc:
Fixes https://github.com/openhab/openhab-distro/issues/423

In distributions/openhab/src/main/resources/runtime/etc/jetty.xml there's quite parallel configuration, just line numbers are different. Should this fix also be applied there?


stefan-muc:
Obsolete through #846 because of contributing guidelines (sign-off line in commits)

PR #846, opened 2019-01-14
Use X-Forwarded-Proto for forwarded protocol as RFC7239 states

### LINKED ISSUE #423: Wrong redirection when behind a reverse proxy server

stefan-muc:
Fixes #423

In distributions/openhab/src/main/resources/runtime/etc/jetty.xml there's quite parallel configuration, just line numbers are different. Should this fix also be applied there?

(Sorry didn't know about contribution guidelines with sign-off line, so this pull request replaces #840)

### DIFF
--- a/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
+++ b/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
@@ -23,7 +23,7 @@
 					<Call name="addRule">
 						<Arg>
 							<New class="org.eclipse.jetty.rewrite.handler.ForwardedSchemeHeaderRule">
-								<Set name="header">X-Forwarded-Scheme</Set>
+								<Set name="header">X-Forwarded-Proto</Set> <!-- RFC7239 states -Proto not -Scheme as header, Jetty seems to have wrong defaults -->
 								<Set name="headerValue">https</Set> <!-- if this is unset, any value will match against the rule -->
 								<Set name="scheme">https</Set>
 							</New>
@@ -32,7 +32,7 @@
 					<Call name="addRule">
 						<Arg>
 							<New class="org.eclipse.jetty.rewrite.handler.ForwardedSchemeHeaderRule">
-								<Set name="header">X-Forwarded-Scheme</Set>
+								<Set name="header">X-Forwarded-Proto</Set> <!-- RFC7239 states -Proto not -Scheme as header, Jetty seems to have wrong defaults -->
 								<Set name="headerValue">http</Set> <!-- if this is unset, any value will match against the rule -->
 								<Set name="scheme">http</Set>
 							</New>
--- a/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
+++ b/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
@@ -23,7 +23,7 @@
 					<Call name="addRule">
 						<Arg>
 							<New class="org.eclipse.jetty.rewrite.handler.ForwardedSchemeHeaderRule">
-								<Set name="header">X-Forwarded-Proto</Set> <!-- RFC7239 states -Proto not -Scheme as header, Jetty seems to have wrong defaults -->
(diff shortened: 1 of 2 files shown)


BClark09:
Hi @stefan-muc, the file you've named in the description is the file you've changed. Did you mean `/launch/home/etc/jetty.xml`? I believe this file is used for the development IDE so should be changed along with the current file.

distributions/openhab/src/main/resources/runtime/etc/jetty.xml
```diff
@@ -23,7 +23,7 @@
 					<Call name="addRule">
```
BClark09: Could you change the comment above in-line with the new change?

distributions/openhab/src/main/resources/runtime/etc/jetty.xml
```diff
@@ -23,7 +23,7 @@
...
 						<Arg>
 							<New class="org.eclipse.jetty.rewrite.handler.ForwardedSchemeHeaderRule">
-								<Set name="header">X-Forwarded-Scheme</Set>
+								<Set name="header">X-Forwarded-Proto</Set> <!-- RFC7239 states -Proto not -Scheme as header, Jetty seems to have wrong defaults -->
```
BClark09: I don't think these are Jetty defaults as such, probably no need for the comment here.

distributions/openhab/src/main/resources/runtime/etc/jetty.xml
```diff
@@ -32,7 +32,7 @@
...
```
BClark09: Same as above

distributions/openhab/src/main/resources/runtime/etc/jetty.xml
stefan-muc: They are in deed Jetty defaults, see [Jetty API documentation](http://archive.eclipse.org/jetty/8.1.0.RC0/apidocs/org/eclipse/jetty/rewrite/handler/ForwardedSchemeHeaderRule.html) (and I can't find that class name anywhere else in OpenHAB sources). So I would leave the comment, okay?

stefan-muc:
> Hi @stefan-muc, the file you've named in the description is the file you've changed. Did you mean `/launch/home/etc/jetty.xml`? I believe this file is used for the development IDE so should be changed along with the current file.

Yes sorry, I confused the paths - I wanted to refer the path you are mentioning.

Updated my branch to fix this file too.

Regarding my source comment see above: `org.eclipse.jetty.rewrite.handler.ForwardedSchemeHeaderRule`  is indeed a Jetty class, the call it "Scheme" not "Proto"

distributions/openhab/src/main/resources/runtime/etc/jetty.xml
wborn: Did you read RFC7239? It actually specifies a [Forwarded](https://tools.ietf.org/html/rfc7239#section-4) header to  replace and standardize headers like `X-Forwarded-Scheme` and `X-Forwarded-Proto`. Jetty doesn't define a default and allows you to use the `RewriteHandler` to use whatever header you want for this. 

So the comment doesn't make sense and for better compatibility and not breaking peoples existing setups you may want to define additional rules for this. People may currently have setup their NGINX/Apache to use this `X-Forwarded-Scheme` header instead of a `X-Forwarded-Proto` header.

stefan-muc:
You're welcome! Thanks for the merge, I'm glad this got accepted.

PR #854, opened 2019-01-23
added some add-ons that should be part of every package

kaikreuzer:
Re-adds the bundles that were removed with https://github.com/openhab/openhab-core/commit/45a7ce27cbbb359bbd7aba6be0a7101bb2c920f3#diff-1a67ec2e35cbc850af315533c3fec28cL45

Depends on https://github.com/openhab/openhab2-addons/pull/4657

### DIFF
--- a/distributions/openhab/pom.xml
+++ b/distributions/openhab/pom.xml
@@ -132,6 +132,9 @@
                         <feature>service</feature>
                         <feature>system</feature>
                         <feature>openhab-runtime-base</feature>
+                        <feature>openhab-io-javasound</feature>
+                        <feature>openhab-io-webaudio</feature>
+                        <feature>openhab-iconset-classic</feature>
                     </bootFeatures>
                     <archiveZip>false</archiveZip>
                     <archiveTarGz>false</archiveTarGz>


kaikreuzer:
Crap, for some reason, this isn't working.
@maggu2810 Do you have any good advice how we can add certain features that come from openhab2-addons to the distro (in all cases)? So far, only the openhab-runtime-base feature was installed by default, but it included bundles that were now moved to openhab2-addons...

kaikreuzer:
I still didn't find a solution for this 😢. Any help appreciated!

maggu2810:
> @maggu2810 Do you have any good advice how we can add certain features that come from openhab2-addons to the distro (in all cases)? So far, only the openhab-runtime-base feature was installed by default, but it included bundles that were now moved to openhab2-addons...

Sorry!
@kaikreuzer I did not realize such notifications of specific repos as there are two much noise and so I didn't read all that mails (of specific repos) in time.

kaikreuzer:
No worries, I know how easy it is to miss a single ping in the flood of Github notifications...

PR #858, opened 2019-01-30
removed openHAB Core bundles from openHAB IDE setup

kaikreuzer:
With https://github.com/openhab/openhab-distro/pull/856, we removed the ESH checkout already, but the openhab-core project was still left in the IDE setup.

Since https://github.com/openhab/openhab-core/pull/467, the openhab-core projects are incompatible with the PDE based IDE and require a separate IDE setup, so they need to be removed here.

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2576,40 +2576,6 @@
         name="master"
         label=""/>
   </project>
-  <project name="9_core"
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Note: The PR build failure is not due to this change - this PR only impacts (fixes!) the IDE setup.

PR #888, opened 2019-03-04
Update OH2 addons features dependency groupId/artifactId

wborn:
Should be merged when openhab/openhab2-addons#5018 is merged.


maggu2810:
:+1: I had such a commit on my local machine, too.
Thanks for your one!

PR #889, opened 2019-03-08
add a bnd based demo app with core, addons and UIs

maggu2810:


maggu2810:
While thinking about it, a separate repository for the demo application could make sense.
In the long run it should avoid conflicts if we edit the POM and we could things handle as simple as possible.

maggu2810:
@kaikreuzer: Would it be possible to create a repo "openhab-demo" and give me access rights?

maggu2810:
This repo could be transferred to openHAB: https://github.com/maggu2810/openhab-demo

maggu2810:
With this demo that is not bound to any other stuff (the repo here is clean and does not contain any code), we could IMHO take the freedom and remove the demo app that is part of the core.

kaikreuzer:
Just to understand what this "demo app" is about:

So far, all we had is a simply [launch file](https://github.com/openhab/openhab-distro/blob/master/launch/openHAB_Runtime.launch), which used bundles from the TP and launches them on Equinox. It had parameters like `-Dsmarthome.configdir=${project_loc:distro-resources}/src/main/resources`, which means that it used the very same demo configuration that is also used by the demo packaging in the distro.

Why do we now need a whole repo to replace a single file and how is it related to the demo packaging of the distro?

maggu2810:
It is both, a demo packaging and something that could be run and debugged using the Eclipse IDE + Bndtools.

You can use `mvn clean install` to generate a JAR that contains all -- the OSGi runtime and all configured bundle. This can be started on the command line, in every IDE, debugged etc.
No need to use Eclipse IDE.
So it is also already a demo packaging.
It demonstrates how to create a product with a fixed feature set.
Adding e.g. fileinstall will also make it more usable and extendable.

You also can start the bndrun file in the Eclipse IDE.
This allows you easily to use the debugging integration etc.

We are not using a Eclipse TargetPlatform anymore, we rely on Maven artifacts.
So, instead of a targetplatform file, we have dependencies on the POM, Maven and its repositories.
The artifacts can be provided by a reactor, local Maven repository, external Maven repository, ...

As written above, also the application is a build artifact of a Maven project, so versioned etc.

The demo application consumes dependencies of different repositories, the core one, the addons one, the UI ones, ...

A separate repository without introducing dependencies from one repo to the other one.

Also it keeps it clean, easy to understand, easy as a demo application, easy to use and maintain.

I don't need to think about which is broken because a version, plugin etc. in a pom is changed as no one depends on this project, it consumes only other ones.

kaikreuzer:
> So it is also already a demo packaging.
> It demonstrates how to create a product with a fixed feature set.

Hm, as such, it imho should merely be a part of openhab-core (as it is already). "creating a product" is not a use case outside of openhab-core as the rest of the repos is about openHAB as a solution, not as a framework.

When I refer to the "openHAB demo packaging" i, I mean https://www.openhab.org/docs/configuration/packages.html#demo-package-sample-setup by it.

And this is what the current IDE provides: It lets you launch openHAB with the demo files (items, things, rules, etc.) of the demo package from the distro. This let's developers easily test this setup and if this isn't possible anymore, I am pretty certain that it will quickly break, because nobody will ever test it anymore...

I don't really think we need yet another demo application, which needs to be additionally maintained.

That's why I was asking above whether there is a bnd way of having something as simple as the currently existing launch config for starting the demo packaging.

maggu2810:
My idea has been that openhab-demo will be the one and only one demo that is used by every IDE setup regardless if you are using core only, addons only, web ui only, core + addons + webui, ...
The setup file can be modified that this repo is cloned for every repo configuration similar to openhab-distro at the moment.
As core should not depend on stuff of the addons repo and webui repo, I still don't know how you would get it working.

But as already written in the other repo:
I am aware that I am not an distro or product maintainer, so really okay for me if you found a solution that better fits to your setup -- as long as you don't pollute core with dependencies to addons, UI or whatever.

kaikreuzer:
>  if you found a solution that better fits to your setup

Err, well, what I am saying is that there IS a nicely working solution that is just being destroyed and apparently not replaced by moving the addons to bnd.

I just wanted to highlight to everyone who is working on moving stuff to bnd that it would be a shame to lose this. If I am the only one thinking this, then so be it. (/cc @openhab/2-x-add-ons-maintainers).

maggu2810:
I don't understand the "apparently not replaced" comment.

This is exactly what we are talking about here.
About a replacement that works for all migrated stuff (it can only work for non migrated ones as long as the referenced coordinates are up-to-date -- but that's not my target).

So, you will loose the current launch configuration and can use another one.
I don't think it is anything anybody needs to shame about.

kaikreuzer:
The feature I was talking about is the [one mentioned above](https://github.com/openhab/openhab-distro/pull/889#issuecomment-471438955):

> When I refer to the "openHAB demo packaging" i, I mean https://www.openhab.org/docs/configuration/packages.html#demo-package-sample-setup by it.
> And this is what the current IDE provides:
> It lets you launch openHAB with the demo files (items, things, rules, etc.) of the demo package from the distro. This let's developers easily test this setup and if this isn't possible anymore

Maybe I have missed it, but I didn't see any mention of how this demo package relates to the "demo app" and how those could be single-sourced or kept in sync. Or is your assumption that the demo package of the distro should be dropped completely?

maggu2810:
The configuration does not fit exactly to the one in that documentation but that's just a minor modification of what is actually provided:

There is a log configuration file: https://github.com/maggu2810/openhab-demo/blob/master/app/runtime/logback.xml

and also other configuration files (things, items, ...): https://github.com/maggu2810/openhab-demo/tree/master/app/runtime/conf

There is a selection of bundles (e.g. the Paper UI): https://github.com/maggu2810/openhab-demo/blob/master/app/app.bndrun#L50-L63

and you can also create separate branches to show other great features like Jython scripting which I try to support Scott at the moment a little bit (here the jython branch of openhab-demo): https://github.com/maggu2810/openhab-demo/blob/jython/app/app.bndrun#L67-L77

So, perhaps the current master branch of openhab-demo does not contain exactly the demo package sample setup of the openHAB documentation but perhaps we can try to use that configuration in the master or a demo branch...

maggu2810:
> Or is your assumption that the demo package of the distro should be dropped completely?

>> My idea has been that openhab-demo will be the one and only one demo that is used by every IDE setup 

So, yes. IMHO there should be only one (sometime, if we don't need to provide an easy to use PDE based one anymore)

kaikreuzer:
I think you still haven't understood my point. I am NOT talking about the IDE. I am talking about the [binary distro](https://ci.openhab.org/job/openHAB-Distribution/lastSuccessfulBuild/artifact/distributions/openhab/target/openhab-2.5.0-SNAPSHOT.zip), which offers the demo on [this screen](https://www.openhab.org/docs/configuration/packages.html#demo-package-sample-setup).

maggu2810:
Perhaps I don't understand your point.
Does the binary distribution need to use Karaf?

> You can use `mvn clean install` to generate a JAR that contains all -- the OSGi runtime and all configured bundle. This can be started on the command line, in every IDE, debugged etc.
No need to use Eclipse IDE.
So it is also already a demo packaging.

The binaries and run script are already provided out of the box by the Bnd Maven plugins.

The creation of a zip that contains also the configuration files should be no big pain (I hope).

I have to admit that I did not check yet how to use an environment variable for the launcher (e.g. like $OPENHAB_CONF) but my assumption has been that would be just some try and error, reading sources or documentation or talking with the Bnd guys.

The OSGi runtime environment that contains all configured bundles etc. is already created and can be started by:

```
java -jar target/app.jar
```

So, what's missing except the variable configuration location (using e.g. OPENHAB_CONF)?

I really do not get the very critical point.

maggu2810:
But I don't have to convince anyone.
And you don't have to be convinced of me.

I assume both of us knows a lot more useful things to do your our time.

So, I will maintain my openhab-demo in my repo and keep it unrelated to the demo of openHAB distro.

kaikreuzer:
> Perhaps I don't understand your point.

Ok, let me try again:

1. What we had so far is the official openHAB runtime (binary distro, be it installed from the zip, through apt, rpm or however) that comes with a demo package that can be selected upon very first setup (did you already try it? You should end up with an installation similar to http://demo.openhab.org:8080).

The configuration files that are used by this setup are located at https://github.com/openhab/openhab-distro/tree/master/features/distro-resources/src/main/resources.

2. The PDE launch config allowed to run an openHAB demo instance from within the IDE that uses exactly the same resources from https://github.com/openhab/openhab-distro/tree/master/features/distro-resources/src/main/resources.

So these files are being used - and thus validated - by every developer running openHAB from the IDE.
If those files are in future not used anymore by anyone during development, I am pretty sure that stuff will quickly break without anybody noticing. This is the situation that I want to avoid.

I see 3 options:
1. Continue to use the same config files for the demo in the IDE and the distro.
2. Drop the demo setup from the distro completely

> So, I will maintain my openhab-demo in my repo and keep it unrelated to the demo of openHAB distro.

I hope it is now clear that this is not a solution. By moving the addons to bnd and switching from a PDE launch file to a bndrun, the existing feature (1) breaks. So I hoped you have a way to continue doing (1) even with the bndrun. The only alternative to that is imho (2), that's why I asked you [above](https://github.com/openhab/openhab-distro/pull/889#issuecomment-471554331) whether this is your assumption on how to go about it.

maggu2810:
> So I hoped you have a way to continue doing (1) even with the bndrun.

I don't see anything that is not easily doable (just a time invest).

* openhab-demo currently consumes stuff from openhab-core, openhab-addons and openhab-ui
* openhab-distro currently consumes stuff from openhab-core, openhab-addons and openhab-ui (perhaps more than that, but that's not the matter).

So, if we change openhab-distro to consume also one artifact that is created by openhab-demo this should not be a "real" problem, because openhab-demo does not consume anything of openhab-distro and so there will be no cycle.

The configuration files (all below `./features/distro-resources/src/main/resources`) of openhab-distro can be moved to openhab-demo.
openhab-demo can referene that files easily while running in the Eclipse IDE or on command line.

openhab-demo could create an artifact that contains all files of the configuration directory -- not more, not less.

openhab-distro will fetch that artifact and build time, unpack its content and proceed as before to create its distribution.

kaikreuzer:
> I don't see anything that is not easily doable (just a time invest).

Ok, sounds good - so I think it would be nice if we'd head that way.

As our IDE already checks out openhab-distro in any case and the distro repo currently contains the demo parts, I'd think it should be fine to keep the demo within this repo - I'm not really sure whether its worth to have a dedicated repo for it. @openhab/distro-maintainers wdyt?

wborn:
I would welcome a bnd based demo app to this repo. :-)

davidgraeff:
From someone that is not coming from the java world:

It took me an entire day to figure out how the pde/targetplatform stuff works and it was not obvious nor easy to understand in the eclipse version of that time (I think two years ago or so) and fragile on top.

Yesterday I started with openhab-core and maven/bnd as well as maven/bndtools for eclipse.
It took two days before I realized what is going on and how to do the simplest things, a bad reminder of my first contact with openHAB development. And it doesn't help that two build systems are in place at the moment to understand the overall structure.

I would have been totally lost without the bndrun file in the core repo (demo bundle). But I struggled to do binding development in the same IDE with that run file (no reference to anything outside of core, so no paperui etc).

Markus demo repository was the salvation actually. You import his repo as maven project, selectively add your openhab2-addon bundles and start developing. I had never setup an openHAB development environment faster and actually understood what is happening, nothing hidden in lengthy oomph setup scripts.

Regarding the "demo package" (demo thing, item, sitemap, rules files): The very first thing I did for development was to remove those. They spam your log view. They might be useful for the binary distro but are a burden for developers. My 2ct.

My personal wish as a contributor: Please provide a developer distro repo that does not include the "demo package" and that includes a bndrun file for a point and click like development start.

J-N-K:
I fully agree with that.

maggu2810:
It seems there are different views.

> It lets you launch openHAB with the demo files (items, things, rules, etc.) of the demo package from the distro. This let's developers easily test this setup and if this isn't possible anymore, I am pretty certain that it will quickly break, because nobody will ever test it anymore...

@kaikreuzer I see the point that if we don't provide the same configuration in the IDE less people use it.

But I don't care about that configuration if I would like to develop, test, debug for openHAB Core (or addons or whatever).
I would like to use a very simple starting point to get an instance running.

For me a separate repo makes also sense to provide e.g. different setups (a minimal branch, a branch to test e.g. jython, a branch for ..., a full blown setup, ...).
I would like to switch between branches easily without care about a lot of other things that are part of that repo.

My personal feeling is that there is no need that the openHAB repositories are such interlocked.
With the migration to Maven + Bnd the situation is much better.
IMHO as soon as there is no p2 anymore, addons does not consume a parent POM of core (the PDE addons do it), webui does not consume a parent POM of core (there is no need, similar as migrated addons does use a core parent) things will get more relaxed.

With openhab-demo (as @davidgraeff already written) I just need to use that projects and the project I would like to work on. It is fast and easy as soon as you know how pure Maven stuff is working in Eclipse and perhaps a little knowledge about Bnd (a little bit knowledge of OSGi while working on OSGi bundles can be provided).

The main missing part are the DS requirements.
It would make sense to migrate the addons that fail because of the requirements first.

PR #897, opened 2019-03-26
Don't copy old backups to /tmp

Jamstah:
This patch will filter out the copying of old backups before they reach /tmp. If the backups folder isn't in the userdata folder, then the argument will be ignored by find, and there are no bad effects.

On my system, where the backups are in userdata, it was copying GBs of old backups to `/tmp` and then deleting them. `/tmp` is often in memory or quite small, so this should be avoided. It also speeds up my backups significantly.

Signed-off-by: James Hewitt-Thomas <james.hewitt@gmail.com>

### DIFF
--- a/distributions/openhab/src/main/resources/bin/backup
+++ b/distributions/openhab/src/main/resources/bin/backup
@@ -130,9 +130,9 @@ setup "$1" "$2"
 echo "Copying configuration to temporary folder..."
 mkdir -p "$TempDir/userdata"
 if [ -z "$INCLUDE_CACHE" ]; then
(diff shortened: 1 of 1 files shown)


Jamstah:
I've been trying to get a green build for this, but I think the executor is out of disk:
```
16:56:01 java.io.IOException: No space left on device
16:56:01     at java.io.FileOutputStream.writeBytes (Native Method)
16:56:01     at java.io.FileOutputStream.write (FileOutputStream.java:326)
```

kaikreuzer:
@BClark09 Would you want to review this change?

PR #906, opened 2019-05-04
remove usage of ohc parent pom

maggu2810:
The openHAB Distro repo is the only one I know about that is still using a POM of openHAB Core "feature/p2/poms" directory (or subdirectory).

Please check if all used stuff has been moved to this POM (I have tried it but I don't know if I succeed).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,16 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?><project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>org.openhab</groupId>
-        <artifactId>pom</artifactId>
-        <version>2.5.0-SNAPSHOT</version>
-        <relativePath/>
-    </parent>
-
     <groupId>org.openhab.distro</groupId>
     <artifactId>pom</artifactId>
-    <!-- do not delete the version here as it is required by the release process -->
     <version>2.5.0-SNAPSHOT</version>
     <packaging>pom</packaging>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1,16 +1,8 @@
...
-    <parent>
-        <groupId>org.openhab</groupId>
-        <artifactId>pom</artifactId>
-        <version>2.5.0-SNAPSHOT</version>
-        <relativePath/>
-    </parent>
-
     <groupId>org.openhab.distro</groupId>
     <artifactId>pom</artifactId>
-    <!-- do not delete the version here as it is required by the release process -->
```
kaikreuzer: Why did you remove this notice?

pom.xml
```diff
@@ -185,7 +180,7 @@
                     <plugin>
                         <groupId>org.eclipse.tycho</groupId>
                         <artifactId>tycho-versions-plugin</artifactId>
-                        <version>${tycho.version}</version>
+                        <version>1.4.0</version>
```
kaikreuzer: If no tycho stuff is left, we probably also won't need the tycho-versions-plugin anymire - I guess the "prepare-release" profile can be removed (it was used for the old manual release process).

pom.xml
maggu2810: Dropping the version of a POM makes only sense if you use a parent one.
If you don't use a parent POM, the version will not be removed at all.
So no need to state it explicit.

Or can you tell me the version of that coordinate if no parent is used and no one is set. :wink:

pom.xml
maggu2810: The execution ID `update-version` has been configured to execute the `set-version` goal.
So, I assume that function is used anywhere.

pom.xml
kaikreuzer: Good point 🙄

pom.xml
kaikreuzer: I doubt it, but ok, no need to remove it in this PR.

wborn:
Can you fix the [build error](https://ci.openhab.org/job/PR-openHAB-Distribution/89/console)?

`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install (default-install) on project distro: NoFileAssignedException: The packaging plugin for this project did not assign a main file to the project but it has attachments. Change packaging to 'pom'. -> [Help 1]`

We ran into it before and fixed it by pinning the maven-install-plugin to 2.5.2. 

See:
* https://github.com/openhab/openhab-core/pull/408
* [MINSTALL-151](https://issues.apache.org/jira/browse/MINSTALL-151)

maggu2810:
Sure, done.
Let's wait and check the build.

PR #909, opened 2019-05-10
Use UI bom and include relevant bundles for demo app

kaikreuzer:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ maven-metadata-local.xml
 launch/home/logs
 launch/home/userdata
 Californium.properties
+launch/app/runtime/userdata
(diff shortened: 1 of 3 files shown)


kaikreuzer:
If I look at https://ci.openhab.org/job/PR-openHAB-Distribution/, this is not the only PR that fails with that message 😠...

kaikreuzer:
I've reconfigured the PR build plan to
- use a private Maven repo
- to do full, not incremental builds
- to have up to 1GB heap space instead of 512MB
- to use our Artifactory to resolve dependencies

Let's see if that helps :-)

wborn:
Thanks! It helped. :-)

PR #912, opened 2019-05-18
fix item setup

J-N-K:
see https://github.com/openhab/openhab2-addons/pull/5610

Signed-off-by: Jan N. Klug <jan.n.klug@rub.de>

### DIFF
--- a/features/distro-resources/src/main/resources/items/demo.items
+++ b/features/distro-resources/src/main/resources/items/demo.items
@@ -127,7 +127,7 @@ Number Wifi_Level               "Wifi Level [%d/4]"                 <signal>    
 String ImageURL                 "Image URL [%s]"
 
 Group Thermostat                "Thermostat"            <temperature>                    ["Thermostat"]
-Dimmer Temperature_Setpoint     "Temperature [%.1f °C]" <temperature>    (Thermostat)    ["TargetTemperature", "Setpoint", "Temperature"] { habot="" [ control="knob", min=15, max=35, leftIcon="mdi-thermometer", leftInverted=true ] }
-String Thermostat_Mode          "Thermostat Mode [%s]"  <temperature>    (Thermostat)    ["homekit:HeatingCoolingMode", "Control"]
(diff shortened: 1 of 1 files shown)


features/distro-resources/src/main/resources/items/demo.items
```diff
@@ -127,7 +127,7 @@ Number Wifi_Level               "Wifi Level [%d/4]"                 <signal>
 String ImageURL                 "Image URL [%s]"
 
 Group Thermostat                "Thermostat"            <temperature>                    ["Thermostat"]
-Dimmer Temperature_Setpoint     "Temperature [%.1f °C]" <temperature>    (Thermostat)    ["TargetTemperature", "Setpoint", "Temperature"] { habot="" [ control="knob", min=15, max=35, leftIcon="mdi-thermometer", leftInverted=true ] }
-String Thermostat_Mode          "Thermostat Mode [%s]"  <temperature>    (Thermostat)    ["homekit:HeatingCoolingMode", "Control"]
+Number Temperature_Setpoint     "Temperature [%.1f °C]" <temperature>    (Thermostat)    ["homekit:TargetTemperature", "Setpoint", "Temperature"] { habot="" [ control="knob", min=15, max=35, leftIcon="mdi-thermometer", leftInverted=true ] }
+String Thermostat_Mode          "Thermostat Mode [%s]"  <temperature>    (Thermostat)    ["homekit:TargetHeatingCoolingMode", "Control"]
```
kaikreuzer: Just to make sure that our migration info is up to date: When was `homekit:HeatingCoolingMode` renamed to `homekit:TargetHeatingCoolingMode`? Already before 2.5? If not, do we have a notice in the release notes?

J-N-K:
It‘s not exactly breaking because the old version is still supported (only deprecated). I can add a note to the release notes if you want.

kaikreuzer:
Ok, thanks for the info. Yes, I think a note nonetheless makes sense as it will allow us to also remove the old string in some future version.

PR #915, opened 2019-05-19
fix javax.mail logger

J-N-K:
javax.mail logs a WARN if `javamail.default.address.map` is not found. Unfortunately it is neither possible to put that file in a location where it can be found (in an OSGi-environment) nor to stop javax.mail to look for this file (which in fact is not needed). The suggested method is to suppress the warning by an appropriate logger configuration.

See https://github.com/openhab/openhab2-addons/issues/5530

Signed-off-by: Jan N. Klug <jan.n.klug@rub.de>

### DIFF
--- a/distributions/openhab/src/main/resources/userdata/etc/org.ops4j.pax.logging.cfg
+++ b/distributions/openhab/src/main/resources/userdata/etc/org.ops4j.pax.logging.cfg
@@ -85,6 +85,10 @@ log4j2.logger.lsp4j.level = OFF
 log4j2.logger.karservice.name = org.apache.karaf.kar.internal.KarServiceImpl
 log4j2.logger.karservice.level = ERROR
 
+# Filters known issues of javax.mail, see
+# https://github.com/openhab/openhab2-addons/issues/5530
+log4j2.logger.javaxmail.name = javax.mail
+log4j2.logger.javaxmail.level = ERROR
 
 # Appenders configuration
 


stagf15:
I just loaded OpenHAB2.5 Snapshot #1655 and subsequently found out about the mail action-binding changes.  After stumbling around I got it configured properly and it does send mail, but I am still getting this error:

[WARN ] [javax.mail] - expected resource not found: /META-INF/javamail.default.address.map

Should Snapshot #1655 have the fix mentioned above, or will it be forthcoming?  Thanks!

kaikreuzer:
Your file `org.ops4j.pax.logging.cfg` might not be overwritten (depending on how you did the upgrade). You can simply add these lines manually and the logging should be gone.

stagf15:
@kaikreuzer, Thank you for the response!  That was it.  I have a script for windows (based on the script for linux) that will delete the necessary files (and backup) prior to an upgrade, but it doesn't delete `org.ops4j.pax.logging.cfg` so I was not getting the newest version when I upgraded.

_Not related to this post_, but I've found a significant logging problem in OpenHAB2.5 Snapshot #1655 and the milestone build on Win10 with JSR223 loaded, if you have a moment, please check out here: [https://community.openhab.org/t/jsr223-jython-breaks-karaf-or-logging/80351](url) and the same issue by another user here: [https://community.openhab.org/t/windows-all-oh-logs-stop-when-jython-is-in-the-classpath/79654](url)

PR #929, opened 2019-06-26
fix some demo app stuff

maggu2810:


maggu2810:
* do not use Java 1.5 but at least 1.8
* as the demo app do not use any parent, there has no configuration been delegated for all the Maven plugins
* update the snapshots only ever 24h from the remote instead of all the time

kaikreuzer:
Thanks!

PR #933, opened 2019-06-29
Add ZWave binding

cdjackson:

### DIFF
--- a/launch/app/pom.xml
+++ b/launch/app/pom.xml
@@ -32,7 +32,6 @@
       <type>pom</type>
       <scope>runtime</scope>
     </dependency>
-
     <dependency>
       <groupId>org.openhab.addons.bom</groupId>
       <artifactId>org.openhab.addons.bom.openhab-addons</artifactId>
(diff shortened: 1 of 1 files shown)


launch/app/pom.xml
```diff
@@ -32,6 +32,11 @@
       <type>pom</type>
       <scope>runtime</scope>
     </dependency>
+     <dependency>
```
kaikreuzer: identation looks wrong here

kaikreuzer: Could you add it AFTER all the bom dependencies?

launch/app/pom.xml
```diff
@@ -32,6 +32,11 @@
...
+      <groupId>org.openhab.addons.bundles</groupId>
+      <artifactId>org.openhab.binding.zwave</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
kaikreuzer: Not sure if we should also add the runtime scope as for the others?

cdjackson:
Updated

launch/app/pom.xml
```diff
@@ -32,14 +32,22 @@
...
       <version>${project.version}</version>
       <type>pom</type>
       <scope>runtime</scope>
     </dependency>
+    
+    <dependency>
+      <groupId>org.openhab.addons.bundles</groupId>
+      <artifactId>org.openhab.binding.zwave</artifactId>
+      <version>${project.version}</version>
+      <type>pom</type>
```
kaikreuzer: This artifact is not a pom - so this line should not be added.

PR #936, opened 2019-07-01
Add Zigbee binding to App BOM

triller-telekom:
Signed-off-by: Stefan Triller <stefan.triller@telekom.de>

### DIFF
--- a/launch/app/pom.xml
+++ b/launch/app/pom.xml
@@ -48,14 +48,20 @@
       <type>pom</type>
       <scope>runtime</scope>
     </dependency>
-    
+    <dependency>
+      <groupId>org.openhab.addons.bom</groupId>
+      <artifactId>org.openhab.addons.bom.zigbee</artifactId>
+      <version>${project.version}</version>
+      <type>pom</type>
+      <scope>runtime</scope>
+    </dependency>
+
     <dependency>
       <groupId>org.openhab.addons.bundles</groupId>
       <artifactId>org.openhab.binding.zwave</artifactId>
       <version>${project.version}</version>
       <scope>runtime</scope>
     </dependency>
-    
   </dependencies>
 
   <build>


launch/app/pom.xml
```diff
@@ -55,7 +55,14 @@
       <version>${project.version}</version>
       <scope>runtime</scope>
     </dependency>
-    
+
+    <dependency>
+      <groupId>org.openhab.addons.bundles</groupId>
+      <artifactId>org.openhab.binding.zigbee</artifactId>
```
wborn: The BOM added in https://github.com/openhab/org.openhab.binding.zigbee/pull/453 has a different artifactId:

```suggestion
      <artifactId>org.openhab.binding.zigbee.bom</artifactId>
```

triller-telekom: Thanks, I have force pushed the correction.

triller-telekom:
However, now eclipse complain that it cannot find the dependency...

launch/app/pom.xml
```diff
@@ -55,7 +55,14 @@
       <version>${project.version}</version>
       <scope>runtime</scope>
     </dependency>
-    
+
+    <dependency>
```
kaikreuzer: I'd suggest to move the bom dependency up to the other bom dependencies (zwave is the only exception and should be at the end.

launch/app/pom.xml
```diff
@@ -55,7 +55,14 @@
...
+      <groupId>org.openhab.addons.bundles</groupId>
```
kaikreuzer: We should sync the GAV naming scheme with the other boms - the groupID ends with "bom" there.
```suggestion
      <groupId>org.openhab.addons.bom</groupId>
```

launch/app/pom.xml
```diff
@@ -55,7 +55,14 @@
...
+      <artifactId>org.openhab.binding.zigbee.bom</artifactId>
```
kaikreuzer:
```suggestion
      <artifactId>org.openhab.addons.bom.zigbee</artifactId>
```

triller-telekom:
I have updated this PR and created https://github.com/openhab/org.openhab.binding.zigbee/pull/455 to change the bom name in the zigbee binding again (I hope I didn't mess anything up, I am new to this whole bnd bom magic).

triller-telekom:
Since https://github.com/openhab/org.openhab.binding.zigbee/pull/455 has been merged now, can someone please have a look at this PR if it now correctly integrates the zigbee binding and if it does, also merge it?

launch/app/pom.xml
```diff
@@ -48,14 +48,19 @@
       <type>pom</type>
       <scope>runtime</scope>
     </dependency>
-    
+    <dependency>
+      <groupId>org.openhab.addons.bom</groupId>
+      <artifactId>org.openhab.addons.bom.zigbee</artifactId>
+      <version>${project.version}</version>
+      <scope>runtime</scope>
```
kaikreuzer: Just noticed another difference to the other bom dependencies:
```
<type>pom</type>
```
is missing here.

Wrt testing: Can't you (or @cdjackson)  test whether it is possible to use the Zigbee binding from within the IDE with that? After all, you want to use it so it is most important that it's working for you :-)

triller-telekom: In th zwave PR you mentioned that zwave is not a pom. So where is the difference?

https://github.com/openhab/openhab-distro/pull/933#discussion_r298796665

kaikreuzer: The difference is that this is a pom. (in z-wave we are referencing the jar artifact, not a bom)

triller-telekom: If I add the `<type>pom</type>` eclipse does not complain anymore.

However, I cannot select the zigbee binding as a bundle to be started within `app.bndrun`.

triller-telekom:
Since Eclipse does not complain about this dependency anymore, I wonder what has to be done to make the zigbee binding appear in the list of to be started bundles in `app.bndrun`.

I thought adding this dependency is enough.

launch/app/pom.xml
triller-telekom: I had added the type again.

PR #932, opened 2019-06-28
migrate from EclipseSource JAX-RS publisher to openHAB one

maggu2810:
This PR requires the related change in openHAB Core.

Related to: https://github.com/openhab/openhab-core/pull/905

PR #944, opened 2019-07-16
Revert "migrate from EclipseSource JAX-RS publisher to openHAB one (#932)"

kaikreuzer:
This reverts commit a4e5cbd7a4713360bc9a29ab5adca5e93d5b4e37.

See https://github.com/openhab/openhab-core/issues/924#issuecomment-511787620

### DIFF
--- a/launch/app/app.bndrun
+++ b/launch/app/app.bndrun
@@ -32,7 +32,7 @@ feature.openhab-base: \
     bnd.identity;id='org.openhab.core.io.net',\
     bnd.identity;id='org.openhab.core.io.http',\
     bnd.identity;id='org.openhab.core.io.rest',\
-    bnd.identity;id='org.openhab.core.io.rest.publisher',\
+    bnd.identity;id='org.openhab.core.io.rest.optimize',\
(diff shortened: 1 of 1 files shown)


maggu2810:
You are missing the Signed-off-by line in the commit message.

kaikreuzer:
I used Gits "reverse commit" feature, which does not add any sign-offs (as they don't make much sense since the commit does not include any copyrightable content from me).

Will check, why this PR is now in conflict...

kaikreuzer:
I used Gits "reverse commit" feature, which does not add any sign-offs (as they don't make much sense since the commit does not include any copyrightable content from me).

Will check, why this PR is now in conflict...

kaikreuzer:
Conflict resolved.

maggu2810:
> I used Gits "reverse commit" feature, which does not add any sign-offs (as they don't make much sense since the commit does not include any copyrightable content from me).

Already answered to the other topic. If you use "-s" on `git revert` it will add the sign off header similar to "-s" on `git commit`.
AFAIK every code contribution regardless if it adds code that has been already part some time of the repo or new one should be signed off.

PR #963, opened 2019-09-02
Some small improvements to the IDE setup

kaikreuzer:
As discussed in https://github.com/openhab/openhab-distro/issues/927

### DIFF
--- a/launch/app/app.bndrun
+++ b/launch/app/app.bndrun
@@ -75,7 +75,9 @@ feature.openhab-model-runtime-all: \
 	osgi.console.enable.builtin=false,\
 	logback.configurationFile=file:${.}/runtime/logback.xml
 
--runblacklist: bnd.identity;id='org.apache.aries.javax.jax.rs-api'
+-runblacklist: bnd.identity;id='org.apache.aries.javax.jax.rs-api',\
+    bnd.identity;id='org.apache.aries.jpa.container',\
+    bnd.identity;id='org.openhab.core.test'
 
 #
 # done
--- a/launch/app/app.bndrun
+++ b/launch/app/app.bndrun
@@ -75,9 +75,7 @@ feature.openhab-model-runtime-all: \
 	osgi.console.enable.builtin=false,\
 	logback.configurationFile=file:${.}/runtime/logback.xml
 
--runblacklist: bnd.identity;id='org.apache.aries.javax.jax.rs-api',\
-    bnd.identity;id='org.apache.aries.jpa.container',\
-    bnd.identity;id='org.openhab.core.test'
+-runblacklist: bnd.identity;id='org.apache.aries.javax.jax.rs-api'
 
 #
 # done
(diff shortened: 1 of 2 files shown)


launch/app/app.bndrun
```diff
@@ -75,7 +75,9 @@ feature.openhab-model-runtime-all: \
 	osgi.console.enable.builtin=false,\
 	logback.configurationFile=file:${.}/runtime/logback.xml
 
--runblacklist: bnd.identity;id='org.apache.aries.javax.jax.rs-api'
+-runblacklist: bnd.identity;id='org.apache.aries.javax.jax.rs-api',\
+    bnd.identity;id='org.apache.aries.jpa.container',\
+    bnd.identity;id='org.openhab.core.test'
```
Hilbrand: Maybe also add `    bnd.identity;id='org.openhab.core.io.transport.serial.javacomm'` to the blacklist. 
It gives me the warning:
```
 [org.openhab.core.io.transport.serial.javacomm version=2.5.0-SNAPSHOT]
      ⇒ osgi.wiring.package: (&(osgi.wiring.package=javax.comm))
```

launch/app/app.bndrun
```diff
@@ -211,4 +212,5 @@ feature.openhab-model-runtime-all: \
 	org.threeten.extra;version='[1.4.0,1.4.1)',\
 	slf4j.api;version='[1.7.25,1.7.26)',\
 	tec.uom.lib.uom-lib-common;version='[1.0.3,1.0.4)',\
-	tec.uom.se;version='[1.0.10,1.0.11)'
+	tec.uom.se;version='[1.0.10,1.0.11)',\
+	slf4j.simple;version='[1.7.21,1.7.22)'
```
maggu2810: I don't think slf4j-simple should be used by the OSGi container.
I also don't think slf4j-simple should be part of the dependency tree of any OH OSGi bundles at all.

launch/app/app.bndrun
maggu2810: I don't think the correct approach is to blacklist OH bundles that dependencies are currently not part of the index.

If stuff is missing, don't at it to the requirements or add the dependencies to the index.

launch/app/app.bndrun
kaikreuzer: I didn't touch this section, it is automatically filled when doing a "resolve" - so I committed the change in order to avoid people seeing local changes in their IDE.
Are you saying that bnd does NOT add this line, when you run the resolution?

launch/app/app.bndrun
kaikreuzer: Well, if I do a "resolve", one of the messages is that the test bundle cannot resolve junit.
You are right, the better approach would be to make sure that the test bundle does not at all occur as a run requirement - but as I have no clue why it does, I decided to simply put it to the blacklist.
Same for the jpa container - I don't know, why it appears, but I do know that we neither want nor need it, hence I added it to the blacklist.

launch/app/app.bndrun
maggu2810: The resolve logic is using the index generated by the dependencies.

As written above slf4j-simple should not be part of the index / dependency tree at all. If bnd does not know about a slf4j-simple bundle it cannot add them.

The runtime needs "some" implementation of the slf4j API.
So "resolve" may choose one of all the implementation that are part of the index.
If could be the one this time, another one the next time. It depends on the wiring at all.

Hilbrand:
The changes made here in `app.bndrun` seem to need some extra research. I also have the feeling it depends on what additional projects are added to the app. So in order to get at least the changes in the pom.xml merged can we split this pr? Because those changes are very relevant for the users and would relief us from mentioning it every time someone needs help with the ide.

kaikreuzer:
@Hilbrand Done. This PR now only contains the changes in the pom.xml.

kaikreuzer:
Additionally created https://github.com/openhab/openhab-distro/pull/969

PR #965, opened 2019-09-04
[tplinksmarthome] Added 2.5.0 breaking changes alert

Hilbrand:
Related to https://github.com/openhab/openhab2-addons/pull/6020

Also changed case of sensebox as it was the only one that didn't start with an uppercase character.


HerzScheisse:
@Hilbrand 

small typo... i guess it should read "energy" ?! :)

`The enery and rssi channels`

hakan42:
aaah, senseBox was actually intentionally lowercased. See

https://sensebox.de/

PR #969, opened 2019-09-06
added a few more bundles to the blacklist

kaikreuzer:

### DIFF
--- a/launch/app/app.bndrun
+++ b/launch/app/app.bndrun
@@ -75,7 +75,9 @@ feature.openhab-model-runtime-all: \
 	osgi.console.enable.builtin=false,\
 	logback.configurationFile=file:${.}/runtime/logback.xml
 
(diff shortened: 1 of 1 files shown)


wborn:
Should we still merge this PR and will it fix/prevent issues @kaikreuzer ? 
I haven't closely followed the "IDE setup" discussion. :male_detective:

kaikreuzer:
It clearly prevented issues in my setup and I would expect the same for others.
What I wonder is why there is so little feedback if others see it or if it is all ok. This makes me believe that meanwhile everybody has created his setup by trial-and-error over time and doesn't dare to touch anything anymore. Giving us all a slightly different environment, which isn't a good situation at all, but what to do...

wborn:
Yes I also don't like to fiddle around with my Eclipse too much since it's such a major time killer when it breaks. I've also been using m2e for quite a long time so that also helps when resolving local build issues. The m2e performance issues when opening many projects are also not specific to openHAB.

PR #972, opened 2019-09-25
update version of com.zsmartsystems.zigbee to 1.2.3

sesuncedu:
The version of the zigbee libraries in use is out-of-date.

### DIFF
--- a/features/addons/src/main/feature/feature.xml
+++ b/features/addons/src/main/feature/feature.xml
@@ -6,11 +6,11 @@
     <feature name="openhab-binding-zigbee" description="ZigBee Binding" version="${project.version}">
         <feature>openhab-runtime-base</feature>
         <feature>openhab-transport-serial</feature>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee/1.2.1</bundle>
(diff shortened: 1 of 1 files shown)


cdjackson:
Sorry @cweitkamp - I didn't see this until it was merged just now.

I thought that one of the advantages of the new system with bnd is that all dependencies were now managed in one place (ie in the binding repository) and we didn't need to update these other files now? I assume I'm wrong with that?

cweitkamp:
@cdjackson Yes, that is partly true. To archive that we have to move the feature file from this location into the https://github.com/openhab/org.openhab.binding.zigbee/ repository. Similar to what we did with other openHAB 2 add-ons (see https://github.com/openhab/openhab2-addons/pull/5555).

cdjackson:
Ok, thanks. Is there anything stopping us doing that? I could guess there might be something else required to consolidate all the feature files? Or is that already in place and we only need to move the file?

cweitkamp:
I can remember we talked about this some time ago (see https://github.com/openhab/org.openhab.binding.zigbee/pull/454#issuecomment-516131313) but to be honest I do not know if someone started working on it.

cdjackson:
Thanks, but this just says we should move the file into the Zigbee repo - is it just that simple, or is there something else required to ensure that the file from the zigbee repo is incorporated into the system?

PR #979, opened 2019-10-08
Renamed service pid for regional settings from "org.eclipse.smarthome.core.localeprovider" -> "org.eclipse.smarthome.core.i18nprovider" to "org.eclipse.smarthome.i18n"

cweitkamp:
- Renamed service pid for regional settings from "org.eclipse.smarthome.core.localeprovider" -> "org.eclipse.smarthome.core.i18nprovider" to "org.eclipse.smarthome.i18n"

According to our code the service pid is `org.eclipse.smarthome.i18nprovider`. It was not possible for me to find any reference to `org.eclipse.smarthome.core.localeprovider`.

https://github.com/openhab/openhab-core/blob/a44c55e0e00ff4c8a072385097b43c4144489708/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/I18nProviderImpl.java#L81-L83

Depends on https://github.com/openhab/openhab-core/pull/1112

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -47,3 +47,5 @@ DELETEDIR;$OPENHAB_USERDATA/kar
 
 [2.5.0]
 REPLACE;org.eclipse.smarthome.automation.dto.RuleDTO;org.openhab.core.automation.dto.RuleDTO;$OPENHAB_USERDATA/jsondb/automation_rules.json
+MOVE;$OPENHAB_USERDATA/config/org/eclipse/smarthome/core/i18nprovider.config;$OPENHAB_USERDATA/config/org/eclipse/smarthome/i18n.config
+REPLACE;org.eclipse.smarthome.core.i18nprovider;org.eclipse.smarthome.i18n;$OPENHAB_USERDATA/config/org/eclipse/smarthome/i18n.config
--- a/distributions/openhab/src/main/resources/conf/services/runtime.cfg
+++ b/distributions/openhab/src/main/resources/conf/services/runtime.cfg
@@ -4,13 +4,13 @@
 # The ISO 639 alpha-2 or alpha-3 language code (if there is no alpha-2 one).
 # Example: "en" (English), "de" (German), "ja" (Japanese), "kok" (Konkani)
 #
-#org.eclipse.smarthome.core.i18nprovider:language=
+#org.eclipse.smarthome.i18n:language=
 
 # The region that should be used.
 # ISO 3166 alpha-2 country code or UN M.49 numeric-3 area code.
 # Example: "US" (United States), "DE" (Germany), "FR" (France), "029" (Caribbean)
 #
-#org.eclipse.smarthome.core.i18nprovider:region=
+#org.eclipse.smarthome.i18n:region=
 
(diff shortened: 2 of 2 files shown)


kaikreuzer:
This indeed seems to be an outdated configuration which dates from before the time when we split the LocaleProvider into multiple interfaces.

I wonder if we should change the configurationPid completely, so that it is aligned with the standard pattern that we use, which is `org.eclipse.smarthome.<configtype>` (i.e. without "core" and also not using technical terms such as "provider"). Imho the pid should be changed to `org.eclipse.smarthome.i18n`, wdyt?

cweitkamp:
Yes, I think that is a good idea. Can we use the [update.lst](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/update.lst) to move the existing config files to the new location?

wborn:
> Can we use the update.lst to move the existing config files to the new location?

This might work:

```
MOVE;$OPENHAB_USERDATA/config/org/eclipse/smarthome/core/i18nprovider.config;$OPENHAB_USERDATA/config/org/eclipse/smarthome/i18n.config
REPLACE;org.eclipse.smarthome.core.i18nprovider;org.eclipse.smarthome.i18n;$OPENHAB_USERDATA/config/org/eclipse/smarthome/i18n.config
```

cweitkamp:
I modified the update script. If I should add a notice too let me know.

kaikreuzer:
> There may also be users with a file based configuration that need to update their files based on an upgrade notice?

I doubt that there are users with this. As we didn't use the standard layout (so that you can e.g. create a `i18n.cfg` file), I would say that those parameters weren't yet meant to be used through files - therefore we do not need to provide migration docs.

mhilbush:
> I doubt that there are users with this.

Maybe just one user. LOL

```
org.eclipse.smarthome.core.i18nprovider:language=en
org.eclipse.smarthome.core.i18nprovider:region=US
org.eclipse.smarthome.core.i18nprovider:location=39.xxxxxxx,-76.xxxxxxx
org.eclipse.smarthome.core.i18nprovider:timezone=US/Eastern
```

hakan42:
Make that two. I use that to set up my docker image (s), both production and testing of translations 😄

kaikreuzer:
When updating on my Mac, the `MOVE` didn't work either. I found this as an output:
```
Performing post-update tasks for version 2.5.0:
  Replacing: String org.eclipse.smarthome.automation.dto.RuleDTO to org.openhab.core.automation.dto.RuleDTO in file /Applications/openHAB/userdata/jsondb/automation_rules.json
sed: 1: "/Applications/openHAB/u ...": invalid command code o
  Moving:   From /Applications/openHAB/userdata/config/org/eclipse/smarthome/core/i18nprovider.config to /Applications/openHAB/userdata/config/org/eclipse/smarthome/i18n.config
  Replacing: String org.eclipse.smarthome.core.i18nprovider to org.eclipse.smarthome.i18n in file /Applications/openHAB/userdata/config/org/eclipse/smarthome/i18n.config
sed: 1: "/Applications/openHAB/u ...": invalid command code o
```

Issue #668, opened 2018-03-29
Auto-generated release notes


kaikreuzer:
With the past releases, we manually compiled the release notes (partially with the help of a script that @BClark09 put together).

In the light of https://github.com/openhab/openhab-distro/pull/662#issuecomment-375442417, we should come to a more automated process that should be based on the information in the update.xyz  file as well as on labels on merged PRs from Github, where we should formalize the labels across the different repos and the style of the PR text, so that they are useful when ending up in the release notes.

yveshanoulle:
I assume this is only major releases? Or is there a way to know about minor releases? 

Aka I have now openHAB 2.4.0 Build #1308 installed. is there a way to know what has changed since a previous installed version?

wborn:
> formalize the labels across the different repos and the style of the PR text

I think the way how issues and PRs are labeled in ESH would be a good starting point for this. Such labels would allow for grouping most PRs as is currently done in the current release notes.

> I have now openHAB 2.4.0 Build #1308 installed. is there a way to know what has changed since a previous installed version?

Currently that's not easy. An easy (but not totally correct) approach could be to determine when `#1308` was build (e.g. based on the timestamps of the OH files) and then see what commits in the repositories were made since that timestamp.

kaikreuzer:
@BClark09 As we are approaching a 2.4.0.M1 build, would it be worth to move the current version of your script to this repo here and maybe even have a Jenkins plan that executes it? This could help to see how well we are doing with the PR labelling and once we are satisfied with the result, we could copy it over to the release descriptions, wdyt?

yveshanoulle:
> Currently that's not easy. An easy (but not totally correct) approach could be to determine when `#1308` was build (e.g. based on the timestamps of the OH files) and then see what commits in the repositories were made since that timestamp.

It would be nice if 
 - the openhab webpage would also add the date of the build (next to the build number)
- somewhere on openhab website we have a list of all builds with their date
==> this way we don't have to go to the openhab server to check timestamps.

kaikreuzer:
This is now in place with https://ci.openhab.org/view/Utility/job/openhab-build-changelog/ - at least as a semi-automated process (that usually still takes some manual steps for the final result).

PR #982, opened 2019-10-14
[networkupstools] openHAB1 binding moved to legacy

Hilbrand:
Signed-off-by: Hilbrand Bouwkamp <hilbrand@h72.nl>

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -26,6 +26,7 @@ ALERT;Homekit: Some tags have been renamed. Old names will not be supported in f
 ALERT;LGWebOS Binding: The binding parameter 'localIP' has been removed. The binding now uses system defaults for network communication. The thing type parameter 'deviceId' is no longer a parameter, but a property. Parameters 'host' and 'key' have been added.
 ALERT;Mail Action: The mail action has been replaced by a new mail binding.
 ALERT;MQTT Binding: Homie channel names may have changed if special characters are used for MQTT topic names.
+ALERT;Network UPS Tools: The Network UPS Tools openHAB 1 binding has been replaced by a new binding and the openHAB 1 binding has been moved to legacy.
 ALERT;OneWire Binding: Some thing types have changed and need to be updated in textual configurations. See documentation for further information.
 ALERT;OpenSprinkler Binding: The stationXX channels have been removed and replaced by a bridge/thing combination. See documentation for further information.
 ALERT;Pushbullet Action: The pushbullet action has been replaced by a new pushbullet binding.


kaikreuzer:
We so far have NOT added such infos to the update.lst file (only for actions that were replaced by a binding). If we say that it isn't enough to list new legacy bindings in the release notes, we should make sure that we add all those to the update.lst. I don't have any preference for the one or the other.

Hilbrand:
I have no preferences. I added it here thinking about what happened when mqtt binding moved to legacy. This surprised a lot of people because at once the binding wasn't directly visible anymore. So I thought to add the alert here so people know why it might not be visible. Although I think this change shouldn't come as a surprise, given the engagement with the community. So if you think it's sufficient to have it in the release notes, feel free to close this pr.

J-N-K:
I would prefer to see this for every binding.

wborn:
If Paper UI didn't show any legacy add-ons by default (only after enabling some filter) we could just keep the legacy add-ons enabled by default. That way the impact of moving add-ons to legacy will be minimized and it would only impact KAR file users (who don't install the legacy add-ons KAR).

Until that time adding such documentation may help. But then it should also list all other such changes because users would then expect all such changes to be listed.

kaikreuzer:
> we could just keep the legacy add-ons enabled by default

I wouldn't change this. But we should add a clear general warning to the release notes that people need to turn on legacy support, if they use such bindings - and we can have a list of the "new" legacy bindings next to it.

wborn:
> I wouldn't change this

What are your reasons for not changing this?

There are also milestone/snapshot users who don't closely follow all community threads who may still run into this issue unless it's addressed by something like automatically generated release notes (https://github.com/openhab/openhab-distro/issues/668).

kaikreuzer:
> What are your reasons for not changing this?

Because it will make all of them appear in the Paper UI (as well as the Karaf console feature list), which might lead to confusion for newbies (and makes the list unnecessarily long).
I consider the legacy support rather to be a feature for long-term users that still have it working. Potentially planning to remove legacy support in future versions, we should encourage people to use 2.x versions and not start with 1.x stuff anymore.
I could imagine that we add a note to the release notes that we recommend anybody who is using 1.x bundles (whether legacy or not) to enable legacy support, then we won't have such problems anymore.

Hilbrand:
Maybe we could add a separate section on OH1 bindings 'replaced' by OH2 bindings in the release notes?

kaikreuzer:
Yes, definitely - we can split the "new bindings" section accordingly.

wborn:
After giving it some thought I think we shouldn't spend to much time on this for OH2 (i.e. just add some documentation) because in OH3 Paper UI might be replaced and support for OH1 add-ons may dropped.

Hilbrand:
I agree to not spent to much time on. But since it can come as a surprise to users when a binding 'just disappears' I think it validates it to have a least some extra attention in release notes. With the oh1 label it should not be to hard to track those changes. I'm going to close my pr here as it doesn't seem needed anymore.

PR #1003, opened 2019-11-15
resolve after openHAB Core runtime dependency changes

maggu2810:
Cross reference: https://github.com/openhab/openhab-core/pull/1218


maggu2810:
Just a side note: As long as the openHAB runtime dependencies are used by the demo app only, it could make sense to move it to the distro repo.

PR #1005, opened 2019-11-17
Add withResolver profile so the runbundles can be resolved from the CLI

wborn:
Add withResolver profile so the runbundles can be resolved from the CLI.


maggu2810:
The resolved bundle PR has already provided by #1003 as the same time I created the other one in openHAC Core.

wborn:
I'll then merge your PR and update this one to add the profile!

PR #1010, opened 2019-11-29
Misc improvments to demoapp

Hilbrand:
- Added transform map to demo pom.xml
- Added runconfiguration options to app.bndrun:
  - Added option that configured to port it runs on, so users can easily change the port.
  - Added bootdelegation option that seems to be required when running a osgiify version of dagger.

### DIFF
--- a/launch/app/app.bndrun
+++ b/launch/app/app.bndrun
@@ -74,7 +74,9 @@ feature.openhab-model-runtime-all: \
 	smarthome.userdata=${.}/runtime/userdata,\
 	osgi.console.enable.builtin=false,\
 	logback.configurationFile=file:${.}/runtime/logback.xml,\
-	log4j.configuration=file:${.}/runtime/log4j.xml
+	log4j.configuration=file:${.}/runtime/log4j.xml,\
+	org.osgi.service.http.port=8080,\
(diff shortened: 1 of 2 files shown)


maggu2810:
>     * Added bootdelegation option that seems to be required when running a osgiify version of dagger.

Another option that allows to solve it by an OSGi bundle instead of startup parameters: https://github.com/diffplug/osgiX

Hilbrand:
@maggu2810 I don't know if that solves this issue. The issue isn't about unresolved packages, but crashes at runtime due to reporting missing classes.

maggu2810:
> Added bootdelegation option that seems to be required when running a osgiify version of dagger.

My reading has been that dagger will use sun.misc on runtime. Perhaps this is not correct.
If dagger needs to access sun.misc the OSGi-ified version should contain an import of that package. And so a bundle that exports it should be enough.

If you add sun.misc to bootdelegation every can use it without declaring a package import.
If it is caused be dagger where the manifest is created by openHAB, the manifest should be fixed.
If this is caused by another OSGi bundle this one should be identified and fixed.

Adding packages to bootdelegation should (IMHO) only be a temporary workaround.

PR #1024, opened 2019-12-22
Discard openhab1-addons and tycho compatibility from releasebuild (3.x)

pfink:
follow-up of #1023

### DIFF
--- a/launch/Jenkinsfile
+++ b/launch/Jenkinsfile
@@ -16,7 +16,6 @@ try {
     
     node {
         deleteDir()
-        installTychoVersionsHotfix()
         releaseOpenHab2Component('openhab-core', false)
     }
     
@@ -26,19 +25,14 @@ try {
     parallel "openhab2-addons": {
         node {
             deleteDir()
-            installTychoVersionsHotfix()
             releaseOpenHab2Component('openhab2-addons', true)
+        }    
+    }, "openhab2-extra-addons": {
+        node {
+            deleteDir()         
             releaseOpenHab2Component('org.openhab.binding.zwave', true)
             releaseOpenHab2Component('org.openhab.binding.zigbee', true)
             releaseOpenHab2Component('openhab-webui', true)
-        }    
-    }, "openhab1-addons": {
-        if(!env.SKIP_OH1_BUILD?.toBoolean()) {
-            node {
-                deleteDir()
-                installTychoVersionsHotfix()
-                releaseOpenHab1Component('openhab1-addons', false)
-            }
         }
     }
     
@@ -46,7 +40,6 @@ try {
     
     node {
         deleteDir()
-        installTychoVersionsHotfix()
         releaseOpenHab2Component('openhab-distro', true)
     }
 }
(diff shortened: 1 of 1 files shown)


launch/Jenkinsfile
```diff
@@ -26,27 +25,21 @@ try {
     parallel "openhab2-addons": {
         node {
             deleteDir()
-            installTychoVersionsHotfix()
             releaseOpenHab2Component('openhab2-addons', true)
+        }    
+    }, "openhab2-extra-addons": {
```
wborn: Are you sure this works? These projects depend on a parent that is build in openhab-addons which is released in parallel by the other node:

https://github.com/openhab/org.openhab.binding.zwave/blob/c6ffb28e6c7e4bf37779bc4b740778fec9eff346/pom.xml#L6

launch/Jenkinsfile
```diff
@@ -60,11 +53,6 @@ if("FAILURE".equals(currentBuild.previousBuild?.result)) {
 }
 
 
-def releaseOpenHab1Component(componentName, updateParent) {
-    def branch = env.OH1_BRANCH ?: 'master'
-    releaseOpenHabComponent(componentName, branch, env.OH1_RELEASE_VERSION, env.OH1_NEXT_VERSION, updateParent, false)
-}
-
 def releaseOpenHab2Component(componentName, updateParent) {
```
wborn: Recently the repo was renamed from openhab2-addons to openhab-addons so the 2 can probably be removed in many places. There will also be no confusion because this PR removes releasing the OH1 from the pipeline.

launch/Jenkinsfile
pfink: Good point! I thought it can be fetched from nexus, but probably it won't work for the first "next version snapshotbuild" test.

launch/Jenkinsfile
pfink: Thanks for the hint. For now, I put this PR on WIP and will try to get it fully green before we go ahead.

launch/Jenkinsfile
wborn: I once had a look if it were possible to release them all in parallel and stumbled upon this. ;-)

Do you know if all Maven plugins used during the release are thread safe? It might speed up releases considerably if you can use a parallel release build (e.g. adding `-T 1C` to mvn commands). 

I do everything with these parallel builds including releases on another project and most of the parallelization issues that I know of for openHAB have been solved except for SAT (https://github.com/openhab/static-code-analysis/pull/365).

pfink: No, I don't know. But I'd vote against a parallel release build. Given the assumption that it doesn't cause technical issues (which is already brave), it's not a good thing for releasebuild. With parallel builds, you gain speed, but you loose traceability and sometimes also reproducibility. For CI snapshotbuilds, parallel builds can be great to reduce the time developers get feedback about their changes. But for releasebuilds, traceability is more important than speed.

wborn: Initially you might run in to some issues with these parallel builds of course. But like everything, the more you use them the better you get with them. Then all issues are fixed and they gain full confidence. After that you can no longer imagine living in a world without them. ;-)

wborn:
There's now also a related PR https://github.com/openhab/openhab-distro/pull/1026 to remove the other OH1 add-ons remainders from the distro .

kaikreuzer:
@pfink Is this still WIP as the title says? Looks good to me already...

pfink:
I started a test run yesterday: https://ci.openhab.org/view/Sandbox/job/sandbox-openhab3-release/1/

It currently fails due to an issue within openhab-distro, but generally I think the script should be fine and can be merged from my point of view.

PR #1026, opened 2019-12-22
Remove OH1 add-ons

wborn:
OH3 no longer has the OH1 compatibility layer so it shouldn't package or support such add-ons.

Related to: https://github.com/openhab/openhab-core/pull/1284


distributions/openhab/src/main/filtered-resources/userdata/etc/version.properties
```diff
@@ -7,6 +7,5 @@ Repository        Version
 ----------------------------------------
 openhab-distro  : ${project.version}
 openhab-core    : ${ohc.version}
-openhab1-addons : ${oh1.version}
-openhab2-addons : ${oh2.version}
+openhab-addons  : ${oh2.version}
```
kaikreuzer: `oh2.version` does not seem to be appropriate anymore... How about changing it to `oha.version`?

wborn: I thought of using that abbreviation too. But I didn't update it because it's also used by the pipeline.

kaikreuzer: Ok, then let's leave that to a follow-up PR.

PR #1035, opened 2019-12-30
Update demo app version to 2.5.1-SNAPSHOT

cdjackson:
@kaikreuzer Is this left at ```2.5.0-SNAPSHOT``` for a reason?

Currently I'm fighting to get the ZigBee binding to run in the IDE. While the IDE version is compiling, when it runs, it seems to be running a ```2.5.0 SNAPSHOT``` version from mid December which then throws exceptions as the source is not inline with the code as they pull in different libraries.  I suspect that the reason it's running the ```2.5.0-SNAPSHOT``` version stems from this, but when I change this to ```2.5.1-SNAPSHOT``` I get dependency errors in the pom - **hence merging this change will break the build**. 

So, the question is - should this be ```2.5.0-SNAPSHOT``` or ```2.5.1-SNAPSHOT```, and the followup is why changing the 2.5.1-SNAPSHOT causes the dependency problems below.

image

### DIFF
--- a/launch/app/pom.xml
+++ b/launch/app/pom.xml
@@ -6,7 +6,7 @@
 
   <groupId>org.openhab.demo</groupId>
   <artifactId>org.openhab.demo.app</artifactId>
-  <version>2.5.0-SNAPSHOT</version>
+  <version>2.5.1-SNAPSHOT</version>
 
   <name>openHAB Demo :: App</name>
 


kaikreuzer:
As @wborn suggested in https://github.com/orgs/openhab/teams/add-ons-maintainers/discussions/6?from_comment=12#discussion-6-comment-12, we should probably change the IDE setup to checkout 2.5.x branches from the repo and have the 2.5.0 release of the core artifacts referenced by the IDE right now. Haven't found the time to look into that - was hoping that someone else jumps on it 🙄.

cdjackson:
Ok, thanks - I’ve no idea how to do that, but at the moment I’ve been struggling for the past 4 days just trying to get the ZigBee Binding to run so have not managed to really get anything that I’d planned done over the holidays :( If there’s any way we can get a working system again it would really be appreciated.

kaikreuzer:
I've created https://github.com/openhab/openhab-distro/pull/1036 to address the issue.

Issue #1031, opened 2019-12-28
openHAB 2.5 initial dashboard stops if openhab2-addons is installed


bwosborne2:
openHAB 2.5 initial dashboard stops if openhab2-addons is installed in a fresh installation 
From my testing, it appears this broke sometime between Milestone 1 and Milestone 2.
I tested this with apt packages and manual installs on a Raspberry Pi 3B+. 

Removing the addons & OH and then reinstalling OH appears to allow OH to run. In my quick testing I was able to upgrade from M1 to M2 to M3 and still have OH work so this appears to be a fresh installation issue.

We have had several users run into this issue recently. The openhab log just shows the dashboard stopping either before or after LSP starts. OH keeps running but Jetty gives a 404 error for /start/index.

kaikreuzer:
I would assume that this is likely to be a duplicate of https://github.com/openhab/openhab-distro/issues/1031. Planning to have a closer look at it soon.

kaikreuzer:
Closing as a duplicate of https://github.com/openhab/openhab-distro/issues/961.

I hope the workaround explained in https://github.com/openhab/openhab-distro/issues/961#issuecomment-569721834 is acceptible for existing 2.5 users.

bwosborne2:
It is only acceptable if the Linux Installation and the Download pages are updated with the workaround. When this was discovered as broken months ago the pages should have at least been updated to warn users of breakage.

kaikreuzer:
The de-installation is not directly related to https://github.com/openhab/openhab-distro/issues/961, so I re-open it as it isn't really an exact duplicate, just closely related.

I have created https://github.com/openhab/openhab-webui/pull/163, which fixes this issue, once 2.5.1 add-ons are released.

kaikreuzer:
Fixed with #1041.

bwosborne2:
Does this mean there may be a mew snapshot or milestone?

kaikreuzer:
Even a release. See https://github.com/openhab/openhab-distro/issues/961#issuecomment-570090499 and https://community.openhab.org/t/development-of-openhab-3-0-0-and-2-5-x/89376.

Issue #961, opened 2019-08-27
Fresh (offline) installation fails


B0ycee:
Attempting to start openhab on a Linux based device without internet access having extracted runtime distribution .zip and placed addons .kar in the addons directory:

Using 2.5.0M1 - succeeds
Using 2.5.0M2 (and also latest snapshot) - fails with following:

org.apache.karaf.features.internal.util.MultiException: Error:
        Error downloading mvn:org.openhab.ui.bundles/org.openhab.ui.dashboard/2.5.0-SNAPSHOT
        at org.apache.karaf.features.internal.download.impl.MavenDownloadManager$MavenDownloader.<init>(MavenDownloadManager.java:91)
...

Perhaps this might be a regression associated with not having internet access available?  Someone else has hit the same error ( https://community.openhab.org/t/openhab-not-starting-error-downloading-mvn-org-openhab-ui-bundles-org-openhab-ui-dashboard-2-5-0-m2/ ) although perhaps the resolution by re-installation is due to using an earlier version or having internet access on re-install?

kaikreuzer:
I think, I have found the culprit: It is this change: https://github.com/openhab/openhab-distro/pull/883

This causes Karaf to install the Dashboard even before the KAR is extracted and its content made available.

If you remove that line from `userdata/etc/org.apache.karaf.features.cfg`, the error goes away. As a result, you won't see the Dashboard in a fresh installation, though.

You can either install it on the openHAB console with
```
feature:install openhab-ui-dashboard
```
or, if you are anyhow using textual configuration, add `dashboard` to the `ui=` section in `conf/services/addons.cfg`:

```
# A comma-separated list of UIs to install (e.g. "ui = basic,paper")
ui = dashboard,paper,basic
```

Unfortunately, a fix cannot be included in the 2.5.1 addons, since the issue is within the distro, for which we do not plan to do a 2.5.1 release.

But I hope the above information is a suitable workaround for anybody wanting to do an offline installation.

Issue will be fixed for openHAB 3.

bwosborne2:
> Unfortunately, a fix cannot be included in the 2.5.1 addons, since the issue is within the distro, for which we do not plan to do a 2.5.1 release.

I thought a 2.5.1 patch release was planned.

openhab-5iver:
2.5.1 (and newer 2.5.x) will unfortunately not include updates to OHC and OHD. They have moved on in development towards OH 3.0 and will stay locked at 2.5.0.

[...]

PR #1041, opened 2020-01-01
Changes required for delivering patch releases for 2.5

kaikreuzer:
- reactivated distro build for 2.5.1
- changed online repo urls
- patched addons feature.xml to use version ranges
- moved definition of dashboard feature from openhab-webui to distro features
- added dashboard to boot feature packaging to make sure that offline installs work
- removed the online-repo as it is no longer required

requires https://github.com/openhab/openhab-webui/pull/164
fixes https://github.com/openhab/openhab-distro/issues/1031
fixes https://github.com/openhab/openhab-distro/issues/961


pom.xml
```diff
@@ -193,7 +193,7 @@
             </activation>
             <properties>
                 <build.number>Milestone Build</build.number>
-                <online.repo>${oh.repo.releaseBaseUrl}/online-repo-milestone/${online.repo.version}</online.repo>
+                <online.repo>https://openhab.jfrog.io/openhab/libs-milestone</online.repo>
```
wborn: What's the rationale for changing the repo URLs? It renders the [properties from the super POM](https://github.com/openhab/infrastructure/blob/f0d31d47a3b5819900336ca30777c3c6f70877f8/pom.xml#L54-L59) to be rather useless.

kaikreuzer: Well, the main reason being that it was already changed for the Release builds with https://github.com/openhab/openhab-distro/pull/1014, so now the snapshot&milestone urls are in line.

But yes, for the master branch, we should best get it back to use properties of the super pom.

PR #1043, opened 2020-01-02
Update ZSS ZigBee libraries to 1.2.9

cdjackson:
This updates to the latest ZSmart Systems ZigBee Libraries ```1.2.9```. It also adds the console bundles into the feature so these are available to users without manual installation.

### DIFF
--- a/features/addons/src/main/feature/feature.xml
+++ b/features/addons/src/main/feature/feature.xml
@@ -6,16 +6,20 @@
     <feature name="openhab-binding-zigbee" description="ZigBee Binding" version="${project.version}">
         <feature>openhab-runtime-base</feature>
         <feature>openhab-transport-serial</feature>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee/1.2.3</bundle>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.dongle.cc2531/1.2.3</bundle>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.dongle.ember/1.2.3</bundle>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.dongle.telegesis/1.2.3</bundle>
-        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.dongle.xbee/1.2.3</bundle>
+        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee/1.2.9</bundle>
+        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.dongle.cc2531/1.2.9</bundle>
+        <bundle start-level="80">mvn:com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.dongle.ember/1.2.9</bundle>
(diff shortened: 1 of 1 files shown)


cdjackson:
@kaikreuzer I recall discussing (with either @wborn or @Hilbrand) the possibility of removing the feature file from this repository and putting the relevant sections into the ```zigbee``` and ```zwave``` repositories so that this is all in one place. At the moment we have two repositories to update and currently they are unsynchronised.

Is this possible at all? It would make the maintenance a lot easier.

Hilbrand:
I can't remember discussing this, so I'm guessing it was with @wborn. But it seems more than logical to move these features to the specific repositories. I don't see why it wouldn't be possible. I'm not familiar with how to make this possible, but would assume something needs to be changed here, and maybe something in the pom's here/in the repos? The openhab-addons might be a reference to see how it works. For zigbee it would be similar to how bluetooth is handled in the openhab-addons repo. Maybe @wborn can help?

cdjackson:
Thanks @Hilbrand - I think it probably was with you while we were working through the bnd/IDE issues 4 or 5 months ago, but I can't find the discussion now.

Either way, it doesn't matter too much - but if we can move the feature file so it's all in one place that would make maintenance a lot easier and more consistent.

Thanks.

kaikreuzer:
We definitely had this discussion and I thought we even created an issue in the zwave/zigbee repos - but it seems that was forgotten.

We definitely should change this, but only for master as we must not change anything about the feature GAVs and content for 2.5.x. So as the master branch of zwave/zigbee are currently not maintained, we will probably have to postpone this change until summer.

@cdjackson Wrt this PR, I will merge it tomorrow after the 2.5.1 build as I want to avoid any risks that might be associated with it (due to its repo-distributed nature).

cdjackson:
Hi Kai,
I don’t remember such an issue in the Zigbee repo, but maybe it was missed somehow.

Please can we merge it before 2.5.1? Currently the binding doesn’t work at all so it can’t get any worse.

Chris

kaikreuzer:
> Currently the binding doesn’t work

??? What do you mean?
I hope you remember that the 2.5.x branch has to be stable all the time and not have any regressions or breaking changes to the 2.5.0 release. There will be a huge shitstorm tomorrow, if the Zigbee binding is completely broken when users receive 2.5.1.

cdjackson:
Well it’s not possible for it to be consistent as we have this problem where the binding has one configuration, and the distro has a different configuration. Since we can’t merge them both at the same time there will always be an inconsistency which is why I wanted to get the feature.xml into the ZigBee Binding repo. In any case, since I assumed that this would have been merged quickly I had merged the changes into the binding already.

If we merge this first, or the other one first, it will break. So, now it’s broken and I’ve been trying to get this merged for the last week or so, so if we can merge it, it would be appreciated.

kaikreuzer:
We have often situations where PRs need to be merged at the same time, that is business as usual.
But we always document this in the PR, so that it is clear to any concerned maintainer. So please leave a clear statement in the PR description in the future. Also, please never merge anything that requires another PR to be merged, if that hasn't been approved or confirmed to be merged at a given moment.

>  I’ve been trying to get this merged for the last week or so

How? I cannot see a single comment here telling that it urgently needs to be merged...

Also, as a maintainer make sure to follow the discussions in the group - just one day before you created this PR, I wrote [here](https://github.com/orgs/openhab/teams/maintainers/discussions/2?from_comment=48#discussion-2-comment-48): "I'd suggest to do a 2.5.1 release on Jan 12 [...] So please consider this date and make sure to only merge stuff into 2.5.1 which is backward compatible and not too risky."

> so it can’t get any worse.

Ok, I hope what you meant is that it will be fully functional and successfully tested in the next snapshot distro - to get this done in time for 2.5.1, I'll merge it right away and also launch a build.

Issue #820, opened 2018-12-08
Replace yahooweather in demo due to EOL


wborn:
Because the Yahoo Weather API is being [retired](https://developer.yahoo.com/weather/) we should update the weather Thing used in the demo to another service. It's likely the API will go EOL after 2.4.0 is released and before 2.5.0 is released.

See also https://github.com/eclipse/smarthome/issues/6491.

The following occurrences need to be updated: 

https://github.com/openhab/openhab-distro/blob/7188264c4130a0b516a7c0dd7292a80f7982d37b/features/distro-resources/src/main/resources/things/demo.things#L1

https://github.com/openhab/openhab-distro/blob/7188264c4130a0b516a7c0dd7292a80f7982d37b/features/distro-resources/src/main/resources/items/demo.items#L102

kaikreuzer:
Unfortunately, there is no other service, which would not require an api token, so there's no easy drop-in replacement...

kaikreuzer:
Any suggestion for a replacement or shall we for now simply remove these items/things completely?

wborn:
We could also simulate the weather in Berlin until we find another good replacement. I.e. similar to what is done for the room temperatures:

https://github.com/openhab/openhab-distro/blob/7188264c4130a0b516a7c0dd7292a80f7982d37b/features/distro-resources/src/main/resources/rules/demo.rules#L91-L99

cweitkamp:
> there is no other service, which would not require an api token

Is it possible to sign up for an API token?

Otherwise OWM provides a [sample URL](https://samples.openweathermap.org/data/2.5/weather?lat=35&lon=139&appid=b6907d289e10d714a6e88b30761fae22) which does neither need a valid API key nor allows to change the location. We can think about implementing a "demo" mode for the binding.

kaikreuzer:
Instead of introducing a demo mode in OWM, we could also simply use the (new) HTTP binding to get hold of the data there...

PR #1050, opened 2020-01-06
[demo] Remove yahooweather items

Hilbrand:
There seems to be the intention to replace it with something else (#820), but until then it doesn't make sense to keep non usable items in the demo app.

Signed-off-by: Hilbrand Bouwkamp <hilbrand@h72.nl>


Hilbrand:
Ok I'll fix that too

PR #1055, opened 2020-01-11
Fix textual configuration (2.5.x)

wborn:
The UI configuration keys changed from org.eclipse.smarthome to org.openhab

It would be nice to also have #1015 merged in 2.5.1.

### DIFF
--- a/features/distro-resources/src/main/resources/services/basicui.cfg
+++ b/features/distro-resources/src/main/resources/services/basicui.cfg
@@ -1,5 +1,5 @@
 # Defining the default sitemap to use
-org.eclipse.smarthome.basicui:defaultSitemap=demo
+defaultSitemap=demo
 
 # The icon type to use, either png or svg
-org.eclipse.smarthome.basicui:iconType=svg
+iconType=svg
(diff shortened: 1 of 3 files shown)


launch/app/runtime/conf/smarthome.cfg
```diff
@@ -5,12 +5,12 @@ org.eclipse.smarthome.folder:scripts=script
 org.eclipse.smarthome.folder:persistence=persist
 org.eclipse.smarthome.folder:things=things
 
-org.eclipse.smarthome.classicui:defaultSitemap=demo
+org.openhab.classicui:defaultSitemap=demo
```
kaikreuzer: The `org.openhab` prefix can be removed from all the entries here.

features/distro-resources/src/main/resources/services/basicui.cfg
```diff
@@ -1,5 +1,5 @@
 # Defining the default sitemap to use
-org.eclipse.smarthome.basicui:defaultSitemap=demo
+defaultSitemap=demo
```
kaikreuzer: Hm, so is this actually a bug in our 2.5.0 release then?

wborn: Yes when opening Basic UI in the demo you'll see the list of sitemaps instead of going directly to the demo sitemap.

launch/app/runtime/conf/smarthome.cfg
kaikreuzer: ... but as it is just a cherry-pick, let's fix this on master only.

PR #1058, opened 2020-01-11
[eclipsesetup] Added option for 2.5.x

Hilbrand:
- Added branch 2.5.x as first installation option to all projects except bacnet and core
- Added label 3.0 to branch selection master to make clear it's for 3.0 development. When 2.5.x is removed this can be changed back.

P.s. when this is merged I'll update the eclipse documentation to match this installation setup.

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2392,8 +2392,10 @@
           locateNestedProjects="true"/>
     </setupTask>
     <stream
+        name="2.5.x"/>
+    <stream
         name="master"
-        label=""/>
+        label="3.0"/>
   </project>
   <project name="2_addons2"
       label="openHAB Add-ons">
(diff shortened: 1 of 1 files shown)


kaikreuzer:
> In Addons only import astro binding. Importing all bindings is useless, so changed to import only one as an example.

We once discussed this and the better solution would be to close projects after import - I just didn't yet find a way how to do it programmatically, so atm the users are asked to do so manually.

I would suggest to leave this change out of this PR as it is anyhow not related to the rest.

Hilbrand:
I can put it in a different pr. But at the moment I keep advising to just not select the option to select openhab-addons because it takes a long time (especially when having a slow connection), gives sometimes errors on certain bindings, people don't read the documentation not enough so having problems with the ide. So to counter this I had changed it here to just import 1 binding.

PR #1059, opened 2020-01-11
Use the latest 1.14.x add-on from Maven

kaikreuzer:
This is a safety measure to potentially allow to ship patches for 1.x add-ons in future.
As 2.5.1 is our last distro, this change will make it possible to nonetheless receive 1.x add-on patches with it, similar to how it already receives 2.5.x add-on patches.


kaikreuzer:
@openhab/distro-maintainers It would be cool if that could be merged tonight, so that it get's into the 2.5.1 distro build.

PR #1061, opened 2020-01-11
[ide] do not import add-ons

Hilbrand:
Importing all bindings is useless, so changed to import only one as an example.

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2422,13 +2422,6 @@
             pattern=".*/openhab-addons/bundles/.*"/>
       </workingSet>
     </setupTask>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Hm, I would actually prefer to not import any project in this case as astro isn't any special to any other add-on, so I would treat them all the same way. If the doc explains how to import add-ons into the workspace, this should be just fine.

kaikreuzer:
Thanks - would be great if you could update the docs accordingly then!

Hilbrand:
Yes I had something prepared. I'll need to update it to add some details about how to import a binding and then will create a pull request.

PR #1065, opened 2020-01-18
Add default UI boot feature

ghys:
Remove Jetty redirect rule to the former dashboard
since the UI will be served at the root path.

Depends on https://github.com/openhab/openhab-webui/pull/178

### DIFF
--- a/distributions/openhab/pom.xml
+++ b/distributions/openhab/pom.xml
@@ -139,6 +139,7 @@
                         <feature>service</feature>
                         <feature>system</feature>
                         <feature>openhab-runtime-base</feature>
+                        <feature>openhab-ui-default</feature>
                     </bootFeatures>
                     <archiveZip>false</archiveZip>
                     <archiveTarGz>false</archiveTarGz>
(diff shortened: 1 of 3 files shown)


distributions/openhab/pom.xml
```diff
@@ -139,6 +139,7 @@
                         <feature>service</feature>
                         <feature>system</feature>
                         <feature>openhab-runtime-base</feature>
+                        <feature>openhab-ui-default</feature>
```
kaikreuzer: I'd like to keep the `openhab-ui-` prefix for real add-ons.
Let's consider the UI to be a central part for runtime and call it `openhab-runtime-ui`, which is a good fit to the `base` feature.
```suggestion
                        <feature>openhab-runtime-ui</feature>
```

features/distro/src/main/feature/feature.xml
```diff
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <features name="${project.artifactId}-${project.version}" xmlns="http://karaf.apache.org/xmlns/features/v1.0.0">
 
+    <feature name="openhab-ui-default" description="openHAB UI" version="${project.version}">
```
kaikreuzer:
```suggestion
    <feature name="openhab-runtime-ui" description="openHAB UI" version="${project.version}">
```

ghys:
On build #1789 the feature is there but wasn't started by default for some reason, I had to "feature:install openhab-runtime-ui" in the console manually. Is something missing?

kaikreuzer:
I noticed the same and didn't yet figure out the reason for it.
Maybe @wborn has an idea, what we are missing?

PR #1056, opened 2020-01-11
Remove Paper UI, Classic UI and HABmin leftovers

wborn:
Related to: https://github.com/openhab/openhab-webui/pull/154

PR #1068, opened 2020-01-19
Some clean up

kaikreuzer:
- removed outdated online-repo (not used anymore)
- removed simple and expert packages as they aren't of much use anymore
- made "standard" the default installed package on first startup


kaikreuzer:
@ghys FYI: With this change, I don't think you will need to include any package selection in the new UI - imho, we should simply have a standard packaging for everyone.

ghys:
Yes was thinking about suggesting something like this - that and to get rid of the "autolinks"/"simple" mode. The setup wizard could maybe offer users to check some add-ons.
I planned to use the installed package to determine the "first-time run" status and show the "welcome" card on the home page, but will now find another way, maybe a flag in the UI's service config to signal the wizard has already run.

J-N-K:
Why would you want to remove simple-mode?

kaikreuzer:
Because the new UI should make it simple and we should streamline the way how openHAB is used - simple mode rather caused a lot of issues for people, does not have an option to be properly deactivated when a setup grows to not-so-simple-anymore and it is quite buggy altogether. Don't tell me that you are using simple mode yourself...?

kaikreuzer:
I've rebased the changes after https://github.com/openhab/openhab-distro/pull/1056 was merged.

ghys:
> Why would you want to remove simple-mode?

> Because the new UI should make it simple and we should streamline the way how openHAB is used - simple mode rather caused a lot of issues for people, does not have an option to be properly deactivated when a setup grows to not-so-simple-anymore and it is quite buggy altogether. Don't tell me that you are using simple mode yourself...?

Yes, the simple mode often gives less-than-pretty names to items, doesn't care about putting them in a proper group hierarchy, and overall doesn't give the user much control; there would be features in the UI to link channels in bulk and create items - but after choosing which channels are relevant, and haven given the user the opportunity to customize the items before creating them.
It can also delete links and items in bulk if necessary.

When simple mode was activated by accident (or by a package and you weren't aware of it) you would end up with a mess of unwanted items after adding a new thing, which were difficult to clean up.

Perhaps the feature could stay, but should IMO have to be completely opt-in and renamed to stop suggesting it makes things "simpler".

kaikreuzer:
> Perhaps the feature could stay

I'd definitely prefer to completely remove it.

distributions/openhab/src/main/resources/conf/services/addons.cfg
```diff
@@ -3,18 +3,13 @@
 # Note: If you want to specify your add-ons yourself through entries below, set the package to "minimal"
 # as otherwise your definition might be in conflict with what the installation package defines.
 #
-# Optional. If not set, the dashboard (https://<yourserver>:8080/) will ask you to choose a package.
-#
 # Valid options:
 #   - minimal  : Installation only with dashboard, but no UIs or other add-ons. Use this for custom setups.
```
ghys: Leftover reference to the dashboard here.
```suggestion
#   - minimal  : Installation only with the UI but no other add-ons. Use this for custom setups.
```

distributions/openhab/src/main/resources/conf/services/addons.cfg
```diff
@@ -3,18 +3,13 @@
...
-#   - simple   : Setup for using openHAB purely through UIs - you need to expect MANY constraints in functionality!
 #   - standard : Default setup for normal users, best for textual setup
-#   - expert   : Setup for expert users, especially for people migrating from openHAB 1.x
-#   - demo     : A demo setup which includes UIs, a few bindings, config files etc.
 #
 # See https://www.openhab.org/docs/configuration/packages.html for a detailed explanation of these packages.
 #
-#package = minimal
+package = standard
```
ghys: What would happen to the demo setup? It's sometimes useful.
The setup wizard could have a step asking users if they want "a few things configured for them" (that would install the standard package), a demo setup (install the demo package), or "skip" the step entirely (stay on minimal), wdyt?

distributions/openhab/src/main/resources/conf/services/addons.cfg
kaikreuzer: Not even with the UI - the idea for the minimal setup was to use it e.g. for a headless system. Will fix the formulation.

distributions/openhab/src/main/resources/conf/services/addons.cfg
kaikreuzer: As written above: I would suggest to completely remove the package selection from the wizard. I for now kept the package mechanism still in place, so that "minimal" and "demo" can be picked for very special setups (e.g. our demo server). But the package is installed upon the very first startup already, so once the UI is there, there is no need to select anything anymore - it simply means the user has chosen a "standard" setup. Does that make sense to you?

ghys: Yes it's clearer, I thought the minimal package included the UI (doesn't it though, because it's a boot feature rather than an add-on?)

kaikreuzer: You're right, that's likely something we will have to change - adding the dashboard as a boot feature was only meant as a quick hack to solve the offline installation issues.

J-N-K:
I'm using simple mode for development. It's really nice to see all values in the UI immediately after channels are added.

kaikreuzer:
We should be able to offer something suitable in the "dev tools" section of the new UI as a replacement.

Issue #1060, opened 2020-01-11
Move ZigBee binding feature file to ZigBee repo


cdjackson:
As discussed [here](https://github.com/openhab/openhab-distro/pull/1043#issuecomment-573330300) it would be good to move the zigbee feature file to the zigbee binding so that everything remains consistent with the libraries defined in one place.

PR #1076, opened 2020-02-07
Use feature file from ZigBee repository

cdjackson:
This removes the feature file definition for ZigBee and uses a file from the ZigBee repository.
https://github.com/openhab/openhab-distro/issues/1060

**Must not be merged before https://github.com/openhab/org.openhab.binding.zigbee/pull/556**, and the build will presumably fail until this is merged.

### DIFF
--- a/distributions/openhab-addons/pom.xml
+++ b/distributions/openhab-addons/pom.xml
@@ -42,6 +42,13 @@
             <classifier>features</classifier>
             <type>xml</type>
         </dependency>
+        <dependency>
+            <groupId>org.openhab.addons.features.karaf</groupId>
+            <artifactId>org.openhab.addons.features.karaf.zigbee</artifactId>
+            <version>${oh2.version}</version>
+            <classifier>features</classifier>
+            <type>xml</type>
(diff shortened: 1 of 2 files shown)


kaikreuzer:
As mentioned in https://github.com/openhab/openhab-distro/pull/1043#issuecomment-573348219, these PRs should go against master (and this it probably only makes sense to do it later this year, when we have switched add-on development to master).

cdjackson:
Ok, I guess I will close this for now. It would be good if you could at least look at https://github.com/openhab/openhab-distro/issues/1060.

Thanks.

kaikreuzer:
>  It would be good if you could at least look at #1060.

I did, but there isn't much to see.
Your PRs should be just fine to solve #1060, but only for master and better later.

cdjackson:
Ok, there was a question about how to get other features merged into this since at the moment it only includes the dependencies from the pom. I also wanted to get it to add -:

```
        <feature>openhab-runtime-base</feature>
        <feature>openhab-transport-serial</feature>
```

PR #1078, opened 2020-02-22
Use XML instead of properties file for log4j2 configuration

wborn:
Note that Karaf processes the log4j2 configuration to add the sshd logger and this processing of XML elements is case sensitive.
It will throw exceptions if lower case element names are used or when the root logger is at the end of the list of loggers.

When new loggers are added, the XML configuration is serialized such that:
* The `<Configuration>` element is added to the same line as the `<?xml>` element
* Attributes are sorted so therefor the 'name' is at the end of the appender/logger lines

Related to #516

---

Including other configuration files still results in issues (see: https://github.com/openhab/openhab-distro/issues/516#issuecomment-350902692).


kaikreuzer:
I guess this will also require some documentation updates.
Could you add it to the backlog for the docs?

wborn:
I've created https://github.com/openhab/openhab-docs/issues/1163 for this.

Do you think we should also replace Logback with Log4j2 for OH3 development/tests? It might be easier to configure logging when users/contributors are already familiar with the log4j2 XML syntax when it is also used by the OH3 distro. Since log4j2 is more feature rich (complex) it may impact build times. Log4j2 probably also has more/bigger dependencies/artifacts. I do not know the exact reasons why Logback is currently used during development.

kaikreuzer:
Logback is currently used because openHAB used to use logback from the beginning.
Yes, in general it makes a lot of sense to have the same logging in the IDE as in the distro.
I wouldn't really expect negative impact on build times by using a modern logging framework...

PR #1086, opened 2020-03-08
Suppress Nashorn deprecation warnings

wborn:
Suppresses the Nashorn deprecation warnings that show on startup on newer Java versions:

> Warning: Nashorn engine is planned to be removed from a future JDK release

See also: https://bugs.openjdk.java.net/browse/JDK-8210140


openhab-5iver:
Thank you!

PR #1091, opened 2020-03-21
Update demo app version

wborn:
As the pipeline doesn't do this we need to manually update the app version with every release.
See also #1077


ssalonen:
Is version bump needed again?

wborn:
It's always best to keep it in sync. Can you create a PR to update the versions?

PR #1094, opened 2020-03-25
added openhab3-specific add-ons and set rrd4j as default persistence service

kaikreuzer:

### DIFF
--- a/distributions/openhab-addons/pom.xml
+++ b/distributions/openhab-addons/pom.xml
@@ -29,6 +29,13 @@
             <type>xml</type>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


kaikreuzer:
@ghys & @wborn: I think we have just reached a major milestone - with this change, the distro works with the new admin authentication, includes persistence services by default and is thus now in a state that I'll soon consider switching to it for my production system :-)

ghys:
Excellent 😎 I'll take this remark as the go-ahead to start working on the historical data visulization ;)

> now in a state that I'll soon consider switching to it for my production system

same, but I (and probably many others) would still be missing a working [http binding](https://github.com/openhab/openhab-addons/pull/4342) or something similar :)

PR #1108, opened 2020-05-04
Migrate to JAX-RS Whiteboard

### LINKED ISSUE #945: add REST doc to demo app
The (swagger based) REST documentation should be added to the demo application.

### LINKED ISSUE #1045: Demo app 2.5.x. misses swagger.json and some rest apis
The demo app for the 2.5.x branch misses `rest/swagger.json`. It also misses the `/rest/extensions` (Visible in PaperUI as the menu entry Add-ons is missing) and seems to miss some other rest apis as well. I couldn't directly see why they were missing or is this intentional? I will continue looking into it, but if it's intentional of if someone has an idea that might cause this, that would be helpful.

wborn:
When `de.maggu2810.thirdparty.modified.org.reflections` is used as runtime dependency, the Swagger bundles resolve properly because it has an optional dependency on `com.google.common.base` .
So now `/rest/swagger.json` is also generated and the REST Docs UI will properly work. :slightly_smiling_face: 

Fixes #945
Fixes #1045

Depends on https://github.com/openhab/openhab-core/pull/1443

### DIFF
--- a/launch/app/app.bndrun
+++ b/launch/app/app.bndrun
@@ -59,7 +59,7 @@ feature.openhab-model-runtime-all: \
 	bnd.identity;id='org.openhab.core.automation.rest',\
 	bnd.identity;id='org.openhab.core.io.console.rfc147',\
 	bnd.identity;id='org.openhab.core.io.rest.sitemap',\
-	bnd.identity;id='org.openhab.core.io.rest.swagger1',\
+	bnd.identity;id='org.openhab.core.io.rest.swagger',\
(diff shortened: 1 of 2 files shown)


wborn:
The code is ready for review and it can be merged as soon as Jersey is removed as dependency from the remaining add-ons.

wborn:
See https://github.com/openhab/org.openhab.binding.zigbee/issues/578 regarding the build failures.

kaikreuzer:
Yeah, I was just after that :-)
The rest went well, so it does not stop me from merging!

PR #1112, opened 2020-05-17
Use JSON Storage in Demo App

wborn:
It simplifies debugging because it is easier to check what is stored by checking the .json files.
Because we also use it in the distro it allows for easier reproducing issues or reusing existing .json storage files.
If there are breaking changes in the stored data these will also be easier to patch manually.
It is probably also not easy to patch OH2 mapdb storage files when contributors start developing with OH3.


wborn:
I think it will mostly be a burden to keep maintaining mapdb.

It may help a little to see if an issue is specific to the JSON storage. It sometimes also helps with getting a good interface definition when there are multiple implementations. Though we'll also have to update the implementation which takes time and the Storage interface isn't frequently updated.

PR #1117, opened 2020-06-22
Remove Java 11 nashorn.args

wborn:
They cause issues when used with Java 8.

Fixes https://github.com/openhab/openhab-addons/issues/7880

### DIFF
--- a/distributions/openhab/src/main/resources/bin/setenv
+++ b/distributions/openhab/src/main/resources/bin/setenv
@@ -100,7 +100,6 @@ export JAVA_OPTS="${JAVA_OPTS}
   -Djava.library.path=${OPENHAB_USERDATA}/tmp/lib
   -Djetty.host=${HTTP_ADDRESS}
   -Djetty.http.compliance=RFC2616
-  -Dnashorn.args=--no-deprecation-warning
   -Dorg.ops4j.pax.web.listening.addresses=${HTTP_ADDRESS}
   -Dorg.osgi.service.http.port=${HTTP_PORT}
   -Dorg.osgi.service.http.port.secure=${HTTPS_PORT}"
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Interesting. How can this setting have such an impact...?

wborn:
Yes I also thought it would be pretty harmless when used with Java 8.
But it seems to panic when it doesn't know what to do with it. :disappointed:

helmar74:
Hi,

how could I solve the issue in my Openhab 2.5.6 environment? (Updated today)

Thanks a lot

wborn:
Remove the line from the `runtime/bin/setenv(.bat)` script as you can see in the [changes](https://github.com/openhab/openhab-distro/pull/1117/files).

helmar74:
ok, do I have to stop OH before and restart after the change?

wborn:
It doesn't really matter because the file is only read at startup. Just make the change and restart openHAB.

kaikreuzer:
Tested latest distro and I can confirm that it solves the issue, thanks!
It'll probably make sense to just do a new 2.5.6 distro binary with that as none of the add-ons have changed.

BClark09:
Hi @kaikreuzer, have you started the new build? I'll make sure to build 2.5.6-2 for the APT and YUM repos and then manually delete 2.5.6-1, this way people who have already updated can update again.

kaikreuzer:
Yes, just uploading the distro to Bintray (https://ci.openhab.org/view/Release%20Jobs/job/openhab-release/85), should be finishing soon.
If you can deal with the linux package update to 2.5.6-2, it would be great, thanks!

kaikreuzer:
Distro is on Bintray. @BClark09, you can go ahead!

BClark09:
Thanks again Kai, done and available on APT/YUM repos now :)

We can delete 2.5.6-1 at a later time.

helmar74:
Updated to 2.5.6-2, issue is fixed.

Thanks a lot

DerOetzi:
Was the docker image 2.5.6 rebuilt after this fix, because I cannot find this line inside my test instance?

wborn:
Yes it was. That means you're using an up to date image. :-)

PR #1126, opened 2020-07-24
Remove the restdocs UI and jaxrswb-swagger1-gen from the demo app

ghys:
Related to https://github.com/openhab/openhab-core/pull/1496 & https://github.com/openhab/openhab-webui/pull/297

### DIFF
--- a/launch/app/app.bndrun
+++ b/launch/app/app.bndrun
@@ -64,10 +64,8 @@ feature.openhab-model-runtime-all: \
 	bnd.identity;id='org.openhab.ui',\
 	bnd.identity;id='org.openhab.ui.basic',\
 	bnd.identity;id='org.openhab.ui.iconset.classic',\
-	bnd.identity;id='org.openhab.ui.restdocs',\
 	bnd.identity;id='org.apache.aries.jax.rs.whiteboard',\
-	bnd.identity;id='org.ops4j.pax.web.pax-web-extender-whiteboard',\
-	bnd.identity;id='jaxrswb-swagger1-gen'
+	bnd.identity;id='org.ops4j.pax.web.pax-web-extender-whiteboard'
 
 -runfw: org.eclipse.osgi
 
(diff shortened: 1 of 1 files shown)


launch/app/app.bndrun
```diff
@@ -259,7 +254,9 @@ feature.openhab-model-runtime-all: \
 	org.osgi.util.promise;version='[1.1.0,1.1.1)',\
 	org.threeten.extra;version='[1.4.0,1.4.1)',\
 	org.yaml.snakeyaml;version='[1.24.0,1.24.1)',\
-	reflections;version='[0.9.10,0.9.11)',\
 	slf4j.api;version='[1.7.25,1.7.26)',\
 	tec.uom.lib.uom-lib-common;version='[1.0.3,1.0.4)',\
-	tec.uom.se;version='[1.0.10,1.0.11)'
+	tec.uom.se;version='[1.0.10,1.0.11)',\
+	jakarta.xml.bind-api;version='[2.3.2,2.3.3)',\
+	javassist;version='[3.26.0,3.26.1)',\
```
kaikreuzer, line 261: These two lines seem unrelated to the PR, right?

ghys: They are now added when resolving because of this (from https://github.com/openhab/openhab-core/pull/1496):

> There was also a problem with a conflict between Guava versions (as discussed [here](https://github.com/openhab/openhab-distro/issues/1045#issuecomment-596255184)), solved by fixing the org.reflections:reflections dependency to version 0.9.12 (swagger-jaxrs depends on version 0.9.11) because this recent release removes the dependency to an outdated release of Guava. This seems to do the trick without side effects and apparently solves resolutions problems in both the demo app and Karaf. This allows in turn to remove the personal fork of this lib (de.maggu2810.thirdparty.modified.org.reflections/reflections/0.9.10.v20160429-1435), in favor of an official version.

kaikreuzer: I don't see how xml.bind and javaassist is related to that discussion, but if you say that this is fine and required, that's ok. I just remembered that we saw xml.bind being added on the addons itests as well and considered this to be an incorrect dependency.

ghys: FYI:

image

image

PR #1131, opened 2020-08-10
fixed download location of openHAB3 snapshots

kaikreuzer:

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update
+++ b/distributions/openhab/src/main/resources/bin/update
@@ -10,9 +10,9 @@ setup(){
     echo "Usage: ./runtime/bin/update [VersionNumber] [openHAB Dir]"
     echo ""
     echo "  e.g. ./runtime/bin/update                << Updates to the next version"
-    echo "       ./runtime/bin/update 2.0.0          << Updates to a specific version"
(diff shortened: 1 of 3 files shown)


distributions/openhab/src/main/resources/bin/update
```diff
@@ -62,7 +62,7 @@ setup(){
   fi
 
   ## Check to see if processes are running before updating
-  if [ ! -z "$(ps aux  | grep "openhab2.*java" | grep -v grep)" ]; then
+  if [ ! -z "$(ps aux  | grep openhab | grep -v grep)" ]; then
```
kaikreuzer, line 65: @wborn I need your expertise, please. This change didn't have seemed to work out - the update script now always claims that openHAB is running, even if it isn't. I done really understand what was the idea behind `"openhab2.*java"` - can you help to fix this?

wborn: There is probably another process running that has "openhab" in its path or arguments. If I run Eclipse it also matches "openhab". So "openhab2.*java" is to prevent false positives and only match the real openHAB java process.

kaikreuzer: Checking it, I see that you are right - there is also a shell process for `runtime/bin/karaf`, which happens to run inside an `openhab` folder...

PR #1134, opened 2020-08-13
fixed openhab process detection

kaikreuzer:
See https://github.com/openhab/openhab-distro/pull/1131/files#r470011989

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update
+++ b/distributions/openhab/src/main/resources/bin/update
@@ -62,7 +62,7 @@ setup(){
   fi
 
   ## Check to see if processes are running before updating
-  if [ ! -z "$(ps aux  | grep openhab | grep -v grep)" ]; then
+  if [ ! -z "$(ps aux  | grep "openhab.*java" | grep -v grep)" ]; then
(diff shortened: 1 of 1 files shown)


distributions/openhab/src/main/resources/bin/update
```diff
@@ -62,7 +62,7 @@ setup(){
   fi
 
   ## Check to see if processes are running before updating
-  if [ ! -z "$(ps aux  | grep openhab | grep -v grep)" ]; then
+  if [ ! -z "$(ps aux  | grep "openhab.*java" | grep -v grep)" ]; then
```
BClark09, line 65:
```sh
if [ ! -z "$(pgrep -f "openhab.*java")" ]; then
```

Would be cleaner, and [maybe safer](https://github.com/openhab/openhab-linuxpkg/pull/146) to use if this works correctly on MacOS?

kaikreuzer: Just tested, `pgrep` seems to work on macOS as well, yes. Wanna do a PR?

PR #1143, opened 2020-09-13
Apply spotless formatter after changes

kaikreuzer:
Related to https://github.com/openhab/openhab-addons/pull/8401

### DIFF
--- a/launch/Jenkinsfile
+++ b/launch/Jenkinsfile
@@ -94,7 +94,9 @@ def releaseOpenHabComponent(componentName, branch, releaseVersion, nextVersion, 
                     }
 
                     sh "mvn unleash:perform -Dworkflow=unleash.phase1.workflow " + mvnOptions
-                    
+
+                    sh "mvn spotless:apply"
+
                     //Build and publish artifacts
                     sh "mvn deploy " + mvnReleaseOptions
                                         
@@ -118,6 +120,8 @@ def releaseOpenHabComponent(componentName, branch, releaseVersion, nextVersion, 
                         }
 
                         sh "mvn unleash:perform -Dworkflow=unleash.phase3.workflow " + mvnOptions
+
+                        sh "mvn spotless:apply"
 
                         sh "mvn deploy " + mvnSnapshotOptions
                     }


wborn (approved):
Let's test if it fixes the sandbox build.

kaikreuzer:
Looks much better, but for some reason the webui build failed when running spotless, see https://ci.openhab.org/view/Sandbox/job/sandbox-openhab2-release/1640/console.

Locally, I see no issue with that. Do you have any idea what might be wrong there?

wborn:
Yes it's the 2.5.x branch for which spotless was never activated.
It only got added on the master branch with https://github.com/openhab/openhab-webui/pull/240.

kaikreuzer:
Good point, I accidentally checked the master branch only.
I've created https://github.com/openhab/openhab-webui/pull/335, which should hopefully address the issue.

kaikreuzer:
Hooray, it is green again: https://ci.openhab.org/job/sandbox-openhab2-release/1647/

wborn:
That's great so we can deliver 2.5.9 on time and proceed with porting the add-ons!

Will this also commit the applied Spotless changes to the repo?

When the previous successful pre-Spotless build was still there, it looked like running the Spotless check goal in the new builds had a significant impact on the build time of all the "Prepare next version" stages.
That could be reduced by adding `-Dspotless.check.skip=true` to some of the Maven [commands](https://github.com/openhab/openhab-distro/blob/44222a04bd236ce16f509957a7b0e11160ddb771/launch/Jenkinsfile#L64).

wborn:
If I compare the stage views between [openhab-release](https://ci.openhab.org/view/Release%20Jobs/job/openhab-release/) and [sandbox-openhab2-release](https://ci.openhab.org/view/Sandbox/job/sandbox-openhab2-release/) the increase seems less significant. It may also have been some Jenkins glitch.

kaikreuzer:
> Will this also commit the applied Spotless changes to the repo?

I'd think so, yes. We will see with the 2.5.9 release :-)

wborn:
The Jenkinsfile also needs to be updated on the master branch to be able to release OH3. Since it did not commit the changes to the repo, adding `-Dspotless.check.skip=true` would yield the same result but faster. ;-)

kaikreuzer:
I've created https://github.com/openhab/openhab-distro/pull/1154 - let's see how well that works.

PR #1146, opened 2020-09-18
Replace Joda-Time with java.time in demo.rules

wborn:
Signed-off-by: Wouter Born <github@maindrain.net>


kaikreuzer (approved):
Good catch, thanks!

PR #1152, opened 2020-09-21
[setup] Renamed master to main for addons setup

Hilbrand:
And rename of addons master to main.

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2425,7 +2425,7 @@
     <stream
         name="2.5.x"/>
     <stream
-        name="master"
+        name="main"
         label="3.0"/>
   </project>
   <project name="4_zigbee"


mhilbush:
Thanks!!

hmerk:
@Hilbrand What about someone wants to enhance a 2.5.x binding or fix bugs ? 2.5.x development would still be needed then, am I wright ?

kaikreuzer:
I'd agree with @hmerk - I would leave the option to set up an IDE for 2.5 as well.

kaikreuzer:
@Hilbrand Would you agree to keep the 2.5.x parts?

Hilbrand:
Yes i had not time to update. I was looking at changing the default to java 11, which should work for 2.5.x. I have some local changes and am working on updating  the documentation.

Hilbrand:
Removed the removal of 2.5.x I'll look into the default to  java 11 in the setup later. Because I need to check this won't break 2.5.x development because that should still work with java 8.

kaikreuzer (approved):
Thanks!

PR #1158, opened 2020-10-03
Reenable ZigBee binding

wborn:
Also updates the versions and adds the new serial and console bundles.

CC: @cdjackson

### DIFF
--- a/distributions/openhab-addons/pom.xml
+++ b/distributions/openhab-addons/pom.xml
@@ -29,6 +29,13 @@
             <type>xml</type>
         </dependency>
         <dependency>
+            <groupId>org.openhab.addons.features.karaf</groupId>
+            <artifactId>openhab-binding-zigbee</artifactId>
+            <version>${project.version}</version>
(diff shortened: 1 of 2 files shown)


cdjackson:
Thanks @wborn, however, can we instead start using the feature generated by the binding itself? I added this a few months back with a view that we can avoid having this feature file here - all dependencies would then be controlled in one place :)

wborn:
I had a look at that and with a few changes it seems to work in general. However currently the generated feature name prevents it from showing up in the UI. The feature name should be `openhab-binding-zigbee` while it currently seems to be derived from the artifactId `org.openhab.addons.features.karaf.zigbee` .

wborn:
OK all is green now. Now we need to find one of the @openhab/distro-maintainers who wants to push the merge button.

kaikreuzer (approved):
Thanks!
So shall we also move the ZWave feature to its repo as a next step, @cdjackson?

PR #1165, opened 2020-10-05
set default logging to debug for bindings

marcelrv:
As in the demo environment we typically use for testing, bindings can be more verbose

Signed-off-by: Marcel Verpaalen <marcel@verpaalen.com>

### DIFF
--- a/launch/app/runtime/logback.xml
+++ b/launch/app/runtime/logback.xml
@@ -39,5 +39,7 @@
   </root>
 
   <logger name="org.openhab.core" level="INFO" />
+  <logger name="org.openhab.core" level="DEBUG" />
+
 
 </configuration>


launch/app/runtime/logback.xml
```diff
@@ -39,5 +39,7 @@
   </root>
 
   <logger name="org.openhab.core" level="INFO" />
+  <logger name="org.openhab.core" level="DEBUG" />
+
```
wborn, line 43: I would personally prefer it to match the logging configuration of the distro as much as possible (see [log4j2.xml](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/userdata/etc/log4j2.xml)). That way abnormal exceptions, errors, warnings stand out more are are more likely fixed. It will also be easier to find spammy loggers using the wrong log level (info+).

```suggestion
  <logger name="org.openhab" level="INFO" />
```

wborn (changes requested):

launch/app/runtime/logback.xml
marcelrv: If we don't want to have it as DEBUG, which I prefer as developer, at least let's make sure we keep the lines for it so it can be fast tweaked

e.g. we keep it to 
```
<logger name="org.openhab.core" level="INFO" />
<logger name="org.openhab.core" level="INFO" />
<logger name="org.openhab" level="INFO" />
```

wborn: I'm also a developer but only want to see selective debug logging when really needed. :wink: 
Adding some place holder(s) which can be easily uncommented would be helpful.

wborn:
Succeeded by https://github.com/openhab/openhab-distro/pull/1273

PR #1170, opened 2020-10-18
Update default addons.cfg

wborn:
* Add automation type
* Sort content
* Replace freetts

Related to openhab/openhab-core#1722

### DIFF
--- a/distributions/openhab/src/main/resources/conf/services/addons.cfg
+++ b/distributions/openhab/src/main/resources/conf/services/addons.cfg
@@ -17,20 +17,23 @@ package = standard
 #
 #remote = true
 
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):

lolodomo:
@wborn : I believe that the example with `misc = myopenhab` is wrong as we have to use `openhabcloud` rather than `myopenhab`.

```
openhab> bundle:list -s | grep loud
225 x Active x  80 x 3.0.0.202010220354      x org.openhab.io.openhabcloud
```

kaikreuzer:
@lolodomo Good catch, you are clearly right! Want to push a PR?

lolodomo:
Done quickly through an edit in github

PR #1173, opened 2020-10-19
Remove NoOfLights from demo rrd4j.persist

wborn:
There is no such item.

I found this after I was able to enable debug logging of the PersistenceManagerImpl.


kaikreuzer (approved):
If you say so 😎 .

PR #1183, opened 2020-11-12
Disable periodic logging of performance data by the JVM

### LINKED ISSUE #847: openHAB causing a lot of disc IOs for Java PerfLogs
On my machine (Ubuntu 18.04, Openhab testing, Open JDK 1.8) openhab is permanently writing PerfLogs to /tmp. This keeps the disc from entering standby mode, even if nothing is happening.

Could we add the following option to the default Java parameters to disable PerfLogs for the simple user?
`-XX:-UsePerfData`

fwolter:
Many JVMs log performance data to /tmp/hsperfdata_openhab/[pid] every 30s. This wears SD cards and prevents HDDs from spinning down. Closes #847

I came to this because the default JDK of Raspbian on my raspi has above behavior.

Signed-off-by: Fabian Wolter <github@fabian-wolter.de>

### DIFF
--- a/distributions/openhab/src/main/resources/bin/setenv
+++ b/distributions/openhab/src/main/resources/bin/setenv
@@ -117,7 +117,7 @@ export JAVA_OPTS="${JAVA_OPTS}
 # set JVM options
 #
 ARCH=`uname -m`
-EXTRA_JAVA_OPTS_COMMON="-Djava.awt.headless=true"
+EXTRA_JAVA_OPTS_COMMON="-Djava.awt.headless=true -XX:-UsePerfData"
 EXTRA_JAVA_OPTS_ARCH=""
 case "$ARCH" in
     *arm*) ;;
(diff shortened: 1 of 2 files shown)


lolodomo:
I just checked on 2 RPI with a Zulu8 JVM (recommended JVM for OH 2.x) and a Zulu11 JVM and you're absolutely correct, I found this file updated frequently.

kaikreuzer (approved):
Interesting, I wasn't aware of that feature (and that it is enabled by default). Thanks!

kaikreuzer:
Ah, I just learned myself that I actually WAS aware of it (https://github.com/openhab/openhab-distro/issues/847#issuecomment-489445729) 😆

wborn:
With this change you can no longer troubleshoot issues of a running instance with tools like VisualVM out of the box.

kaikreuzer:
That's bad - might make sense to keep it in place when started in debug mode then.
@fwolter Wanna do a follow-up PR?

fwolter:
Yep.

PR #1193, opened 2020-11-29
Keep logging of performance data enabled when debugging

fwolter:
Signed-off-by: Fabian Wolter <github@fabian-wolter.de>

### DIFF
--- a/distributions/openhab/src/main/resources/bin/karaf
+++ b/distributions/openhab/src/main/resources/bin/karaf
@@ -268,9 +268,12 @@ run() {
             fi
         fi
         JAVA_OPTS="${JAVA_DEBUG_OPTS} ${JAVA_OPTS}"
+
+### START openHAB customization
     else
         JAVA_OPTS="${JAVA_NON_DEBUG_OPTS} ${JAVA_OPTS}"
     fi
+### END openHAB customization
 
     while true; do
         # When users want to update the lib version of, they just need to create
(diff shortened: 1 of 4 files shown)


kaikreuzer (commented):
Thanks.
Would you find a way to do it without touching the `karaf` file? Afaik, we do not customize it so far, which means that such changes might get easily lost with the next Karaf upgrade.

I see that Karaf does not have any variable for specific non-debug parameters, so it might not be easy to solve, though.

fwolter:
The karaf script executes the setenv script before the debug parameters are processed. `start.sh` and `start_debug.sh` are not used when using systemd. I slept over it, but no solution came to my mind without touching the karaf file.

kaikreuzer:
@wborn As you are our "Karaf upgrade master": Would you have any good recommendation on how to proceed here?

wborn:
> Would you have any good recommendation on how to proceed here?

It is easy for such changes to get lost when upgrading Karaf the next time. So to prevent that from happening it's probably best to add some comments to those lines to clearly indicate that those are openHAB customizations.

fwolter:
Done

kaikreuzer (approved):
Thank you!
And now you could try to do a PR against https://github.com/apache/karaf, maybe then we won't need to merge our customizations when upgrading anymore 😎 .

fwolter:
Done apache/karaf#1261

fwolter:
The karaf PR has been merged.

PR #1196, opened 2020-12-02
an attempt at migrating Karaf logging to OH3

mstormi:
This is a best effort listing of required changes from an experienced OH user (no core dev).
Maybe not perfect but a lot better than to reset to defaults (= empty) which will have users search for module names that have changed.

Requires: https://github.com/openhab/openhab-linuxpkg/pull/186

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -38,6 +38,9 @@ ALERT;TP-Link Smart Home Binding: The energy and rssi channels are now using Uni
 ALERT;Tesla Binding: The binding has been heavily refactored. It is required to delete old Tesla Things and set them up from scratch.
 ALERT;Vitotronic Binding: The following channels have been renamed: 'outsite_temp' to 'outside_temp', 'pelletburner:power' to 'pelletburner:powerlevel', 'party_temp' to 'party_temp_setpoint' and 'save_temp' to 'save_temp_setpoint'
 
+[3.0.0]
+ALERT;We tried to migrate Karaf logging configuration in 'org.ops4j.pax.logging.cfg' to openHAB3 format but there may have happened mistakes in doing so. Please cross-check.
+
 [[PRE]]
 [2.2.0]
 DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg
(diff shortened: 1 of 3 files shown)


distributions/openhab/src/main/resources/bin/update.lst
```diff
@@ -53,6 +55,21 @@ MOVE;$OPENHAB_USERDATA/config/org/eclipse/smarthome/core/i18nprovider.config;$OP
 REPLACE;org.eclipse.smarthome.core.i18nprovider;org.eclipse.smarthome.i18n;$OPENHAB_USERDATA/config/org/eclipse/smarthome/i18n.config
 
 [3.0.0]
+REPLACE;org_eclipse_smarthome_model;org_openhab_core_model;$OPENHAB_USERDATA/org.ops4j.pax.logging.cfg
+REPLACE;org.eclipse.smarthome.model;org_openhab.core.model;$OPENHAB_USERDATA/org.ops4j.pax.logging.cfg
+REPLACE;org_eclipse_smarthome;org_openhab;$OPENHAB_USERDATA/org.ops4j.pax.logging.cfg
+REPLACE;org.eclipse.smarthome;org.openhab;$OPENHAB_USERDATA/org.ops4j.pax.logging.cfg
+REMOVE;logger.paperui;$OPENHAB_USERDATA/org.ops4j.pax.logging.cfg
```
BClark09, line 62: New commands need to be interpreted by: 

https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/update

for Linux and macOS

and

https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/update.ps1

for Windows.

mstormi: Ok I gave it a try. But I'm a total noob on PowerShell. Can you validate and eventually fix the PS command I inserted please ?

wborn (changes requested):
Since https://github.com/openhab/openhab-distro/pull/1078 OH3 uses an XML configuration for logging. The only content in [org.ops4j.pax.logging.cfg](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/userdata/etc/org.ops4j.pax.logging.cfg) should be:

`org.ops4j.pax.logging.log4j2.config.file=${karaf.etc}/log4j2.xml`

So there is no logger config to migrate and if this file isn't properly reset, we should find a way to do that instead.

mstormi:
just found out about log4j2.xml to replace it - you should really have pointed that out better as a breaking change

ok then this PR does not make much sense or is there a tool to convert old logging config into new one ?

wborn:
> you should really have pointed that out better as a breaking change

It probably went a bit under the radar and I had almost forgotten about this change (it's been like this since Feb '20). We can surely highlight it in the release notes and documentation (https://github.com/openhab/openhab-docs/issues/1163) which will help everyone who is only just now starting to use OH3.

> I thought that's a case of "lazyness" you didn't want to migrate the config

If I wanted to be lazy I wouldn't put in several hours of testing XML logging configs and migrating the existing config to XML. :wink:

> ok but then this PR does not make much sense or is there a tool to convert old logging config into new one ?

When the `log4j2.xml` properly works after fresh installs and updates, I think it is good enough and there are more important thing to fix at the moment.

mstormi:
yeah agree.
So let's move to something more useful such as documentation .. brrrr ;-)

mstormi:
Ok I tried to put this experience to good use and have quickly created https://github.com/openhab/openhab-docs/pull/1303 to reflect this for everybody.
Wouter I've commented out the part from the logback section on to the end.
Can you validate what of that is still ok for OH3 and/or replace it with some (shortish) docs on the XML format ?

wborn:
OK great! Note that the change for the distro is that we've switched from a Log4j (1) .properties configuration to a Log4j2 XML configuration. The Log4j2 XML config is more common and is also more easy to read and use. It may also one day allow us to separate the default and user customized logging configuration using XML includes.

Logback is another logging framework (like Log4j2) that can also be used for logging with the commonly used SLF4J API in Java code. Logback is still used as logging framework in some unit tests and when running the development Demo App.

mstormi:
Struggling to change default logging config in openHABian (or maybe OH's default log4j2.xml ?),
we should at least have it gzip rotated files. 
But I'm overwhelmed by the new format. @wborn do you have a fast path tip ?

mstormi:
Struggling to change default logging config in openHABian (or maybe OH's default log4j2.xml ?),
we should at least have it gzip rotated files, plus explicitly defined the number of logs to keep. 
But I'm overwhelmed by the new format. @wborn do you have a fast path tip ?

mstormi:
Ok I made it
https://github.com/openhab/openhab-distro/pull/1312

Issue #1192, opened 2020-11-28
migration path / namespace migration


mstormi:
The overall namespace change from `org.eclipse.smarthome` to `org.openhab` needs to be reflected in JSONDB and possibly more locations. This means to a) change some config file names and b) change namespaces inside these.
See https://community.openhab.org/t/oh2-things-gone-missing-in-oh3/109094/2

I'd think the best place to migrate a config is the restore script `distributions/openhab/src/main/resources/bin/restore`, wdyt ?

kaikreuzer:
Could you be more specific, please? 
This should all already be covered in https://github.com/openhab/openhab-distro/pull/1030 and https://github.com/openhab/openhab-distro/pull/1057.

mstormi:
Short of a migration tool, I installed 3.0M3 package from scratch and copied my OH2 config file over next (I actually used openhab-cli restore albeit fixed the path - see also #1190)
The JSONDB file was of course still named org.eclipse.smarthome.yaddayadda and also contained several strings with the old namespace, too. How is that supposed to be changed if not by the restore tool ?

kaikreuzer:
Copying openHAB 2.x config files onto a 3.x installation is simply not a supported use case.
Only the mechanisms that make use of the update.lst file are supported, because that's the file that contains all relevant information on how to migrate user data.
If you think it would be valuable to have a script that does a data migration from version X to Y, this can theoretically be implemented - it's just that somebody would need to find the time to create and maintain it...

mstormi:
but isn't this what any user would do ? I wouldn't even know what the proper migration path would be i.e. what the "mechanisms that make use of the update.lst file" are

kaikreuzer:
> but isn't this what any user would do by intuition ?

Hopefully not, because it was never ever supported to update your openHAB installation that way.

> I don't even know what the proper migration path would be

Well, clearly the ones that are the documented update mechanisms:

https://www.openhab.org/docs/installation/linux.html#upgrade
https://www.openhab.org/docs/installation/linux.html#apt-based-systems-2

To be honest, I am not sure how well they work for a 2.x to 3.x switch, especially as the linux package names have changed (@BClark09 is the expert here).
[...]

PR #1199, opened 2020-12-03
have restore script migrate OH2 config to OH3

mstormi:
Relates: #1192 

### DIFF
--- a/distributions/openhab/src/main/resources/bin/restore
+++ b/distributions/openhab/src/main/resources/bin/restore
@@ -1,6 +1,15 @@
 #!/bin/sh
 
 # taken from postinst script
+
+getVersionNumber(){
+    firstPart="$(echo "$1" | awk -F'.' '{print $1}')"
+    secondPart="$(echo "$1" | awk -F'.' '{print $2}')"
+    thirdPart="$(echo "$1" | awk -F'.' '{print $3}')"
+    thirdPart="${thirdPart%%-*}"
+    echo $((firstPart*10000+secondPart*100+thirdPart))
+}
+
 runCommand() {
     string="$1"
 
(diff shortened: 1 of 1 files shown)


distributions/openhab/src/main/resources/bin/restore
```diff
@@ -1,5 +1,109 @@
 #!/bin/sh
 
+# taken from postinst script
+
+getVersionNumber(){
```
BClark09, line 5: Is it worth moving these functions to a shared file and sourcing it? That way, changes made to functions only need to be made once and all you have to do in this script is provide previousVersion and currentVersion parameters.

distributions/openhab/src/main/resources/bin/restore
```diff
@@ -12,6 +116,13 @@ setup(){
     exit 0
   fi
 
+  # change file locations to read defaults from based on installed version
+  # according to dpkg only either "openhab" or "openhab2" packages are installed
+  ohPkgName=openhab2
```
BClark09, line 121: You won't need to check for this, the master branch is for OH3, the 2.5.X branch is for OH2. So this will really only be "openhab"

distributions/openhab/src/main/resources/bin/restore
mstormi: Certainly so but the origin is the postinst file from openhab-linuxpkg so that should be changed in the same go to use the same shared files. But as that's in another repo I was unsure how to handle that best as I'm new to all the code in these repos. Would you mind to consider this a proof of concept (code worked in my test) and take the ball from here before I hack&slay around too much?

distributions/openhab/src/main/resources/bin/restore
mstormi: you sure? I have migrated to OH3 (the "proper" way this time) and the "backup" and "restore" scripts both are located in the OH3 location
/usr/share/openhab(not "2")/runtime/bin/ BUT they still source the env variables from /etc/profile.d/openhab2.sh so backup the OH2 installation not my OH3 one

distributions/openhab/src/main/resources/bin/restore
mstormi: Ben is that ok can I leave that "outsourcing" to you ?

distributions/openhab/src/main/resources/bin/restore
BClark09: > Are you sure? 

The files that you are editing now get added to a zip file ready for use, at the moment they still contain old paths which is the issue here.

linuxpkg takes that zip and puts the files in the correct locations, it is not aware of the contents of these files and does not change them, so the issue of having outdated references is propagated to the apt install.

The 2.5.x branch in openhab-linuxpkg uses the zip file created by the 2.5.x branch of openhab-distro.
The master branch in openhab-linuxpkg uses the zip file created by the master branch of openhab-distro, this is the up-to-date build for openHAB 3

distributions/openhab/src/main/resources/bin/restore
BClark09: I may be able to create a PR this week but I cannot edit this branch as I am not a maintainer of openhab-distro.

mstormi: oh I have ever been thinking you are that's why I kept poking at you :) Sorry!
Yes if you could make that a PR I'd be glad. I don't want to screw things up while we're rushing up on the finishing line.

mstormi:
Superseded

PR #1204, opened 2020-12-07
Remove "2.x.x" in PowerShell update script header

ghys:
Updated a few comments as well - even if they're invisible to the user.


kaikreuzer (approved):
Thanks!

PR #1206, opened 2020-12-07
Remove SmartHome leftovers

wborn:
Depends on https://github.com/openhab/openhab-core/pull/1901

### DIFF
--- a/distributions/openhab/src/main/resources/userdata/etc/log4j2.xml
+++ b/distributions/openhab/src/main/resources/userdata/etc/log4j2.xml
@@ -55,23 +55,23 @@
 
 		<Logger level="INFO" name="org.openhab"/>
 
-		<Logger level="ERROR" name="smarthome.event.ItemStateEvent"/>
-		<Logger level="ERROR" name="smarthome.event.ItemAddedEvent"/>
-		<Logger level="ERROR" name="smarthome.event.ItemRemovedEvent"/>
-		<Logger level="ERROR" name="smarthome.event.ItemChannelLinkAddedEvent"/>
-		<Logger level="ERROR" name="smarthome.event.ItemChannelLinkRemovedEvent"/>
-		<Logger level="ERROR" name="smarthome.event.ThingStatusInfoEvent"/>
-		<Logger level="ERROR" name="smarthome.event.ThingAddedEvent"/>
-		<Logger level="ERROR" name="smarthome.event.ThingUpdatedEvent"/>
-		<Logger level="ERROR" name="smarthome.event.ThingRemovedEvent"/>
-		<Logger level="ERROR" name="smarthome.event.InboxUpdatedEvent"/>
-		<Logger level="ERROR" name="smarthome.event.RuleStatusInfoEvent"/>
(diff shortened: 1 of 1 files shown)


wborn:
This one can also be merged now @openhab/distro-maintainers.

kaikreuzer (approved):
sure, thanks!

kaikreuzer:
@wborn I think there is some issue with logging when upgrading, see https://community.openhab.org/t/openhab-3-0-release-candidates-discussion/110200/28.
I actually experienced the same and was wondering what is happening. I can confirm that the update installed the latest log4j2.xml in my etc folder (containing `openhab.event), but still, I didn't see any logging in console or events.log.
I fiddled a bit around with log settings in my console and stuff started to appear. Do you have any explanation (and a possible solution) for this?

wborn:
My guess would be that the logging not always showing and rules not being loaded on first startup are regressions introduced by  https://github.com/openhab/openhab-core/pull/1914.

kaikreuzer:
Good point - I'll look into that.

kaikreuzer:
... but at the same time, there's still an issue with the logging configuration being changed/picked up, see https://community.openhab.org/t/openhab-3-0-release-candidates-discussion/110200/44.

Reading it, it reminded me that I had the same - it was not just a "first startup" issue.

kaikreuzer:
I just updated an older OH3 snapshot and noticed that the log4j2.xml is not updated at all.
The issue is most likely that it is not a [system file](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/userdata_sysfiles.lst) and is thus not overwritten automatically. I guess we hence have to add it to https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/update.lst, but must be clear that any user specific settings will be lost.

wborn:
We never added `org.ops4j.pax.logging.cfg` to the system files either. Instead we sometimes added a command to reset it to the defaults. If it was a system file users would have their custom config reset after every upgrade.

See: https://github.com/openhab/openhab-distro/blob/e20362081ebdacea67f7a09c6a6592a88fd7e771/distributions/openhab/src/main/resources/bin/update.lst#L43

So we probably want to reset both `org.ops4j.pax.logging.cfg` and `log4j2.xml` to make sure that everyone uses the correct logging config after an upgrade.

kaikreuzer:
Sounds like a plan - can you do a PR?

wborn:
I can add it but after looking into it the instruction won't change much for DEB/RPM users, see https://github.com/openhab/openhab-linuxpkg/pull/181. Those users were probably already prompted by the package manager to inspect the changes. They probably have developed a habit to always reject changes to the logging configuration. So if there is something that should really be applied like this configuration change it is also rejected by them. But the next time they upgrade they can also answer the question again with the right answer and it should be fixed for them. :wink:

BClark09:
I'm unsure how to resolve that. I agree that files such as that should really be forced to change, but working out a way that allows DEB and RPM package managers to keep them listed as configuration files is complicated.

PR #1213, opened 2020-12-15
Use default logging configuration after update

wborn:
Adds DEFAULT instructions for:

* log4j2.xml
* org.ops4j.pax.logging.cfg

---

See: https://github.com/openhab/openhab-distro/pull/1206#issuecomment-745427684


kaikreuzer (approved):

BClark09:
DEFAULT commands must go into `PRE` instead of `POST`. This change will leave no `.xml` or `.cfg` file in place after the update is finished.

https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/update#L148

The DEFAULT command works by renaming the file in the PRE stage, the update script then adds any configuration file that is missing.

BClark09:
Made #1214, if you agree :)

kaikreuzer:
Sounds very worthwhile, thanks ;-)

PR #1214, opened 2020-12-16
Move logging DEFAULT commands from POST to PRE

BClark09:
Replaces #1213 

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -42,6 +42,10 @@ ALERT;Vitotronic Binding: The following channels have been renamed: 'outsite_tem
 [2.2.0]
 DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg
 
+[3.0.0]
+DEFAULT;$OPENHAB_USERDATA/etc/log4j2.xml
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):

wborn:
I hope it works because it looks like it will only restore the original file if it is in the list of system files. WDYT? :thinking: 

https://github.com/openhab/openhab-distro/blob/a46f10257474c2fa7f790331e124036bee3bf1c0/distributions/openhab/src/main/resources/bin/update#L311-L317

BClark09:
Should do! 😉  The main upgrade lines are:
https://github.com/openhab/openhab-distro/blob/a46f10257474c2fa7f790331e124036bee3bf1c0/distributions/openhab/src/main/resources/bin/update#L324-L326

The lines before are the PRE section, and the lines after are the POST section.

In order the update process should be:

- The DEFAULT commands in `PRE` renames each file.
- All files in runtime are moved to a temp folder (system files)
- All files listed in `filelist` are moved to a temp folder (system files un userdata) (the lines you quoted).
- The lines at the top of this post are run, and any files missing are replaced (all of the above files moved or renamed).
- The `POST` section is run.

wborn:
Yes I agree that it should then work! :slightly_smiling_face: The update of existing data works a bit different in the Docker container as it will not download a .zip but uses files that are already part of the Docker image. But I made sure that the DEFAULT command also works well with Docker.

PR #1216, opened 2020-12-18
Rollover log files on startup

pfink:
This is something that I use for my setup, as it makes investigating log files much easier because they are rollled over on startup. This means, you really start with a "clean" log file when you restart openHAB. I used to do the rollover manually in the past as I didn't had the idea that it's configurable via log4j, and probably many users often do so as well.

First, I wanted to put it in the docs, but then I thought: Why not using it as default? I think the majority of users will like it more than the current default.


kaikreuzer (approved):
I like this a lot as well, so let's get that in as the default!

PR #1224, opened 2020-12-23
Create archive with Demo files

### LINKED ISSUE #1147: Create archive with Demo files
As discussed in [openhab-webui#306](https://github.com/openhab/openhab-webui/pull/306#issuecomment-678621976) it might be better to provide the Demo setup using an archive instead of a package because it is complicated to switch from the "demo" to the "standard" package afterwards. The new UI also no longer allows for selecting the package.

wborn:
Fixes #1147

This will probably be a good starting point for an archive with Demo files.

If we add files to `userdata`, then they can be overlayed on top of an existing installation and the Demo could be more easily modified.


wborn:
Do you think this can be merged @openhab/distro-maintainers?

distributions/openhab-demo/src/main/resources/userdata/config/org/openhab/addons.config
```diff
@@ -0,0 +1,8 @@
+:org.apache.felix.configadmin.revision:=L"1"
```
kaikreuzer, line 1: I'm wondering if it is advisable to include this file in a demo package. People might get the impression that this file is meant to be manually created/maintained.
Wouldn't it be better to stick to the `addons.cfg` file? Yes, the downside would be that people cannot easily change the installed add-ons, but if we consider the demo package to also be an example of how to package openHAB configuration as text files, it would imho more appropriate.

kaikreuzer (changes requested):

distributions/openhab-demo/src/main/resources/userdata/config/org/openhab/addons.config
wborn: OK no problem for me. I will just put an overlay on top of this overlay to make it workable for me. :smile:

distributions/openhab-demo/src/main/resources/conf/services/addons.cfg
```diff
@@ -0,0 +1,39 @@
+# The installation package of this openHAB instance
+# Note: This is only regarded at the VERY FIRST START of openHAB
+# Note: If you want to specify your add-ons yourself through entries below, set the package to "minimal"
+# as otherwise your definition might be in conflict with what the installation package defines.
+#
+# Valid options:
+#   - standard : Standard setup for normal use of openHAB
+#   - minimal  : Installation of core components, but no UIs or other add-ons. Use this for special headless setups.
+#   - demo     : A pre-configured demo setup which can be used for demonstration.
```
kaikreuzer, line 9: This PR would be a very good opportunity to remove this line :-D
Also in https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/conf/services/addons.cfg, of course.

kaikreuzer (changes requested):

distributions/openhab-demo/src/main/resources/conf/services/addons.cfg
wborn: Ah yes. Good catch! I never read docs. :wink:

kaikreuzer (approved):
Thanks!

PR #1239, opened 2021-01-05
Add --non-interactive CLI argument to backup script, enabling backups from GUI

robclark56:
See more details here.
https://community.openhab.org/t/when-will-backup-restore-be-in-openhab-gui/112816/11

### DIFF
--- a/distributions/openhab/src/main/resources/bin/backup
+++ b/distributions/openhab/src/main/resources/bin/backup
@@ -1,5 +1,33 @@
 #!/bin/sh
 
+usage() {
+   echo "Usage: backup [--noninteractive] [--full] [filename]"
+   echo ""
+   echo "  e.g. ./backup                     << Makes a file with a timed filename"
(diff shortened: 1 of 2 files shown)


BClark09:
Hi @robclark56,

The change is a good idea but what you've suggested forces specific paths to be used. You can be more general and add a "--non-interactive" option that simply skips the prompts. That way, what you've implemented can be done with:

```
./backup --non-interactive <path>
```

robclark56:
Updated with changes recommended by @BClark09

kaikreuzer (approved):
Many thanks and sorry for the delayed merge!

kaikreuzer:
Argh, well, I actually cannot merge, because the DCO check fails.
@robclark56 Could you please add a sign-off (either to your commits or as a comment here on the PR, which I'll then attach to your squashed commits). Thanks!

kaikreuzer:
Thanks!

PR #1240, opened 2021-01-07
Monitor log4j2.xml for changes

### LINKED ISSUE #1187: Update of log4j2.xml does not trigger config update
In OH2 when you updated org.ops4j.pax.logging.cfg the logging config would be updated immediately.  In OH3, when you update log4j2.xml, it is not detected.  "touch org.ops4j.pax.logging.cfg" forces a reread.  It looks like the xml needs to be added to the list to watch for updates.

https://github.com/openhab/openhab-distro/pull/1078 made the change to the xml.

wborn:
Fixes #1187

We might want to reset the logging config again. If we do that now it will reset everyones logging configuration again everytime they update from one snapshot to another. But if we don't do it now it will also not fix the issue immediately. :-/

### DIFF
--- a/distributions/openhab/src/main/resources/userdata/etc/log4j2.xml
+++ b/distributions/openhab/src/main/resources/userdata/etc/log4j2.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?><Configuration>
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><Configuration monitorInterval="10">
 
 	<Appenders>
 		<!-- Console appender not used by default (see Root logger AppenderRefs) -->


distributions/openhab/src/main/resources/userdata/etc/log4j2.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?><Configuration>
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><Configuration monitorInterval="10">
```
cweitkamp, line 1: The value is in seconds. Is it?

cweitkamp (approved):
LGTM. Thanks.

cweitkamp:
We can create a new issue and add it to https://github.com/orgs/openhab/projects/4 (or just add a new to-do card) as a reminder to add it ight before next release.

PR #1247, opened 2021-01-13
[miio] add alert on miio channels changing

marcelrv:
related to  ao https://github.com/openhab/openhab-addons/pull/9756 and https://github.com/openhab/openhab-addons/pull/9384

Signed-off-by: Marcel Verpaalen <marcel@verpaalen.com>

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -38,6 +38,9 @@ ALERT;TP-Link Smart Home Binding: The energy and rssi channels are now using Uni
 ALERT;Tesla Binding: The binding has been heavily refactored. It is required to delete old Tesla Things and set them up from scratch.
 ALERT;Vitotronic Binding: The following channels have been renamed: 'outsite_temp' to 'outside_temp', 'pelletburner:power' to 'pelletburner:powerlevel', 'party_temp' to 'party_temp_setpoint' and 'save_temp' to 'save_temp_setpoint'
 
+[3.1.0]
+ALERT;Xiaomi MiIo Binding: Many channels have been converted from Number to a Number with dimension, e.g. temperature channels have been changed from Number to Number:Temperature. Please see the binding documentation for details.
+
 [[PRE]]
 [2.2.0]
 DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg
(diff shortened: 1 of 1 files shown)


distributions/openhab/src/main/resources/bin/update.lst
```diff
@@ -46,6 +46,9 @@ DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg
 DEFAULT;$OPENHAB_USERDATA/etc/log4j2.xml
 DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg
 
+[3.1.0]
+ALERT;Xiaomi MiIo binding: With 3.1 release many channels have been converted from Number to a quanityType number. e.g. temperature channels have been changed from Number to Number:Temperature. You need to update your item definition to the new definitions. Please see the binding readme for the item details for your device.
```
kaikreuzer, line 50:
```suggestion
ALERT;Xiaomi MiIo Binding: Many channels have been converted from Number to a Number with dimension, e.g. temperature channels have been changed from Number to Number:Temperature. Please see the binding documentation for details.
```

kaikreuzer (changes requested):
Question: Is this really a breaking change?
It should always be possible to link a dimensionless item to such a channel, so ideally this should be backward compatible.

marcelrv:
> Question: Is this really a breaking change?
> It should always be possible to link a dimensionless item to such a channel, so ideally this should be backward compatible.

To be honest I'm still not 100% clear indeed if it is breaking or not. I recall I had some issues with rules after the change overs.
Nevertheless, I'm very fine with the rewording you did.

distributions/openhab/src/main/resources/bin/update.lst
```diff
@@ -46,6 +46,9 @@ DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg
 DEFAULT;$OPENHAB_USERDATA/etc/log4j2.xml
 DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg
 
+[3.1.0]
```
kaikreuzer, line 49: I'd think this sections needs to be moved to https://github.com/openhab/openhab-distro/pull/1247/files#diff-8be360a52adcae6cc237c7d0e4658b2d50028192596d6489f64ee61e4d43f4aaR40, i.e. under `[[MSG]]` and not under `[[PRE]]`.

kaikreuzer (changes requested):

kaikreuzer (approved):
Thanks!

PR #1249, opened 2021-01-17
Add parameter "enable CORS" to runtime.cfg file

philippwaller:
This function is not new, but hardly documented. Since there are already some questions about this parameter, I suggest to mention it in the runtime configuration.


kaikreuzer (approved):
Many thanks!

PR #1250, opened 2021-01-18
Update pipeline and Oomph files to use main instead of master branches

wborn:
Let's rename the branches after creating releases.

### DIFF
--- a/launch/Jenkinsfile
+++ b/launch/Jenkinsfile
@@ -16,7 +16,7 @@ try {
     
     node {
         deleteDir()
-        releaseComponent('openhab-core', false, "main")
+        releaseComponent('openhab-core', false, env.OH_BRANCH)
     }
     
     // checkpoint "openHAB Core released"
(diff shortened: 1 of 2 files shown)


launch/Jenkinsfile
```diff
@@ -25,16 +25,16 @@ try {
     node {
         deleteDir()
         releaseComponent('openhab-addons', true, env.OH_BRANCH)
-        releaseComponent('org.openhab.binding.zwave', true, "master")
-        releaseComponent('org.openhab.binding.zigbee', true, "master")
-        releaseComponent('openhab-webui', true, "master")
+        releaseComponent('org.openhab.binding.zwave', true, "main")
+        releaseComponent('org.openhab.binding.zigbee', true, "main")
+        releaseComponent('openhab-webui', true, "main")
```
wborn, line 30: We probably also want to use `env.OH_BRANCH` everywhere instead of `"main"`

kaikreuzer (approved):

PR #1263, opened 2021-02-24
Update update.lst for 3.1.0 with keba

MikeTheTux:
in relation to https://github.com/openhab/openhab-addons/pull/10179


MikeTheTux:
I did not sign off because imho this is only a single doc change.

kaikreuzer:
Thanks @MikeTheTux. These messages need to go into the [MSG] section of the file - I have placed it there with https://github.com/openhab/openhab-distro/pull/1264, so we can close this PR.

Issue #1256, opened 2021-02-06
Move away from Bintray


kaikreuzer:
Bintray is about to be closed: https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

We will therefore need to remove any links to Bintray and find a suitable substitution.

The simplest option atm seems to be the move to Artifactory (openhab.jfrog.io) - the risk here is though that it is operated by the same company as Bintray, so we might be warned that this could turn out to be a fragile option as well. Since we have all our snapshot and milestone artifacts there already. I'd nonetheless suggest to go this way. It might be a worthwhile effort to check whether we can set up a [CNAME DNS entry](https://jfrog.com/knowledge-base/how-to-configure-custom-cname-for-artifactory-cloud/), so that we would at least be under control of the DNS name and could point it to somewhere else in case the free JFrog Artifactory is ended some day.

A better option for releases would be to directly publish them to Maven Central - I am not sure how easy and stable we could get that into our release build jobs.

This issue is meant as an aggregate to track the efforts that are needed.

- [X] Change "update" script in distro  - #1271
- [x] Change website download location
- [x] Remove mentions of Bintray from documentation/website
- [x] Change "distributionManagement" section in all pom-xmls
- [x] Adapt build plans
- [x] Adapt Docker build/containers (https://github.com/openhab/openhab-docker/pull/356)
- [x]  inform users to change download locations e.g. in their linux distributions (apt/rpm)
- [x] Change openhab-osgiify to not require artifacts from Bintray
- [x] Change artifactory to not include Bintray locations anymore

BClark09:
Also commenting here to track and to mention that Bintray is our stable host for the Linux packages.

Fortunately, our stable packages are available on Artifactory so I suspect a notification to transfer and documentation change is all that's needed here.

We have openHAB 1.x packages available from Bintray. I know that's going back a bit, but should we transfer these?

kaikreuzer:
For the zipped distro and KARs, I am just pushing the latest+major releases (2.5.12, 2.5.0, 2.4.0, ... 2.0.0, 1.8.3) to Maven Central, see https://repo1.maven.org/maven2/org/openhab/distro/openhab/

If it is possible to move the linux packages as well (I guess our Artifactory is the only real target option, right?), it might be nice not to loose them. Is there some mechanism you could use to transfer them?

Orifahrer:
[...]

PR #1271, opened 2021-03-26
Removed Bintray and JFrog dependency from update url

kaikreuzer:
Requires https://github.com/openhab/website/pull/278/files
Related to openhab/openhab-distro#1256

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update
+++ b/distributions/openhab/src/main/resources/bin/update
@@ -95,11 +95,11 @@ setup(){
     echo "You are already on openHAB $CurrentVersion" >&2
     exit 1
   elif [ -n "$milestoneVersion" ]; then
-    DownloadLocation="https://openhab.jfrog.io/openhab/libs-milestone-local/org/openhab/distro/openhab/$OHVersion/openhab-$OHVersion.zip"
-    AddonsDownloadLocation="https://openhab.jfrog.io/openhab/libs-milestone-local/org/openhab/distro/openhab-addons/$OHVersion/openhab-addons-$OHVersion.kar"
+    DownloadLocation="https://www.openhab.org/download/milestones/org/openhab/distro/openhab/$OHVersion/openhab-$OHVersion.zip"
+    AddonsDownloadLocation="https://www.openhab.org/download/milestones/org/openhab/distro/openhab-addons/$OHVersion/openhab-addons-$OHVersion.kar"
   else
-    DownloadLocation="https://bintray.com/openhab/mvn/download_file?file_path=org%2Fopenhab%2Fdistro%2Fopenhab%2F$OHVersion%2Fopenhab-$OHVersion.zip"
-    AddonsDownloadLocation="https://bintray.com/openhab/mvn/download_file?file_path=org%2Fopenhab%2Fdistro%2Fopenhab-addons%2F$OHVersion%2Fopenhab-addons-$OHVersion.kar"
+    DownloadLocation="https://www.openhab.org/download/releases/org/openhab/distro/openhab/$OHVersion/openhab-$OHVersion.zip"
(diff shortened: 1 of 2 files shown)


distributions/openhab/src/main/resources/bin/update
```diff
@@ -95,11 +95,11 @@ setup(){
...
-    DownloadLocation="https://openhab.jfrog.io/openhab/libs-milestone-local/org/openhab/distro/openhab/$OHVersion/openhab-$OHVersion.zip"
-    AddonsDownloadLocation="https://openhab.jfrog.io/openhab/libs-milestone-local/org/openhab/distro/openhab-addons/$OHVersion/openhab-addons-$OHVersion.kar"
+    DownloadLocation="https://www.openhab.org/download/milestones/org/openhab/distro/openhab/$OHVersion/openhab-$OHVersion.zip"
+    AddonsDownloadLocation="https://www.openhab.org/download/milestones/org/openhab/distro/openhab-addons/$OHVersion/openhab-addons-$OHVersion.kar"
   else
-    DownloadLocation="https://bintray.com/openhab/mvn/download_file?file_path=org%2Fopenhab%2Fdistro%2Fopenhab%2F$OHVersion%2Fopenhab-$OHVersion.zip"
-    AddonsDownloadLocation="https://bintray.com/openhab/mvn/download_file?file_path=org%2Fopenhab%2Fdistro%2Fopenhab-addons%2F$OHVersion%2Fopenhab-addons-$OHVersion.kar"
+    DownloadLocation="https://www.openhab.org/download/releases/org/openhab/distro/openhab/$OHVersion/openhab-$OHVersion.zip"
+    AddonsDownloadLocation="https://www.openhab.org/download/releases/org/openhab/distro/openhab-addons/$OHVersion/openhab-addons-$OHVersion.kar"
```
wborn, line 102: It gives a 404 with this URL:

https://www.openhab.org/download/releases/org/openhab/distro/openhab/3.0.1/openhab-3.0.1.zip

There is no content in:

https://openhab.jfrog.io/artifactory/libs-release-local/

wborn (changes requested):

distributions/openhab/src/main/resources/bin/update
kaikreuzer: Yes, because so far no releases have been published there. We will need to do a 3.0.2 patch release in April, before Bintray shuts down our account.

wborn: Will previous releases also be available there before Bintray sunsets?
Some users might like to download them after February 1st, 2022.

wborn: > We will need to do a 3.0.2 patch release in April, before Bintray shuts down our account.

I would also need to update the [URL](https://github.com/openhab/openhab-docker/blob/4b021c7f6ca71ee4e1a1c50b344a3945d3325796/debian/Dockerfile#L93) for the Docker build for this. So I'd have to keep track of multiple download URLs to also support building OH <= 3.0.1 containers. Something similar may need to be done for .deb/.rpms by @BClark09 and others.

kaikreuzer: @wborn Let us better discuss this on https://github.com/openhab/openhab-distro/issues/1256.
This change here is merely about the update script as this is rolled out to the users and hence we have to make sure that all snapshot/milestone/release users that do manual upgrades through this script have the new urls on their local installation already before Bintray is shut down. For this case, I'd think it is ok to only support "future" releases as the script is primarily meant for upgrades, not for downgrades.

wborn: OK then let's merge this but please answer my questions in #1256. The checklist definitely also needs updating download URLs in all repos that package openHAB.

wborn (approved):

PR #1273, opened 2021-03-30
Upgrade to Karaf 4.3.1

### LINKED ISSUE #1167: Upgrade Karaf to 4.3.1
These Karaf versions will be released soon so we should upgrade the Karaf version used in the distro to get the latest bugfixes and enhancements.

wborn:
* Upgrades Karaf to 4.3.1
* Use Pax Logging as runtime dependency instead of Felix Log
  This dependency change makes the Demo App behave more like the actual distro.
  It also allows for using a single log4j2.xml which simplifies the logging config and makes it more like the distro.
  Fixes some noisy Pax Web commons-logging in the Demo App that would occur when using Felix Log.

The Log4j2 Demo App configuration is now very similar to that of the distro except for:

* Karaf specific logger configuration has been removed
* A commented example logger for a binding is added to help with debug logging

To allow for reconfiguring Log4j2 without making changes some enviroment variables can be used:

* The default root/openhab logger level can be overriden
* Several logger level/names can be reconfigured similarly

Fixes #1167


distributions/openhab/src/main/resources/bin/karaf
```diff
@@ -197,10 +199,6 @@ run() {
     nodebug=false
     while [ "${1}" != "" ]; do
         case "${1}" in
-            'clean')
```
kaikreuzer, line 200: Does this mean that "clean" isn't available anymore? Is this a change that Karaf 4.3.1 introduced? Many people who believe "cleaning the cache" solves most issues might be pretty disappointed...

BClark09: ${KARAF_DATA} points to all of ${OPENHAB_USERDATA}. Running this command would typically break openHAB, so this is probably a good thing.

wborn: It's this upstream change 

https://github.com/apache/karaf/commit/aa870306c5bca0f1dd84bf20199fa00d63443c0c#diff-e3e003cfbe43c1837ab85d8f50141dc3f1936658fe35fd50ae637a00a2ae28fc

I think the openhabian scripts have their own cleaning commands to only clean the cache and tmp dir.

kaikreuzer (approved):
Many thanks for taking care of this upgrade!

PR #1278, opened 2021-04-19
Update streams in openHAB2.setup

wborn:
The 3.0.x stream is missing and the main branch has OH 3.1 nowadays.

If 2.5.x no longer compiles we may also want to remove it.

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2452,10 +2452,7 @@
           locateNestedProjects="true"/>
     </setupTask>
     <stream
-        name="main"
-        label="3.0"/>
-    <stream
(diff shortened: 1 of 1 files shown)


kaikreuzer:
3.0.2 will be built tomorrow and I think it'll be the last 3.0.x build.
So I'd actually think that all we need is 3.1 and then it is better to rename it to "main", so that we don't have to update it in future. Wdyt?

wborn:
Yes it probably only makes sense to setup the "main" branches in the IDE because those are currently the only branches on which PRs by contributors are accepted.

kaikreuzer (approved):
Thanks!

PR #1310, opened 2021-07-27
Add --textconfig and --uiconfig options to restore config parts

mstormi:
Depends: openhab/openhab-linuxpkg#206
Depends: https://github.com/openhab/openhab-linuxpkg/pull/206


distributions/openhab/src/main/resources/bin/restore
```diff
@@ -1,16 +1,52 @@
-#!/bin/sh
+#!/bin/bash
```
BClark09, line 1: openHAB is setup to be as platform agnostic as possible. We should keep this script POSIX compliant to ensure compatibility with most systems and to reduce the chance of changes being e.g. MacOS incompatible. This also means that testing with bash `[[ ]]` can't be done.

BClark09: As a side note, these same changes would be ideally be replicated in Windows' PowerShell scripts if possible.

BClark09 (changes requested):
Thanks @mstormi, I agree that partial restoration would be a nice inclusion in the scripts. I haven't tested the new script itself but had some comments on specific changes.

You've removed executable permissions from:

- `distributions/openhab/src/main/resources/bin/backup`
- `distributions/openhab/src/main/resources/bin/karaf`
- `distributions/openhab/src/main/resources/bin/setenv`
- `distributions/openhab/src/main/resources/bin/update`
- `distributions/openhab/src/main/resources/userdata/etc/scripts/shell.completion.script`

Was there a reason for this?

distributions/openhab/src/main/resources/bin/restore
mstormi: > You've removed executable permissions from:
> Was there a reason for this?

No it was an accident I didn't even notice. I've changed that back.

mstormi: > openHAB is setup to be as platform agnostic as possible. We should keep this script POSIX compliant to ensure
> compatibility with most systems and to reduce the chance of changes being e.g. MacOS incompatible.

POSIX feels so 80s and bash is really available everywhere, MacOS even used it as their standard.
Then again you're right keeping POSIX compliance is a good approach and no point in having an argument over this so I changed it back.

mstormi: > As a side note, these same changes would be ideally be replicated in Windows' PowerShell scripts if possible.

I agree but I quickly glanced at the script before and decided I cannot provide that in reasonable time and quality.
I have no experience with PowerShell at all so sorry we need a Win/PS expert to do this. I wouldn't even know about Windows directory locations.

mstormi: Please also test, it's as simple as to create an OH backup, delete any config file and restore from that using the new options.
For --textconfig you can keep OH running.

mstormi:
@BClark09 anything open here ?

mstormi:
@BClark09 Ben would you care to approve ? If no let me know what's missing please.

PR #1311, opened 2021-08-17
Fix %CLASSPATH% compatibility

Ryz3D:
Hi,

I ran into an issue where running start.bat (Windows 10, 64 bit, JDK 11.0.11) resulted in an error like:
`"\QuickTime\QTSystem\QTJava.zip" kann syntaktisch an dieser Stelle nicht verarbeitet werden.`
(roughly translates to "cannot be processed syntactically at this point.").
It turned out the problem was my %CLASSPATH% environment variable. It has this value currently:
`.;slf4j-simple.jar;C:\Program Files (x86)\QuickTime\QTSystem\QTJava.zip`
and when executing line 394 of karaf.bat:
`set CLASSPATH=%CLASSPATH_INITIAL%`
it ran into said issue. I believe it's failing due to the space in the variable or the backslashes, though I am not sure. I have resolved it by adding quotation marks around the `%CLASSPATH_INITIAL%` variable. Don't ask me why line 106
`set CLASSPATH_INITIAL=%CLASSPATH%`
runs without issue despite using `%CLASSPATH%` without quotation marks.
This is obviously outside my area of expertise, but this one fix has made the difference between aforementioned crash and openHAB running fine on my system and I suppose space characters in the classpath could be expected on Windows.

Thank you.

### DIFF
--- a/distributions/openhab/src/main/resources/bin/karaf.bat
+++ b/distributions/openhab/src/main/resources/bin/karaf.bat
@@ -391,7 +391,7 @@ if "%KARAF_PROFILER%" == "" goto :RUN
         MOVE /Y "%KARAF_HOME%\lib.next" "%KARAF_HOME%\lib"
 
         echo "Updating classpath..."
-        set CLASSPATH=%CLASSPATH_INITIAL%
+        set CLASSPATH="%CLASSPATH_INITIAL%"
         pushd "%KARAF_HOME%\lib\boot"
         for %%G in (*.jar) do call:APPEND_TO_CLASSPATH %%G
         popd


martinvw:
@wborn looks reasonable to me, wdyt?

wborn:
I think so, but I'm not a Windows user. Also it would be better to fix this with Karaf upstream, otherwise this fix might get lost in the next Karaf upgrade. :wink:

Ryz3D:
alright, i wasn't quite sure where karaf.bat came from (i still am not, could someone point me to the correct repo? ^^). will open a PR to that then.

wborn:
This is the file in the upstream repo:

https://github.com/apache/karaf/blob/main/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf.bat

There are also some other files that build a `%CLASSPATH%`. Maybe they also have issues when a directory name contains spaces?

```
assemblies/features/base/src/main/filtered-resources/resources/bin/karaf.bat
assemblies/features/base/src/main/filtered-resources/resources/bin/instance.bat
assemblies/features/base/src/main/filtered-resources/resources/bin/client.bat
assemblies/features/base/src/main/filtered-resources/resources/bin/shell.bat
```

It might take a while before a new Karaf version is released with the fix, so if you reopen the PR in this repo, then we can already use your fix. :slightly_smiling_face:

J-N-K:
I have just tried to reproduce the issue to supply an upstream-patch, but your fix doesn't solve it for me. It complains about a lot of other places, too. If I enclose all of them with quotes, then it doesn't start because a space in the installation path (which works fine when no CLASSPATH is set) throws a ClassNotFoundException.

PR #1312, opened 2021-09-01
gzip logs as per default log4j2 config

mstormi:
Primarily to save on disk space or actually on RAM on openHABian-on-Raspi which has this in (Z)RAM 

Also explicitly defined the number of logs := 7 so people know how to change that if they want to.

Fixes: openhab/openhabian#1575

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -55,6 +55,9 @@ DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg
 DEFAULT;$OPENHAB_USERDATA/etc/log4j2.xml
 DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg
 
+[3.2.0]
+DEFAULT;$OPENHAB_USERDATA/etc/log4j2.xml
+
 [[POST]]
 [2.3.0]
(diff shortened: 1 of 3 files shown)


wborn:
I'm fine with the changes, but we probably also want to apply them to upgrades and not only new installations.

So it needs an entry for 3.2.0 to update.lst similar to: 

https://github.com/openhab/openhab-distro/blob/492a502951b3f9fcc01cab15734329a01f809e5d/distributions/openhab/src/main/resources/bin/update.lst#L55

mstormi:
> So it needs an entry for 3.2.0 in update.lst similar to:

sorry I didn't catch that. *Where* would that entry need to be made ? (probably easiest when you do yourself, maintainer edit is enabled).
The sync addition I've made.

wborn:
Good good :+1: 

Just add it to 3.2.0 in the `[[PRE]]` section, like this :

```
[[PRE]]
[2.2.0]
DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg

[3.0.0]
DEFAULT;$OPENHAB_USERDATA/etc/log4j2.xml
DEFAULT;$OPENHAB_USERDATA/etc/org.ops4j.pax.logging.cfg

[3.2.0]
DEFAULT;$OPENHAB_USERDATA/etc/log4j2.xml
```

mstormi:
ok

wborn (approved):
Thank you!

PR #1314, opened 2021-09-17
[homematic] Add alert about DOUBLE_PRESS trigger event removal

maniac103:
See https://github.com/openhab/openhab-addons/pull/11186

Signed-off-by: Danny Baumann <dannybaumann@web.de>


kaikreuzer (approved):
Thanks!

PR #1319, opened 2021-09-25
Add Marketplace to Demo App

wborn:
Related to:

* openhab/openhab-core#2405
* openhab/openhab-core#2491

---

It was also possible to install an add-on test bundle since those are installed using the bundleContext. :slightly_smiling_face:


kaikreuzer (approved):

ghys:
Sorry I intended to do this PR previously - in fact I was about to do it - that but forgot... I agree it's a good fit since it doesn't depend on Karaf and can install real bundles, it can come in handy when testing stuff 👍

PR #1329, opened 2021-11-07
Fix SNAPSHOT KAR update

wborn:
This should also have been updated in https://github.com/openhab/openhab-distro/pull/1131

There was a similar issue in the Docker container. :blush:
See: https://github.com/openhab/openhab-docker/pull/368


kaikreuzer (approved):
Thanks!
Seems hardly anyone is using the KAR as this hasn't been noticed all the time...

wborn:
> Seems hardly anyone is using the KAR as this hasn't been noticed all the time

Yes but only a subset of KAR file users would run into this (snapshot + Docker/manual install). The snapshot .deb files provided by the APT repo don't seem to have this issue, which is I think the most popular installation method.

PR #1340, opened 2021-12-05
This adds a new logger to our default log configuration that can be utilized by user provided scripts.

digitaldan:
The JSScripting rules binding will be introducing nodejs compatible logging, so `console.log`, `console.debug`, `console.error` and so on.  This PR adds a default namespace for user script logging  which can be used by JSScripting as well as any other scripting engine.  The only messages logged using this namespace will be provided by users, which is why its severity level is set as it is. 

See https://github.com/openhab/openhab-addons/pull/11656 for the corresponding changes to the JSScripting binding.  

Signed-off-by: Dan Cunningham <dan@digitaldan.com>


kaikreuzer (approved):

digitaldan:
👍

Issue #1349, opened 2021-12-16
Properly nail log4j security leak


splatch:
Current fix for log4j CVE provided via #1346 is not complete and definitive since there is still an area which allows to exploit vulnerability.

jcz1:
FYI, Severity of CVE-2021-45046 is now Critical even with log4j 2.15 and LOG4J_FORMAT_MSG_NO_LOOKUPS

https://logging.apache.org/log4j/2.x/security.html

kaikreuzer:
Where does it say critical? I only see "moderate severity" and only for use cases that do not apply to us.
On https://nvd.nist.gov/vuln/detail/CVE-2021-45046, it even says low with a base score of 3.7.

jcz1:
F5 maybe? :-)

"Severity is now Critical

CVE-2021-45046 | Remote Code Execution
Severity | Critical
Base CVSS Score | 9.0 (AV:N/AC:H/PR:N/UI:N/S:C/C:H/I:H/A:H)
Versions Affected | All versions from 2.0-beta9 to 2.15.0, excluding 2.12.2"

"The original severity of this CVE was rated as Moderate; since this CVE was published security experts found additional exploits against the Log4j 2.15.0 release, that could lead to information leaks, RCE (remote code execution) and LCE (local code execution) attacks.

Base CVSS Score changed from 3.7 (AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:L) to 9.0 (AV:N/AC:H/PR:N/UI:N/S:C/C:H/I:H/A:H).

The title of this CVE was changed from mentioning Denial of Service attacks to mentioning Remote Code Execution attacks."

"Other insufficient mitigation measures are: setting system property log4j2.formatMsgNoLookups or environment variable LOG4J_FORMAT_MSG_NO_LOOKUPS to true for releases >= 2.10, or modifying the logging configuration to disable message lookups with %m{nolookups}, %msg{nolookups} or %message{nolookups} for releases >= 2.7 and <= 2.14.1."

kaikreuzer:
Ok, thanks. F5 does not help - seems I am on a CDN that still serves the previous version...

splatch:
Solved through #1354 and #1355,

PR #1344, opened 2021-12-11
Upgrade to Karaf 4.3.4

wborn:
* Syncs distro customizations with Karaf 4.3.4
* Resolves app runbundles for the new dependencies

For release notes, see:

https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12311140&version=12350547

Karaf 4.3.4 uses Pax Logging 2.0.12 which fixes CVE-2021-44228 and CVE-2021-45046.

See also these advisories:

* Log4j: [GHSA-jfh8-c2jp-5v3q](https://github.com/advisories/GHSA-jfh8-c2jp-5v3q), [GHSA-7rjr-3q55-vv33](https://github.com/advisories/GHSA-7rjr-3q55-vv33)
* Pax Logging: [GHSA-xxfh-x98p-j8fr](https://github.com/advisories/GHSA-xxfh-x98p-j8fr)

Fixes #1334

---

Depends on:

* https://github.com/openhab/openhab-core/pull/2603
* https://github.com/openhab/openhab-addons/pull/11750
* https://github.com/openhab/openhab-webui/pull/1228

PR #1350, opened 2021-12-16
Override pax logging version to address #1349.

splatch:
Signed-off-by: Łukasz Dywicki <luke@code-house.org>


splatch:
Note - pax 2.0.10 is still listed in contents of archive, can be trimmed by ant or assembly mask. Runtime picks up 2.0.11 which is CVE clear. Feel free to backport it to 3.0.x and 3.1 versions. ;-)

distributions/openhab/processing.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<featuresProcessing xmlns="http://karaf.apache.org/xmlns/features-processing/v1.0.0">
+
+    <bundleReplacements>
+        <bundle originalUri="mvn:org.ops4j.pax.logging/pax-logging-api/[0,2.0.11)" replacement="mvn:org.ops4j.pax.logging/pax-logging-api/2.0.11" mode="maven" />
```
J-N-K, line 5: Why not use 2.0.12?

splatch: I could swear 2.0.11 was fine two days ago ;-), lets go with 2.0.12 then!

kaikreuzer:
Thanks for this fix, @splatch!
I'll keep this as a fallback, if we run into any issues with https://github.com/openhab/openhab-distro/pull/1344.
But as Karaf 4.3.4 fixes some other issues as well, the plan would still be to release openHAB 3.2 directly with Karaf 4.3.4, so this additional fix won't be necessary.
Wrt backporting: Afaics, the criticality for openHAB isn't that high as log4j 2.15 already fixed the main/critical issue.
Assuming that most people on 3.x will anyhow update to 3.2, the effort to do yet another round of patch releases might not be worth it.

jcz1:
FYI 2.0.11 is vulnerable to CVE-2021-45046 (score changed to 9.0 / critical).

https://github.com/ops4j/org.ops4j.pax.logging/issues/420

wborn:
Thanks for the PR @splatch! I've made a few improvements in https://github.com/splatch/openhab-distro/pull/6, just in case the Karaf committee tries to spoil our release party. :partying_face:

splatch:
Looks good, to be fair I didn't know the `exclude` stuff you used.
I don't think there are any show stoppers for Karaf 4.3.4 to get released, unless another surprise gets out of its dependencies ;-)

splatch:
Solved through #1354

Issue #1087, opened 2020-03-12
Update issue from 2.4.0 to 2.5.2 - Could not backup existing distribution


Vonken:
![image](https://user-images.githubusercontent.com/5111849/76511124-f775a680-6452-11ea-92bc-1c1fc7c598dd.png)

wborn:
That path seems to exceed the max path length of Windows that is 260 chars according to these [docs](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation).

There are 264 chars in:

`C:\Users\Familjen i Lerkil\AppData\Local\Temp\openhab\backup-2.4.0\runtime\system\org\eclipse\smarthome\automation\org.eclipse.smarthome.automation.module.script.defaultscope\0.10.0.oh240\org.eclipse.smarthome.automation.module.script.defaultscope-0.10.0.oh240.jar`

Perhaps [enabling longer paths](https://mspoweruser.com/ntfs-260-character-windows-10/) can be used as a workaround?

Vonken:
Thanks for the answer! Did a fresh install and copied all files and settings but this will propebly help me in the next upgrade or someone.

J-N-K:
This also happens if bundles installed from the marketplace (especially kar-files) are used. The reason is that they easily exceed 255 characters as they are unpacked to `<userdata>/tmp/kar/<filename>`.  Since tmp, cache and backup are removed after copying anyway, this is a bit annoying.

PR #1353, opened 2021-12-18
mitigate long filename issues on Windows backup-script

J-N-K:
Related to #1087 

This does not solve the full issue (which is inherent to Windows insisting on the max path length of 260 characters even if NTFS is capable of 65k characters) but it at least mitigates the issue for backups if the long filenames are in `tmp`, `cache` or `backup` (which I believe is the most common situation).


wborn (approved):
Thanks!

PR #1360, opened 2022-01-08
Remove unused configurator remnants

wborn:
So we don't have to update the year in the license file header every time.
These seem to be some leftovers from trying to get the IDE based launch config working in #905.


kaikreuzer (approved):
Good idea, thanks!

PR #1361, opened 2022-01-10
[tapocontrol] Change naming of things

wildcs:
changed naming of things because manufacturer changed namin in new hw-rev.
now all things have the same naming-style

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -60,7 +60,7 @@ ALERT;Sager Weathercaster Binding: The type of the channels 'pressure', 'tempera
 ALERT;Sony Projector Binding: The type of the channel 'power' was updated. You may need to remove and create again your things in case your things were previously created using UI and you want to use this channel.
 
 [3.3.0]
-ALERT;TapoControl: L510_Series and L530_Series Things were renamed to L510 and L530 because of manufacturer changed naming with new HW-rev. You may need to remove and create again this things.
+ALERT;TapoControl: L510_Series and L530_Series Things were renamed to L510 and L530 because of manufacturer changed naming with new HW-rev. You may need to remove and create again these things.
 
 [[PRE]]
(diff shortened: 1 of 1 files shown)


distributions/openhab/src/main/resources/bin/update.lst
```diff
@@ -59,6 +59,9 @@ ALERT;RFXCOM Binding: Lighting4 default command ids are deprecated and will be r
 ALERT;Sager Weathercaster Binding: The type of the channels 'pressure', 'temperature' and 'wind-angle' was updated. You may need to remove and create again your things in case your things were previously created using UI and you want to use one of these channels.
 ALERT;Sony Projector Binding: The type of the channel 'power' was updated. You may need to remove and create again your things in case your things were previously created using UI and you want to use this channel.
 
+[3.3.0]
+ALERT;TapoControl: L510_Series and L530_Series Things were renamed to L510 and L530 because of manufacturer changed naming with new HW-rev. You may need to remove and create again this things.
```
kaikreuzer, line 63:
```suggestion
ALERT;TapoControl: L510_Series and L530_Series Things were renamed to L510 and L530 because of manufacturer changed naming with new HW-rev. You may need to remove and create again these things.
```

kaikreuzer (changes requested):
Do you have a reference to the PR that brought this breaking change?
Was it really necessary to introduce this breaking change? I would have thought that it shouldn't harm, if we have a "_Series" suffix still in place.

distributions/openhab/src/main/resources/bin/update.lst
wildcs: Hi,
pull request [#11967](https://github.com/openhab/openhab-addons/pull/11967) is already merged.
I think it's the better cleaner way to this new binding. So we do not have a lot of things when adding them manually. This will be confusing if there will be more than one thing for the same type of device.

kaikreuzer (approved):
Alright, thank you!

PR #1366, opened 2022-02-06
only install .jar/.kar from addons folder

J-N-K:
Fixes https://github.com/openhab/openhab-core/issues/380


wborn (approved):
Thanks for the fix!

PR #1371, opened 2022-02-25
Added feature openhab-iconset-classic to ui runtime

cweitkamp:
- Added feature openhab-iconset-classic to ui runtime

When beside Main UI no other UI bundle is installed the icons are not displayed.

Closes https://github.com/openhab/openhab-core/issues/2730

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

### DIFF
--- a/features/distro/src/main/feature/feature.xml
+++ b/features/distro/src/main/feature/feature.xml
@@ -3,9 +3,10 @@
 
     <feature name="openhab-runtime-ui" description="openHAB UI" version="${project.version}">
         <feature>openhab-runtime-base</feature>
-        <feature>openhab-iconset-classic</feature>
         <feature>openhab-core-ui</feature>
+        <feature>openhab-core-ui-icon</feature>
         <bundle>mvn:org.openhab.ui.bundles/org.openhab.ui/${project.version}</bundle>
+        <bundle>mvn:org.openhab.ui.bundles/org.openhab.ui.iconset.classic/${project.version}</bundle>
     </feature>
 
     <feature name="openhab-package-standard" description="openHAB Standard Package" version="${project.version}">
--- a/features/distro/src/main/feature/feature.xml
+++ b/features/distro/src/main/feature/feature.xml
@@ -3,6 +3,7 @@
 
     <feature name="openhab-runtime-ui" description="openHAB UI" version="${project.version}">
         <feature>openhab-runtime-base</feature>
+        <feature>openhab-iconset-classic</feature>
         <feature>openhab-core-ui</feature>
         <bundle>mvn:org.openhab.ui.bundles/org.openhab.ui/${project.version}</bundle>
     </feature>


martinvw (approved):

wborn (changes requested):
Looks like it cannot find this feature:

`[ERROR] Message: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=openhab-runtime-ui; type=karaf.feature; version=3.3.0.SNAPSHOT; filter:="(&(osgi.identity=openhab-runtime-ui)(type=karaf.feature)(version>=3.3.0.SNAPSHOT))" [caused by: Unable to resolve openhab-runtime-ui/3.3.0.SNAPSHOT: missing requirement [openhab-runtime-ui/3.3.0.SNAPSHOT] osgi.identity; osgi.identity=openhab-iconset-classic; type=karaf.feature]`

cweitkamp:
The feature `openhab-iconset-classic` is defined in the [webui repository](https://github.com/openhab/openhab-webui/blob/8b91b2336edddb5403a40816461dce5f18a1d49c/features/src/main/feature/feature.xml#L80). Every UI related feature is defined over there except for the new Main UI. Shouldn't we add the feature for Main UI over there too and remove it here?

kaikreuzer:
Afair, this had a specific reason - the MainUI is considered a mandatory part of the "standard" distro and not an add-on. If it's feature would reside in web-ui, it would not be part of the distro zip.

wborn:
Maybe it is better to move the `openhab-iconset-classic` feature to the distro feature file too then as it is also mandatory.

cweitkamp:
I guess that won't work as UI features depend on it too. Maybe we should move it to OHC?

cweitkamp:
Instead of having it as a dependency for `openhab-ui-runtime` we can add `openhab-iconset-classic` to the list of `bootFeatures` in the [distribution pom](https://github.com/openhab/openhab-distro/blob/85b277c73dc917022cd693f6c835f7586cd89c8f/distributions/openhab/pom.xml#L135).

kaikreuzer:
The bootFeatures do not feel as the right place, as we rather have "higher level" features listed in there.

What about handling it similar to the Main UI and include the [bundle directly](https://github.com/openhab/openhab-distro/pull/1371/files#diff-2598a1c3937934ca54a9e27154fb3294b2c16a642f01a7f35d8f2b90048280f0R8) instead of referencing the feature?

cweitkamp:
Okay. I think we can do that. In the same step we have to remove it from [openhab-webui feature file](https://github.com/openhab/openhab-webui/blob/f2f76419131eaf46b0c60df63fd0ba285fa60ea9/features/src/main/feature/feature.xml#L80-l83) and referencing bundles over there.

cweitkamp:
What do you think about https://github.com/openhab/openhab-distro/pull/1371/commits/66687acb8f7042517742185472e1f0ee2244bd37?

kaikreuzer (approved):
Lgtm. Will you also create a PR on webui to remove the iconset feature from there?

wborn (approved):
Thanks!

PR #1376, opened 2022-03-12
Remove BACNet Binding from Oomph setup

wborn:
This binding is not compatible with OH3 and it has no main branch so if you select it the Oomph setup fails with:

```
java.lang.Exception: org.eclipse.jgit.api.errors.RefNotFoundException: Ref refs/remotes/origin/main cannot be resolved
  at org.eclipse.oomph.setup.git.impl.GitCloneTaskImpl.perform(GitCloneTaskImpl.java:974)
```

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2537,35 +2537,7 @@
     <stream
         name="main"/>
   </project>
-  <project name="6_bacnet"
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
Thanks!

PR #1380, opened 2022-03-18
Add update instructions for removed MQTT System Broker

wborn:
* Instruct users to replace their things
* Remove mqttbroker.config

See also:

* openhab/openhab-addons#12157
* openhab/openhab-core#2842


cweitkamp (approved):
LGTM. Thanks.

PR #1398, opened 2022-06-18
Suppress warnings from CXF on startup

J-N-K:
Fixes https://github.com/openhab/openhab-core/issues/2998

Signed-off-by: Jan N. Klug <github@klug.nrw>


kaikreuzer (approved):
lgtm, thanks!

wborn:
Note that these changes will not have been applied if you upgrade from 3.2.0 to 3.3.0 because `update.lst` was not updated like [this](https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/bin/update.lst#L87).

If you upgrade from an older version like 3.1.0 they will be applied because it will then replace it as part of the 3.2.0 upgrade instructions.

It's also missing from [app/runtime/log4j2.xml](https://github.com/openhab/openhab-distro/blob/main/launch/app/runtime/log4j2.xml) so you'll keep running into these while debugging in Eclipse.

PR #1408, opened 2022-07-16
Add move instruction for transformation database

J-N-K:
Related to https://github.com/openhab/openhab-core/pull/3036

Although it is not very likely that someone already used it prior to the refactoring this will not hurt and save some trouble if someone did.

Signed-off-by: Jan N. Klug <github@klug.nrw>


wborn (approved):
Thanks!

Issue #1423, opened 2022-11-27
Eclipse installation fails due to incompatible plug-ins


wborn:
The Eclipse Oomph installer fails for newer Eclipse versions like 2022-12 because some Eclipse plug-ins are no longer compatible:

# Broken plug-ins

## :x: Mylyn (`org.eclipse.mylyn.*.feature.group` requirements)

**Since:** Eclipse 2022-06

**Logging:**

```
  ERROR: org.eclipse.equinox.p2.director code=0 Software being installed: artificial_root 1.0.0.v1669561710648
  ERROR: org.eclipse.equinox.p2.director code=0 Missing requirement: artificial_root 1.0.0.v1669561710648 requires 'org.eclipse.equinox.p2.iu; org.eclipse.mylyn.java_feature.feature.group 0.0.0' but it could not be found
```

Looks like this plug-in is no longer part of the Eclipse release train and is being restructured.

**See also:** [Mylyn not working with eclipse 2022-06](https://bugs.eclipse.org/bugs/show_bug.cgi?id=580257)

Most likely we can just remove these requirements.

---

## :x: M2E Sourcelookup (`bjmi.m2e.sourcelookup.feature.feature.group` requirement)

**Since:** Eclipse 2022-09

**Logging:**

```
  ERROR: org.eclipse.equinox.p2.director code=0 Software being installed: artificial_root 1.0.0.v1669561905048
  ERROR: org.eclipse.equinox.p2.director code=0 Missing requirement: M2E Sourcelookup 1.1.6.v20170717-044847 (bjmi.m2e.sourcelookup 1.1.6.v20170717-044847) requires 'osgi.bundle; org.eclipse.m2e.core [1.1.0,2.0.0)' but it could not be found
  ERROR: org.eclipse.equinox.p2.director code=1 Cannot satisfy dependency:
    ERROR: org.eclipse.equinox.p2.director code=0 From: artificial_root 1.0.0.v1669561905048
    ERROR: org.eclipse.equinox.p2.director code=0 To: org.eclipse.equinox.p2.iu; bjmi.m2e.sourcelookup.feature.feature.group 0.0.0
  ERROR: org.eclipse.equinox.p2.director code=1 Cannot satisfy dependency:
    ERROR: org.eclipse.equinox.p2.director code=0 From: M2E Sourcelookup 1.1.6.v20170717-044847 (bjmi.m2e.sourcelookup.feature.feature.group 1.1.6.v20170717-044847)
[...]

PR #1424, opened 2022-11-28
Removed mylyn from eclipse setup script to make installation setup working again

Hilbrand:
Since mylyn was [removed from eclipse](https://www.eclipse.org/lists/mylyn-pmc/msg00402.html) this causes issue with the installer setup script. This changes removes the requirements of mylyn from the setup script and other requirements that depended on mylyn.

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2354,27 +2354,11 @@
       <requirement
           name="org.eclipse.m2e.feature.feature.group"/>
       <requirement
-          name="org.eclipse.mylyn.java_feature.feature.group"/>
-      <requirement
-          name="org.eclipse.mylyn.bugzilla_feature.feature.group"/>
-      <requirement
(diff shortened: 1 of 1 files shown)


wborn:
> other requirements that depended on mylyn

These plug-ins don't depend on Mylyn, they just have their own incompatbility issues, see #1423.

It will probably take some time before all these plug-ins will be fixed. :disappointed: 

Removing only the Mylyn plug-ins will only allow for using 2022-06 but no newer Eclipse versions (see the "Since:" tags I added to #1423 today).

I still am not sure what the best way forward is so that's why I didn't create a PR like this. The m2e buildhelper plugin and lastnpe plug-ins are very useful which may outweigh being able to run the latest Eclipse version. So let's see what the other maintainers think.

I also have a review comment. :slightly_smiling_face:  When the plug-ins are removed the respective `<repository>` content should also be removed.

PR #1428, opened 2022-12-04
Add breaking change warning for JS Scripting

florian-h05:
Reference https://github.com/openhab/openhab-js/pull/175.

Although I can not imagine that this change breaks a user script, as most users would have parsed the string to a number, we should add a warning. I also changed the name from JS Scripting Automation to JavaScript Automation since we renamed the addon.

Signed-off-by: Florian Hotze <florianh_dev@icloud.com>


wborn (approved):
Thanks!

PR #1437, opened 2022-12-11
GitHub Actions workflow upgrades

wborn:
* Upgrade actions to fix deprecation warnings
* Use ubuntu-22.04 for builds


kaikreuzer (approved):

wborn:
I'll also update the PR rules so it will require the ubuntu-22.04 builds instead after the merge.

PR #1432, opened 2022-12-09
Upgrade minimum JDK to 17

J-N-K:
Related to https://github.com/openhab/openhab-core/pull/2994

 
Signed-off-by: Jan N. Klug <github@klug.nrw>

PR #1444, opened 2022-12-20
Use "OpenJDK 17" as JDK in pipeline

wborn:
Related to #1432

### DIFF
--- a/launch/Jenkinsfile
+++ b/launch/Jenkinsfile
@@ -63,7 +63,7 @@ def releaseOpenHabComponent(componentName, branch, releaseVersion, nextVersion, 
     resetOpenhabFork(gitBaseUrl, componentName, releaseVersion)
 
     withCredentials([usernamePassword(credentialsId: env.GIT_CREDENTIALS_ID, passwordVariable: 'githubPassword', usernameVariable: 'githubUser')]) {
-        withMaven(jdk: 'OpenJDK 11', maven: 'maven (latest)', mavenOpts: '-Xms512m -Xmx2048m', mavenLocalRepo: '.repository', globalMavenSettingsConfig: env.MAVEN_GLOBAL_SETTINGS, options: [artifactsPublisher(disabled: true)]) {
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):

wborn:
Apparently the unleash-maven-plugin does not like Java 17:

`[ERROR] Failed to execute goal com.itemis.maven.plugins:unleash-maven-plugin:2.9.3:perform (default-cli) on project org.openhab.core.reactor: Execution default-cli of goal com.itemis.maven.plugins:unleash-maven-plugin:2.9.3:perform failed: WELD-001524: Unable to load proxy class for bean Implicit Bean [javax.enterprise.event.Event] with qualifiers [@Default] with class interface javax.enterprise.event.Event using classloader ClassRealm[plugin>com.itemis.maven.plugins:unleash-maven-plugin:2.9.3, parent: jdk.internal.loader.ClassLoaders$AppClassLoader@5cb0d902]: Could not initialize class org.jboss.classfilewriter.ClassFile -> [Help 1]`

Now it doesn't help that it is no longer maintained:

https://github.com/shillner/unleash-maven-plugin

> This plugin provides a generic alternative to the error-prone default release plugin provided by Maven.

If there is no easy fix, the default release plugin could be a good long term alternative. 😉

kaikreuzer:
Hm, that's a pity.
@pfink I guess your expertise would help here: Since you had introduced the unleash plugin for some good reason, could you comment on what issues we might come across when dropping it and trying our luck with the release plugin instead?

pfink:
> Hm, that's a pity.
> 
> @pfink I guess your expertise would help here: Since you had introduced the unleash plugin for some good reason, could you comment on what issues we might come across when dropping it and trying our luck with the release plugin instead?

Hard to say, I think we would have to try to find that out. I could have a look into it between Christmas and Silvester.

wborn:
To get it fixed in the `unleash-maven-plugin` we'd have to fork it as well as its [maven-cdi-plugin-utils](https://github.com/shillner/maven-cdi-plugin-utils) dependency and then upgrade Weld to at least 3.1.1, see:

* https://issues.redhat.com/browse/WELD-2572
* http://weld.cdi-spec.org/news/2019/05/07/weld-311Final/

wborn:
I've just created issue #1445 so we can track this.

PR #1460, opened 2023-01-13
Enable GZIP compression for Jetty webserver

J-N-K:
Fixes https://github.com/openhab/openhab-core/issues/3313

### DIFF
--- a/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
+++ b/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
@@ -57,7 +57,13 @@
 			</Arg>
 		</Call>
 	</Get>
-
+	<Call name="insertHandler">
+		<Arg>
+			<New id="GzipHandler" class="org.eclipse.jetty.server.handler.gzip.GzipHandler">
+				<Set name="minGzipSize">1024</Set>
+			</New>
+		</Arg>
+	</Call>
 
 	<New id="httpConfig" class="org.eclipse.jetty.server.HttpConfiguration">
 		<Set name="secureScheme">https</Set>
--- a/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
+++ b/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
@@ -61,6 +61,11 @@
 		<Arg>
 			<New id="GzipHandler" class="org.eclipse.jetty.server.handler.gzip.GzipHandler">
 				<Set name="minGzipSize">1024</Set>
+				<Set name="includedPaths">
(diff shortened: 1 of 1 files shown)


ccutrer:
This doesn't seem to be working for me. I applied the diff to `/usr/share/openhab/runtime/etc/jetty.xml` on my openHAB 3.4.1 installation on an Ubuntu system, then restarted openHAB. I can see my browser making requests that include `Accept-Encoding: gzip, deflate, br`, but the response is still not compressed:

```
HTTP/1.1 200 OK
Date: Wed, 18 Jan 2023 15:28:49 GMT
Transfer-Encoding: Identity
Content-Type: application/json
Connection: keep-alive
Set-Cookie: X-OPENHAB-AUTH-HEADER=true
```

J-N-K:
For something that is larger than 1 kB? like the things endpoint?

ccutrer:
> For something that is larger than 1 kB? like the things endpoint?

Sorry, I edited my original response. And yes, 2.6MB worth of data transferred. Or 9.34MB on `/rest/channel-types?prefixes=system,mqtt` (MQTT is creating a unique channel type for _every_ channel, even though I have hundreds of repetition; I'm looking into it soon).

J-N-K:
I tried it again and it works fine for me when requesting /rest/addons using Safari:

Bildschirm­foto 2023-01-27 um 17 50 45

J-N-K:
Following up to a comment by @ghys I enabled compression only for REST resources to not interfere with other compression and configuration by add-on/UI services.

kaikreuzer (approved):

lolodomo:
As this change seems to "corrupt" the SSE communications between the server and BasicUI, I am asking myself if the client has the ability to request not using compression? 

I am also wondering if the remote openHAB binding was or not broken by this change. I have no way to test that with my current dev/prod environments.

lolodomo:
How does it work? Should the client implement the unzipping??

J-N-K:
That should be done automatically. Can you try 


```
<Call name="insertHandler">
		<Arg>
			<New id="GzipHandler" class="org.eclipse.jetty.server.handler.gzip.GzipHandler">
				<Set name="minGzipSize">1024</Set>
				<Set name="includedPaths">
					<Array type="java.lang.String">
						<Item>/rest/*</Item>
					</Array>
				</Set>
				<Set name="excludedPaths">
					<Array type="java.lang.String">
						<Item>/rest/events/*</Item>
						<Item>/rest/sitemaps/events/*</Item>
					</Array>
				</Set>
			</New>
		</Arg>
	</Call>
```

It seems to work for me.

PR #1465, opened 2023-01-23
Install JS Scripting on openhab-demo server

florian-h05:
With the Blockly port to JS Scripting and the removal of NashornJS from core, we need a JS ScriptEngine in openHAB 4.0 to make Blockly work.

This installs JS Scripting on the openHAB demo server and adds the already installed addons to the config file.

See https://github.com/openhab/openhab-distro/pull/1455#issuecomment-1368064700 for previous discussion.

/cc @kaikreuzer


wborn:
I don't think the demo code in this repo is actively used for https://demo.openhab.org/
But it would be nice if all the data used there is also part of the demo archives generated here.
Maybe @ghys can provide the demo server config?

PR #1472, opened 2023-02-26
Use Eclipse Add-on Service

wborn:
Depends on:

* openhab/openhab-core#3406
* openhab/openhab-webui#1736


kaikreuzer (approved):
That's very useful, thank you!

PR #1476, opened 2023-03-05
Fix SSE streams when using GZIP

J-N-K:
Fixes https://github.com/openhab/openhab-webui/issues/1757

It seems that enabling GZIP for the SSE endpoints breaks the ecent stream.

### DIFF
--- a/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
+++ b/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
@@ -66,6 +66,12 @@
 						<Item>/rest/*</Item>
 					</Array>
 				</Set>
+				<Set name="excludedPaths">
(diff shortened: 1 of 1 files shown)


wborn:
Did you also consider to only include the mime types that are known to work by calling [addIncludedMimeTypes](https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/server/handler/gzip/GzipHandler.html#addIncludedMimeTypes(java.lang.String...))? If that works we won't have to manage the list of endpoints.

J-N-K:
That also works.

wborn (approved):
Thanks! :+1:

lolodomo:
I just tested with this version of jetty.xml and it also works again for me.
Thank you.

PR #1477, opened 2023-03-07
Add breaking change alert for ISM8

lsiepel:
add breaking change alert for ISM8

related to: https://github.com/openhab/openhab-addons/pull/14206


wborn (changes requested):
It should be moved to the right OH version when it gets merged.

lsiepel:
Tried to fix the DCO with commit amend, somehow i failed. Anyway, i guess it should not be an issue as it only is a commit to change the line+ version numbner.

wborn (changes requested):
As there is a 4.1.0 release now, it again should be moved to the right OH version when it gets merged.

lsiepel:
> As there is a 4.1.0 release now, it again should be moved to the right OH version when it gets merged.

Done.

kaikreuzer (approved):

PR #1479, opened 2023-03-10
[Netatmo] Alert about changes

clinique:
No breaking changes, just informations, but important.


clinique:
Related to #14548  and #14543

PR #1516, opened 2023-04-04
Add update warning for changed rule triggering

J-N-K:
See https://github.com/openhab/openhab-core/pull/3141


wborn (approved):
Thanks for the update!

PR #1518, opened 2023-04-09
Added breaking change for icalendar thing-split

daMihe:
This PR depends on https://github.com/openhab/openhab-addons/pull/14766.

It adds a message for the breaking change resulting from splitting the bridge into smaller things.


wborn (changes requested):
It should be moved to the right OH version when it gets merged.

daMihe:
Done. Added 4.1.0 section which may collide. If required i'll update again. I'm waiting for a review in the referenced PR.

wborn (changes requested):
As there is a 4.1.0 release now, it again should be moved to the right OH version when it gets merged.

daMihe:
Okay, i'll do when the code gets merged.

wborn:
Let's close this as https://github.com/openhab/openhab-addons/pull/14766 was also closed.

PR #1524, opened 2023-04-15
Improve addons.cfg documentation

### LINKED ISSUE #1503: Clarify doc in addons.cfg as to when it gets loaded, and that it replaces any UI-based addons
At the top of `addons.cfg` it says “This is only regarded at the VERY FIRST START of openHAB." (Sorry if that file doesn't live in this repo, I [couldn't find](https://github.com/search?q=org%3Aopenhab+addons.cfg&type=code) where it comes from.)

From that description, I would expect that the first launch of openHAB would read `addons.cfg`, and then any subsequent service restart, even after host restart or cache clear, would not even read the file.

However, my experience is that I installed some addons via UI, and then they got uninstalled again. (I had restarted the openHAB service and also cleared the cache while debugging things). And [on the forums](https://community.openhab.org/t/oh3-transformation-service-config-for-js-and-map-in-a-thing-channel-what-worked-qs-about-what-didnt/145507/3) someone advised me that that's expected / not to mix `addons.cfg` and UI installs.

So, it would be great to clarify the file comment. It's a great place to explain the behavior, I definitely saw the warning and tried to hede it.

wborn:
Fixes #1503

### DIFF
--- a/distributions/openhab-demo/src/main/resources/conf/services/addons.cfg
+++ b/distributions/openhab-demo/src/main/resources/conf/services/addons.cfg
@@ -1,11 +1,12 @@
 # The installation package of this openHAB instance
-# Note: This is only regarded at the VERY FIRST START of openHAB
-# Note: If you want to specify your add-ons yourself through entries below, set the package to "minimal"
-# as otherwise your definition might be in conflict with what the installation package defines.
 #
 # Valid options:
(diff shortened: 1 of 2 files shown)


J-N-K (approved):
LGTM, thanks. This is much clearer than before

PR #1528, opened 2023-05-02
Add upgradetool to 'runtime/bin' directory

J-N-K:
Related to https://github.com/openhab/openhab-core/pull/3268

This makes the upgrade tool available in the distributions.

Maybe some shell wizard can add script that detect the correct userdata directory, unfortunately I can't :-)

### DIFF
--- a/distributions/openhab/pom.xml
+++ b/distributions/openhab/pom.xml
@@ -88,7 +88,36 @@
                 </includes>
             </resource>
         </resources>
+
         <plugins>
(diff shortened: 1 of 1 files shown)


J-N-K:
Build fails because tonights snapshot did not succeed due to a test failure and the upgradetool is not available. I'll re-trigger the build tomorrow.

kaikreuzer (approved):

Issue #1481, opened 2023-03-12
Udapte to openhab 4.0 Milestone Build does bot remove JavaSript transformation add-on


kristofejro:
After upgrade from latest 3.4 Milestone Build to 4.0.0.M1 Milestone Build I have some warnings in log during the startup:

```
2023-03-12 06:45:04.811 [WARN ] [core.karaf.internal.FeatureInstaller] - The transformation add-on 'javascript' does not exist - ignoring it.
2023-03-12 06:46:04.798 [WARN ] [core.karaf.internal.FeatureInstaller] - The transformation add-on 'javascript' does not exist - ignoring it.
```

In MainUI in Transformation Add-ons section there is no JavaSript transformation add-on listed but when going to karaf console I can see javascript transformation listed:

```
----------------------------------------------------------------
Pid:            org.openhab.addons
BundleLocation: mvn:org.openhab.core.bundles/org.openhab.core.karaf/4.0.0.M1
Properties:
   automation =
   binding = astro,ipcamera,mail,mqtt,icalendar,http,ntp,networkupstools,zwave,exec,network,openweathermap
   legacy = false
   misc = ,metrics
   package = standard
   persistence = mapdb,influxdb
   remote = true
   service.pid = org.openhab.addons
   transformation = map,jsonpath,regex,javascript
   ui = basic,habpanel
----------------------------------------------------------------
```

As @J-N-K [advised ](https://community.openhab.org/t/openhab-4-0-milestone-discussion/145133/33?u=kristofejro)I am opening this issue for future fix.

[Fix](https://community.openhab.org/t/solved-oh3-how-to-get-rid-of-old-bindings/111049/6?u=kristofejro)

If JavaSript transformation add-on is not present in OH4 so it should be removed automatically.

wborn:
If you skip OH3 completely and upgrade from OH2 to OH4 there is also https://github.com/openhab/openhab-distro/issues/1172.

lolodomo:
Imperihome also removed: #15014

PR #1530, opened 2023-05-04
Remove obsolete addons from managed configuration

J-N-K:
It has been reported that the removal of the javascript transformation in openHAB 4 results in warnings because the add-on could not be found during installation. Since the add-on is no longer present, it does not show up in the UI and can't be uninstalled from there. This adds an update instruction to remove it from the addons.config.

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -215,3 +215,15 @@ MOVE;$OPENHAB_USERDATA/jsondb/org.openhab.core.transform.TransformationConfigura
 REPLACE;,javascript;;$OPENHAB_USERDATA/config/org/openhab/addons.config
 REPLACE;javascript,;;$OPENHAB_USERDATA/config/org/openhab/addons.config
 REPLACE;javascript;;$OPENHAB_USERDATA/config/org/openhab/addons.config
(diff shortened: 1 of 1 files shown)


wborn (commented):
Perhaps the other add-ons (https://github.com/openhab/openhab-distro/issues/1481#issuecomment-1465165450) can also be removed similarly?

J-N-K:
Yes. Do you see any better solution than these three lines? The problem is that we can only do simple "replace" instructions here and have the following cases:

- `binding = b1,remove,b3` -> matched by 1 and 2, 3 would result in `binding = b1,,b3` which is invalid
- `binding = b1,b2,remove` -> matched by 2 only, 3 would result in `binding = b1,b2,` which is invalid
- `binding = remove,b2,b3` -> matched by 1 only, 3 would result in `binding = ,b2,b3` which is invalid
- `binding = remove` -> only matched by 3

wborn:
> Do you see any better solution than these three lines?

The only better solution I can think of is adding a dedicated update commands for removing or rename add-ons. Then you can just use one command per add-on. Though adding new commands is quite involved because they need to work on Linux/macOS/Windows and added to all repos used for packaging openHAB (distro, linuxpkg, docker, etc).

J-N-K:
I'll add commands for the other addons

wborn (approved):
Thanks!

PR #1532, opened 2023-05-10
Call upgrade tool automatically from within the update script

kaikreuzer:
This might need testing under Windows - I only tested the Unix/Mac version of the script.

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update
+++ b/distributions/openhab/src/main/resources/bin/update
@@ -253,6 +253,7 @@ SkipModifier="$3"
 
 ##Run the initialisation functions defined above
 setup "$SpecifiedVersion" "$SpecifiedDir"
+
 download "$SkipModifier"
 
 transferFile="update.lst"
@@ -352,6 +353,15 @@ rm -rf "${TempDir:?}"
 ## Restore file ownership.
 echo "Restoring previous file ownership ($FileOwner:$FileGroup)"
 chown -R "$FileOwner:$FileGroup" "$WorkingDir"
+
+## Start the upgrade tool
+echo "Starting JSON database update..."
+export OPENHAB_USERDATA="$OPENHAB_USERDATA"
+java -jar "$WorkingDir/runtime/bin/upgradetool.jar" || {
+  echo "Update tool failed, please check the openHAB website (www.openhab.org) for manual update instructions." >&2
+  exit 1
+}
+echo "JSON database updated successfully."
 
 echo ""
 echo "SUCCESS: openHAB updated from $CurrentVersion to $OHVersion"
--- a/distributions/openhab/src/main/resources/bin/update.ps1
+++ b/distributions/openhab/src/main/resources/bin/update.ps1
@@ -799,6 +799,14 @@ Function Update-openHAB() {
             return PrintAndThrow "Could not replace the $AddonsFile" $_
         }
 
+        # Now run the upgrade tool to update the JSON database
+        Write-Host "Starting JSON database update..."
+        java -jar "$($PSScriptRoot)\runtime\bin\upgradetool.jar"
+        if ($LASTEXITCODE -ne 0) {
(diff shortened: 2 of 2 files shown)


distributions/openhab/src/main/resources/bin/update
```diff
@@ -353,6 +354,15 @@ rm -rf "${TempDir:?}"
 echo "Restoring previous file ownership ($FileOwner:$FileGroup)"
 chown -R "$FileOwner:$FileGroup" "$WorkingDir"
 
+## Start the upgrade tool
+echo "Starting JSON database update..."
+export OPENHAB_USERDATA="$OPENHAB_USERDATA"
+java -jar "$WorkingDir/runtime/bin/upgradetool.jar" || {
```
J-N-K, line 360: Don't we need to add the commands to the command line? Or do you suggest we change the tool to always execute all steps? This is probably doable since we keep track of which steps have already run in the past.

kaikreuzer: Yes, this was my assumption - that we can call the tool as many times as we want as it itself is aware of what it already did and what is still needed to do. I think this is the only way for it to react correctly, since we wouldn't want to update the update script here whenever there are new steps in the tool available.

J-N-K:
@kaikreuzer can you merge https://github.com/openhab/openhab-core/pull/3607 to have a working upgrade? Thanks.

J-N-K:
Unfortunately this is not working as expected, because when the update is performed, the old update script is still used (at least for my local builds).

lolodomo:
> This is probably doable since we keep track of which steps have already run in the past.

Where is it stored? 
I would like to "erase" It when trying again the tool.

J-N-K:
In the jsondb directory in a JSON database

kaikreuzer:
> the old update script is still used

It is using the update script from the version that it downloaded. So "old" update script means in your case that it is the update script of the version that you want to upgrade to - which is imho exactly what we need.

J-N-K:
Interesting. I'll check that, maybe it's because I have a slightly modified script that deploys my local build.

BClark09:
Hi all,

I've just found this and the discussion over at https://github.com/openhab/openhab-docs/issues/2074. Please feel free to ping me if it's anything Linux installation related! :)

I like the sound of the an upgrade tool coming from openHAB. Since APT and YUM/DNF do their own type of update process, I'm assuming this java executable should always be run post installation? If so, I can make the relevent changes in openhab/linuxpkg which will also affect openHABian users.

An FYI - if you do decide to call the tool after a backup restore - APT/YUM/DNF/openHABian uses the same scripts as is provided in the distro so there's no change there.

J-N-K:
Does the linux package use the update script? If not, it should call the upgrade tool. If it does use the script, there is no need to do anything.

BClark09:
Great, thanks @J-N-K. It doesn't use the update script so I'll add it to the RPM and DEB packages.

Mark-VG:
> This might need testing under Windows - I only tested the Unix/Mac version of the script.

From my testing the update tool doe snot work when run from the Windows update script.
```
Starting JSON database update...
Error: Unable to access jarfile C:\Users\Mark\AppData\Local\Temp\openhab\update\runtime\bin\runtime\bin\upgradetool.jar
Update-openHAB : Update tool failed, please check the openHAB website (www.openhab.org) for manual update instructions.
At C:\openHAB\runtime\bin\update.ps1:584 char:22
+ ...        exit Update-openHAB -OHDirectory $OHDirectory -OHVersion $OHVe ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,Update-openHAB
```
Manually running the script does work.

PR #1542, opened 2023-05-22
Jetty `/static`: Set cache-control header & Enable gzip compression

florian-h05:
This sets the cache-control header for what is served under `/static` to enable storage caching (very useful when serving images for UIs) and enables GZIP compression.

### DIFF
--- a/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
+++ b/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
@@ -51,6 +51,7 @@
 						<New class="org.eclipse.jetty.server.handler.ResourceHandler">
 							<Set name="resourceBase"><SystemProperty name="openhab.conf" />/html</Set>
 							<Set name="directoriesListed">false</Set>
+							<Set name="cacheControl">max-age=31536000</Set>
 						</New>
 					</Set>
 				</New>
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
This might have some negative impact when people expect a live update whenever they do changes in their static folder, but I guess that's a very rare use case.

florian-h05:
Right, but since it's named static I wouldn't expect live updates.

And if one needs them: there are a few ways, e.g. to use a reverse proxy to rewrite headers or directly serve those files using a web server.

florian-h05:
@kaikreuzer One idea that I just had, is to add an additional resource handler for conf/html folder with the name `/dynamic` and `Cache-Control: nocache`, so the user can easily decided whether he wants caching or not. WDYT?

kaikreuzer:
No, I don't think that's necessary. The folder is meant for static files and the main use case would be that people maybe put new images there and want to use them as a background for their UI or so. It might hence make sense to exclude 404s from the `Cache-Control` header. Do you happen to know whether this is the default for Jetty?

florian-h05:
I don‘t know, but it seems that jetty does a good job!

When I request something that does not exist to get a 404:
image

When I request something that exists (200):
image

PR #1543, opened 2023-05-25
Remove standard package (default installation) as replaced by setup wizard

florian-h05:
https://github.com/openhab/openhab-webui/issues/1659 & https://github.com/openhab/openhab-webui/pull/1908 added recommended addons to the setup wizard.
These will get installed on a new openHAB install if the user does not actively de-select them, so the standard package (default installation) can be removed from distro.

See also #1455.

### DIFF
--- a/distributions/openhab/src/main/resources/conf/services/addons.cfg
+++ b/distributions/openhab/src/main/resources/conf/services/addons.cfg
@@ -1,7 +1,7 @@
 # The installation package of this openHAB instance
 #
 # Valid options:
-#   - standard : Standard setup for normal use of openHAB with persistence (rrd4j) and additional UI add-ons (basic,habpanel).
+#   - standard : Standard setup for normal use of openHAB with additional UI add-ons (basic,habpanel).
 #   - minimal  : Installation of core components without additional add-ons.
 #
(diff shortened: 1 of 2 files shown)


J-N-K:
I wonder if we should fully remove package support and add basic UI and HabPanel to the same setup wizard. This would also simplify code in core.

florian-h05:
That‘s a good idea, and if it helps core, let‘s do it.

Is is as simple as just putting `ui-basic` and `ui-habpanel` to the two other addons here:

https://github.com/openhab/openhab-webui/blob/b5176216c11bbdf2e41b95a00d3b21f3f1f07288/bundles/org.openhab.ui/web/src/pages/wizards/setup-wizard.vue#L232

kaikreuzer:
I would actually suggest to ONLY suggest Basic UI to be installed - HABPanel is imho by now a very optional UI that most users won't need.

florian-h05:
I removed the standard package, please let me know if I did it right or if I missed something.

J-N-K (approved):
LGTM. But https://github.com/openhab/openhab-core/pull/3634 needs to be merged before otherwise we might run into trouble.

kaikreuzer (approved):

PR #1517, opened 2023-04-04
Suppress ItemStateUpdatedEvent and GroupStateUpdatedEvent

rkoshak:
Signed-off-by: Richard Koshak <rlkoshak@gmail.com>

PR #1561, opened 2023-07-29
Make sure updated log4j2.xml configuration is used

wborn:
This update command makes sure that log4j2.xml contains the changes of #1398 and #1517.


J-N-K (commented):
Thanks! Does this also make sure it is used if someone upgrades from 4.0.0 to 4.1.0?

wborn:
AFAIK it does.

J-N-K (approved):

PR #1576, opened 2023-08-20
Reminder on how to increase webclient threads

hakan42:
Every so often, increasing the number of webclient threads may become necessary, especially on dockerized systems with higher number of cores. This is a reminder for people on how to perform this.

See:
- https://community.openhab.org/search?q=webclient%20threads
- https://community.openhab.org/t/system-settings-for-other-hardware-platforms-then-pi/76787
-  https://community.openhab.org/t/fix-for-jetty-error-when-running-on-host-with-many-cores/57449

Related to https://github.com/openhab/openhab-addons/pull/15442

### DIFF
--- a/distributions/openhab/src/main/resources/conf/services/runtime.cfg
+++ b/distributions/openhab/src/main/resources/conf/services/runtime.cfg
@@ -275,6 +275,8 @@
 #
 #org.openhab.inbox:autoApprove=false
 
+################################## WEBCLIENT ###################################
+
 # For installations on higher end systems with many cores, increasing number of webclient threads
 # may be necessary. Usually, this is reported with "Insufficient configured threads" in the logs.
 #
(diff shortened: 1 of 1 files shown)


distributions/openhab/src/main/resources/conf/services/runtime.cfg
```diff
@@ -275,3 +275,10 @@
 #
 #org.openhab.inbox:autoApprove=false
 
+# For installations on higher end systems with many cores, increasing number of webclient threads
```
wborn, line 278: How about also adding a header so it doesn't look like this is INBOX config?

```suggestion
################################## WEBCLIENT ###################################

# For installations on higher end systems with many cores, increasing number of webclient threads
```

wborn (approved):
LGTM :+1:

PR #1577, opened 2023-08-20
Exclude log format settings file from update

### LINKED ISSUE #1573: Exclude log format settings file from update
I propose to remove `org.apache.karaf.log.cfg` from `userdata_sysfiles.lst`. There is nothing we need to update on a regular basis and it always overwrites custom log format setting on upgrade. @openhab/distro-maintainers WDYT?

J-N-K:
Closes #1573


wborn (approved):
Thank you!

PR #1579, opened 2023-08-21
[audio] Reencode the doorbell.mp3 sample file

dalgwen:
Reencode doorbell.mp3 , as it cannot be played on the webaudio sink on firefox (but worked fine on chrome).
I also shortened it a bit (there was several seconds of nearly blank sound at the end of the file).

(from [this post](https://community.openhab.org/t/webaudio-not-working-when-trying-to-stream-webradio/148610/8?u=dalgwen))


wborn (approved):
Thanks for fixing the door bell!
Now people no longer have to knock on the door. :wink:

PR #1588, opened 2023-09-06
Blacklist Karaf related bundles in Demo App

wborn:
This will protect developers from themselves in case they try to add Karaf related bundles to the Demo App.

Inspired by this [community topic](https://community.openhab.org/t/eclipse-via-eclipse-installer-adding-binding-fails-with-org-apache-karaf-kar-core-version-4-4-3/149382?u=wborn)


J-N-K (approved):
Thanks

PR #1591, opened 2023-09-27
Upgrade to Karaf 4.4.4

wborn:
* Syncs distro customizations with Karaf 4.4.4
* Resolves app runbundles for the new dependencies
* Add specs features because Pax Web now depends on "asm"
* Use org.ops4j.pax.logging.DefaultServiceLog.level=ERROR in Demo App similar to when running Karaf to suppress "The use of package scanning to locate plugins is deprecated and will be removed in a future release" warnings due to LOG4J2-3644.

For release notes, see:

https://issues.apache.org/jira/secure/ReleaseNote.jspa?version=12352693&styleName=&projectId=12311140

---

Depends on https://github.com/openhab/openhab-core/pull/3814


J-N-K (approved):
Thanks!

PR #1608, opened 2023-11-21
Added breaking change alert for haywardomnilogic

matchews:
This depends on pull request #[15478](https://github.com/openhab/openhab-addons/pull/15478/files/a454fe9bb0ce03390ae4d832b2b96bc3760805e1#)


wborn (changes requested):
As there is a 4.1.0 release now, it should be moved to the right OH version when it gets merged.

lsiepel:
Seems no longer needed.

PR #1612, opened 2023-12-05
[hue] Add update note about light channels

jlaur:
Related to openhab/openhab-addons#15984


kaikreuzer:
Our thing migration framework does not work in this case?

jlaur:
> Our thing migration framework does not work in this case?

No, I tested it with both `update-channel` as well as `remove-channel` + `add-channel` without success. I had a quick look at the migration code, but it didn't seem obvious to me how to fix this. For now I mentioned it here: openhab/openhab-core#3660.

kaikreuzer (approved):
Alrght then, thanks!

PR #1620, opened 2023-12-20
Package addons.xml with the distro instead of as an external feature

kaikreuzer:


J-N-K:
If we do it like this, the file is not replaced when cache is cleared, correct? I'm not sure if it is necessary to update it in this case, because if you clear cache (on a snapshot), other versions of addons will be installed, which might need the updated addon.xml

kaikreuzer:
Correct. But I don't consider cache cleaning a valid operation anyhow, especially not on snapshots. There might also be a new feature in core that the newer snapshot add-ons rely on, while it isn't present in the old snapshot core.

J-N-K (approved):

J-N-K:
Look good. I assume you checked it works? I'm on the phone, so can't check that.

kaikreuzer:
I checked that the packaging works, yes, and I have moved the file now to `runtime/etc`.
Additionally, I have created https://github.com/openhab/openhab-core/pull/3944, so that the core is looking for the file at the new location.
Would be great, if both could be still reviewed/merged tonight, so that we have a new snapshot for tests tomorrow morning.

J-N-K (approved):

launch/app/app.bndrun
```diff
@@ -86,6 +86,7 @@ feature.openhab-model-runtime-all: \
 	org.apache.cxf.osgi.http.transport.disable=true,\
 	org.ops4j.pax.logging.DefaultServiceLog.level=ERROR,\
 	openhab.servicecfg=${.}/runtime/services.cfg,\
+    openhab.runtime=${.}/runtime/runtime,\
```
jlaur, line 89: Very minor, it seems the other lines start with a tab rather than four spaces.

andrewfg:
@kaikreuzer / @J-N-K / @mherwege we have now successfully merged the USB serial suggestion finder service (for suggesting bindings that depend on USB serial devices); and consequently we also have added the discovery-methods to the `addon.xml` files for the ZigBee and Z-Wave bindings. The last step therefore is to include those two individual `addon.xml` files into the OH combined `addons.xml` file. But I don't know how to do that. => Any suggestions?

wborn:
> But I don't know how to do that. => Any suggestions?

https://github.com/openhab/openhab-distro/pull/1626

PR #1626, opened 2024-01-04
Merge addon info of all repos into one addons.xml file

wborn:
Extracts the ZigBee and Z-Wave addon.xml files and merges them into the addons.xml file using Groovy.


distributions/openhab/merge-addon-info.groovy
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
```
jlaur, line 12: Is this license header needed, and if so, is there a configuration in this repository for automatically updating copyright year in license headers?

wborn: I added it because we also have it on our other [Groovy file](https://github.com/openhab/openhab-core/blob/7bac24e58f057ddb739194a7fe370ab6c42758c0/bundles/org.openhab.core.semantics/model/generateTagClasses.groovy#L1-L12).

There is no license-maven-plugin config currently but we can add it before 2025 when the header survives the merge request.

holgerfriedrich:
Just tried it out, addons.xml now contains zwave an zway. It is in the distro zip and accepted by he addon finders. 👍

holgerfriedrich:
> Just tried it out, addons.xml now contains zwave an zway. It is in the distro zip and accepted by the addon finders. 👍

@wborn Bad news. Unfortunately there seem to be issues, at least when building on Windows..... Seems I have not seen this in the first tryout because I touched the new addons.xml and VS Code fixed it when saving....

We seem to have mixed line endings:

image

The XML parser fails.
image


(When I fix this, I seen to break the encoding in the description of nobohub binding, so I removed the UTF8 characters).
After changing both, it works.

wborn:
Does it work better with the new commit?

distributions/openhab/merge-addon-info.groovy
wborn: I created #1627 for this

wborn:
It seems to always [print the XML using Unix line endings](https://github.com/groovy/groovy-core/blob/01309f9d4be34ddf93c4a9943b5a97843bff6181/src/main/groovy/util/IndentPrinter.java#L185-L201) so I've updated the script for that.

holgerfriedrich:
Almost fine, but not yet. Still the parser is unable to handle it. Hope this is not caused by some weird stuff on my machine.

Freshly generated distro zip: fail. Opening ```addons.xml``` with VS Code, saving it. Then it works.
The diff shows some changes related to unicode characters:
image

Not working:
image

After saving:
image

wborn:
With the new commit it seems to work better on my Windows VM.

holgerfriedrich:
Does not work for me - same output as before.
Following up on the screenshot above:
https://www.compart.com/de/unicode/U+00F8, encoded as ``c3b8``` in UTF-8.

This character is correct in ./distributions/openhab/target/addon-xml/addons.xml, but broken in ./distributions/openhab/target/assembly/runtime/etc/addons.xml.

wborn:
How does this BufferedWriter work for you? :slightly_smiling_face:
According to the docs that it will always use UTF-8.

holgerfriedrich:
Yeah, that did the trick! 🚀 👍

holgerfriedrich (approved):

wborn:
Great! 😅 If it works on Linux and Windows I cannot imagine it will not work macOS. :wink:

J-N-K (approved):
LGTM, thanks!

PR #1627, opened 2024-01-05
Improve Maven tooling

wborn:
* Add license, sortpom and spotless Maven plugin configurations similar to other repos
* Add license files
* Sort POMs and apply Spotless

Depends on #1626


J-N-K (approved):
LGTM

PR #1628, opened 2024-01-05
Clear marketplace cache during upgrade

J-N-K:
Related to https://github.com/openhab/openhab-core/pull/4019


wborn (approved):
LGTM! I'll also add this to the Docker update scripts.

jimtng:
I wonder if this is causing my smarthome/j marketplace addon to disappear after every snapshot update (docker). It wasn't reinstalled either. I'm not sure where to start looking. As a workaround I moved the .kar file from marketplace/ to addons/.

J-N-K:
That's why they are removed, but they should be re-installed (and in fact that happens on my system). Can you try if you can re-produce the issue without using docker?

PR #1632, opened 2024-01-30
[mqtt.generic] add breaking change alert for rollershutter stop config

ccutrer:
See also https://github.com/openhab/openhab-addons/pull/16345. This behavior was accidentally introduced in 4.1.0, and made clearer in 4.2.


wborn (approved):
Thanks!

PR #1633, opened 2024-02-02
Add suggestion finder parameters to addons.cfg

mherwege:
Some of the add-on suggestion finders can be disable through the UI, or parameters in the addons.cfg file.

In this PR, these parameters are added (commented) in the configuration file.

### DIFF
--- a/distributions/openhab/src/main/resources/conf/services/addons.cfg
+++ b/distributions/openhab/src/main/resources/conf/services/addons.cfg
@@ -8,6 +8,14 @@
 #
 #includeIncompatible = false
 
+# The system can automatically suggest add-ons for installation using add-on finders.
+# If you don't want add-on suggestions from specific finders, you can disable the respective finders.
+# They are enabled by default.
+
+#suggestionFinderIp=true
+#suggestionFinderMdns=true
+#suggestionFinderUpnp=true
+
 # The add-on configuration in the lists below is applied EVERY TIME openHAB is started.
 # Add-ons installed using the UI that do not occur in the lists will be uninstalled each startup.
 # When lists are commented again any add-ons in the list remain installed and are not removed.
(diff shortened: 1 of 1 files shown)


andrewfg:
@mherwege to be correct, the file name is 'addon.config' (rather than 'addon.cfg')..

mherwege:
> @mherwege to be correct, the file name is 'addon.config' (rather than 'addon.cfg')..

Please check again, it is `addon.cfg` in the `services` directory.

andrewfg:
Apparently the raw 'addon.cfg' in services is auto-magically transformed into 'addon.config' in userdata at run-time. ??

`//OPENHABIAN/openHAB-conf/services/addon.cfg` => `//OPENHABIAN/openHAB-userdata/config/org/openhab/addon.config`

andrewfg:
> a bit more complex than that, because userdata/config/org/openhab/addon.config will also contain the UI configured parameters

Yeah. What I observed was that if you change a finder config via the UI then addon.config is updated. And I think the other way round if you change addon.config manually then the UI reflects that. ?? But presumably if you change addon.cfg that overrides addon.config and makes the parameter effectively read only.

distributions/openhab/src/main/resources/conf/services/addons.cfg
```diff
@@ -8,6 +8,14 @@
 #
 #includeIncompatible = false
 
+# The system can automatically suggest add-ons for installation using add-on finders.
+# If you don't want add-on suggestions from specific finders, you can disable the respective finders.
+# They are enabled by default.
+
+#suggestionFinderIp=true
+#suggestionFinderMdns=true
+#suggestionFinderUpnp=true
```
wborn, line 17:
```suggestion
#
#suggestionFinderIp = true
#suggestionFinderMdns = true
#suggestionFinderUpnp = true
```

wborn (approved):
Thanks!

PR #1644, opened 2024-04-06
Added breaking change alert for evcc

MikeTheTux:
based on:

- https://github.com/openhab/openhab-addons/pull/16114
- https://github.com/openhab/openhab-addons/pull/16428

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.lst
+++ b/distributions/openhab/src/main/resources/bin/update.lst
@@ -132,8 +132,8 @@ ALERT;Twitter Binding has become X. All related references to Twitter and Tweets
 
 [4.2.0]
 ALERT;AndroidTV Binding: The thing configuration of 'port' has been renamed and split to 'googletvPort' and 'shieldtvPort'
-ALERT;ISM8 Binding: Most channels have changed and are now using Units of Measurements. Items must be adapted and the things must be recreated.
 ALERT;evcc Binding: Update to evcc API version 0.123.1 results in several new or updated channels. Existing Items will need to be adjusted.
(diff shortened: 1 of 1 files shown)


distributions/openhab/src/main/resources/bin/update.lst
```diff
@@ -133,6 +133,7 @@ ALERT;Twitter Binding has become X. All related references to Twitter and Tweets
 [4.2.0]
 ALERT;AndroidTV Binding: The thing configuration of 'port' has been renamed and split to 'googletvPort' and 'shieldtvPort'
 ALERT;ISM8 Binding: Most channels have changed and are now using Units of Measurements. Items must be adapted and the things must be recreated.
+ALERT;evcc Binding: Update to evcc API version 0.123.1 results in several new or updated channels. Existing Items will need to be adjusted.
```
holgerfriedrich, line 136:
```suggestion
ALERT;evcc Binding: Update to evcc API version 0.123.1 results in several new or updated channels. Existing Items will need to be adjusted.
ALERT;ISM8 Binding: Most channels have changed and are now using Units of Measurements. Items must be adapted and the things must be recreated.
```

holgerfriedrich (commented):
I think it should be in alphabetic order (not case sensitive).

distributions/openhab/src/main/resources/bin/update.lst
MikeTheTux: fixed

holgerfriedrich (approved):
LGTM

wborn (approved):
Thanks both!

Issue #1590, opened 2023-09-19
Add support for running openHAB with Java 21


wborn:
Today Java 21 will be launched. :rocket:  This is a LTS release, so we should also support using it with openHAB.

We can use this issue to keep track of what still needs to be fixed or investigated (similar to #1316). 

There's also a launch live stream: https://dev.java/community/java-21-launch/

TODO:

* [x] Upgrade Netty to 4.1.99 or newer to prevent JVM crashes (see [4.1.99 release notes](https://netty.io/news/2023/09/21/4-1-99-Final.html)) (https://github.com/openhab/openhab-core/pull/3824)
* [x] Update Java compatibility check in `karaf` and `karaf.bat`

wborn:
It seems that official Java 21 are delayed (see [Temurin blog](https://adoptium.net/blog/2023/09/temurin21-delay/)).

Nevertheless I was able to use the Temurin 21-beta+34-202308082331 nightly build and got OH 4.1.0-SNAPSHOT running without issues so far after disabling the recently added Java compatibility check.

It seems the recent Karaf 4.4.4 upgrade helped a lot because OH 4.0.3 does not start with Java 21:

```
2023-10-02 11:59:11.900 [SEVERE] [org.apache.karaf.main.Main] - Could not launch framework
java.lang.IllegalStateException: Error initializing storage for Equinox container.
	at org.eclipse.osgi.internal.framework.EquinoxContainer.<init>(EquinoxContainer.java:110)
	at org.eclipse.osgi.launch.Equinox.<init>(Equinox.java:53)
	at org.eclipse.osgi.launch.EquinoxFactory.newFramework(EquinoxFactory.java:35)
	at org.eclipse.osgi.launch.EquinoxFactory.newFramework(EquinoxFactory.java:30)
	at org.apache.karaf.main.Main.launch(Main.java:291)
	at org.apache.karaf.main.Main.main(Main.java:183)
Caused by: java.lang.RuntimeException: Error occurred while checking the system module.
	at org.eclipse.osgi.storage.Storage.checkSystemBundle(Storage.java:467)
	at org.eclipse.osgi.storage.Storage.createStorage(Storage.java:187)
	at org.eclipse.osgi.internal.framework.EquinoxContainer.<init>(EquinoxContainer.java:108)
	... 5 more
[...]

PR #1655, opened 2024-05-14
[CI] Add Java 21 compiling to J17 and native J21

holgerfriedrich:
As we have implemented all changes that allow for Java 21 builds, I'd like to introduce it in the CI as well.
For core and addons, the same approach was already implemented quite a while ago.

Refs: #1590


kaikreuzer (approved):
Thanks!

PR #1656, opened 2024-05-15
Use RollingRandomAccessFile appender with openhab.log

### LINKED ISSUE #1654: Logging stops when logfile goes away on zram sync
Sorry upfront for calling this a bug, it's merely a feature request or something inbetween,
but it has severe impact on useability for people to use openHABian (which there are quite a lot).

openHABian now by default syncs ZRAM to disk, un- and remounting the filesystem the /var/log/openhab is on.
openHAB keeps running but that seems to result in logging to stop, probably because the file handle is gone.
Now to issue a log:set command seems to recover logging but unforunately this is difficult to initiate from the outside.

Could you add some sort of periodic logfile handle or logger reinit check that does what issueing `log:set` does ?
Thanks!

See also https://community.openhab.org/t/openhabian-testing/147079/121

wborn:
Closes #1654

### DIFF
--- a/distributions/openhab/src/main/resources/userdata/etc/log4j2.xml
+++ b/distributions/openhab/src/main/resources/userdata/etc/log4j2.xml
@@ -7,14 +7,14 @@
 		</Console>
 
 		<!-- Rolling file appender -->
-		<RollingFile fileName="${sys:openhab.logdir}/openhab.log" filePattern="${sys:openhab.logdir}/openhab.log.%i.gz" name="LOGFILE">
+		<RollingRandomAccessFile fileName="${sys:openhab.logdir}/openhab.log" filePattern="${sys:openhab.logdir}/openhab.log.%i.gz" name="LOGFILE">
 			<PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSS} [%-5.5p] [%-36.36c] - %m%n"/>
 			<Policies>
 				<OnStartupTriggeringPolicy/>
 				<SizeBasedTriggeringPolicy size="16 MB"/>
 			</Policies>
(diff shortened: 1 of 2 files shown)


holgerfriedrich:
Change to ```RollingRandomAccessFile``` does no harm on my system.

I wanted to try if it really fixes the issue by triggering the zram sync.
@mstormi can you tell me what to do that?
In #1654 you wrote ```systemctl start zsync```, but this service is not found (just installed zram via openHABian, and additionally did the update to be on the safe side). ```sudo /usr/local/sbin/zram-config sync``` leads to the output "zram-config sync 2024-05-15-18:09:41-CEST", but OH does a restart.

mstormi:
I had this on very first attempt, too, but after OH restart it's fine now

holgerfriedrich:
Not sure. I tried restarting OH, even clearing the cache, and finally reboot as well. Works until I trigger the sync, from then nothing is logged.

holgerfriedrich:
Not sure. I tried restarting OH, even clearing the cache, and finally reboot as well. Works until I trigger the sync, from then nothing is logged.

After a few hours there seem to be new log entries, but there is a gap. Maybe we need to lower the cache size...?!

mstormi:
> After a few hours there seem to be new log entries, but there is a gap. Maybe we need to lower the cache size...?!

What do you mean by "gap". Are there messages missing or do you actually mean "delay" ?
I wouldn't know how to reduce that. If I understood that stackoverflow comment correctly it's fixed not configurable anyway.
Try producing extra log lines by increasing log level for a test.

mstormi:
Testing shows the change doesn't fix the issue.
I'm still seeing it every day.

PR #1658, opened 2024-05-26
Update persistence extension breaking change text

mherwege:
We start seeing some problem descriptions in the forum with users using persistence extension actions with rules DSL. In certain cases there is a need to change the rule and add an explicit cast to DecimalType in the rule if one wishes to keep working without units, as State is not cast to a DecimalType automatically. See https://community.openhab.org/t/openhab-4-2-milestone-discussion/154316/140?u=mherwege and https://community.openhab.org/t/unknown-variable-or-command-in-rule/156191


holgerfriedrich (approved):

wborn (approved):
Thanks!

PR #1660, opened 2024-05-31
Add config template for USB addon finder

mherwege:
This is already there for the other addon finders that can be enabled/disabled. This one was missing.


wborn (approved):
Thanks!

PR #1663, opened 2024-06-21
Remove unused audit.log

### LINKED ISSUE #1662: [audit.log] Does anything ever get logged to this? Where do failed login attempts get logged?
Since OH 3.0 there's been this empty audit.log file generated in the logs folder. I've recently tried to see if I can make something log to it and have failed. 

1. Is this log actually used by anything?  The only logger that uses the AUDIT logger is "org.apacke.karaf.jaas.modules.audit" but I can't actually find out anything about this package with some brief looking. Maybe this package has gone away or moved?
2. I would have expected stuff like failed login attempts to appear in this log but after a few tries and putting the audit logger into TRACE level logging I could not generate any logs for failed logins nor any log statements at all to be logged into audit.log

Assuming I'm not missing something I would recommend that:

- the audit logger simply be removed
- whatever logs out login attempts (successful or failed) should be logged to openhab.log

If we start logging login attempts to audit.log I fear we've spent the past several years training people to not look in this log file and it'll be missed. But for a whole host of reasons (e.g. fail2ban) it is important that login attempts, both successful and failed, get logged out *somewhere*.

kaikreuzer:
Fixes #1662


wborn (approved):
LGTM!

PR #1664, opened 2024-06-29
Mention change regarding transformations when state is NULL/UNDEF

lolodomo:
Signed-off-by: Laurent Garnier <lg.hc@ree.fr>


kaikreuzer (approved):
Thanks!

PR #1665, opened 2024-06-29
Fix demo setup for dev (Basic UI + transform)

lolodomo:
Signed-off-by: Laurent Garnier <lg.hc@free.fr>

### DIFF
--- a/launch/app/runtime/services.cfg
+++ b/launch/app/runtime/services.cfg
@@ -6,9 +6,7 @@ folder:persistence=persist
 folder:things=things
 
 basicui:defaultSitemap=demo
-basicui:enableIcons=true
-basicui:condensedLayout=false
-basicui:capitalizeValues=false
+basicui:inlineSvg=false
 
 # Uncomment to enable rest api CORS requests
 # cors:enable=true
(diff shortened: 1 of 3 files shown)


launch/app/runtime/services.cfg
```diff
@@ -6,9 +6,7 @@ folder:persistence=persist
 folder:things=things
 
 basicui:defaultSitemap=demo
-basicui:enableIcons=true
-basicui:condensedLayout=false
-basicui:capitalizeValues=false
+basicui:inlineSvg=false
```
kaikreuzer, line 9: Could you briefly elaborate these changes? I am not fully clear, why they are done.

kaikreuzer: Ah, ok, it is the demo setup and you simply removed the things that are anyhow the default values, I guess.

kaikreuzer (approved):

launch/app/runtime/services.cfg
lolodomo: The Basic UI settings are now on client side (browser).
They do not exist anymore on OH server side, except inlineSvg and defaultSitemap.

PR #1697, opened 2024-11-29
Update services.cfg

lo92fr:
fix services.cfg to re-enable startup level 30 in development environment
Link to issue #3823 in openhab-core : https://github.com/openhab/openhab-core/issues/3823

### DIFF
--- a/launch/app/runtime/services.cfg
+++ b/launch/app/runtime/services.cfg
@@ -18,7 +18,8 @@ threadpool:safeCall=3
 
 # Start level definition
 startlevel:20=dsl:items,managed:item,dsl:things,managed:thing,managed:itemchannellink,dsl:persist
-#startlevel:30=persistence:restore // do not rely on the presence of a persistence service in the IDE
+startlevel:20=dsl:items,managed:item,dsl:things,managed:thing,managed:itemchannellink,dsl:persist,managed:metadata
+startlevel:30=persistence:services,persistence:restore,automation:scriptEngineFactories
 startlevel:40=dsl:rules,managed:rule,rules:refresh,rules:dslprovider
 startlevel:50=ruleengine:start
 startlevel:70=dsl:sitemap
--- a/launch/app/runtime/services.cfg
+++ b/launch/app/runtime/services.cfg
@@ -17,9 +17,8 @@ threadpool:discovery=3
 threadpool:safeCall=3
 
 # Start level definition
-startlevel:20=dsl:items,managed:item,dsl:things,managed:thing,managed:itemchannellink,dsl:persist
 startlevel:20=dsl:items,managed:item,dsl:things,managed:thing,managed:itemchannellink,dsl:persist,managed:metadata
-startlevel:30=persistence:services,persistence:restore,automation:scriptEngineFactories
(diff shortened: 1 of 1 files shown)


launch/app/runtime/services.cfg
```diff
@@ -18,7 +18,8 @@ threadpool:safeCall=3
 
 # Start level definition
 startlevel:20=dsl:items,managed:item,dsl:things,managed:thing,managed:itemchannellink,dsl:persist
-#startlevel:30=persistence:restore // do not rely on the presence of a persistence service in the IDE
+startlevel:20=dsl:items,managed:item,dsl:things,managed:thing,managed:itemchannellink,dsl:persist,managed:metadata
+startlevel:30=persistence:services,persistence:restore,automation:scriptEngineFactories
```
lolodomo, line 22: What about the current comment ?
In IDE, we should be able to start openHAB even without any persistence installed.
Your change should not change that.

lo92fr: Hum, don't look at the comment, sorry.

So I don't know what are good modification to solve this.
The matter is that we add new marker as part of @J-N-K modification on September 9.
And we except to openhab go to level 30 to have ThingManagerImpl correctly initialized !

Laurent.

lo92fr: Laurent,

This is related to the issue : https://github.com/openhab/openhab-core/issues/3823.
Don't know if you see my comment this morning about tracing why we have thing not get initialized in development environment.

lolodomo: Take care, you have two different lines for start level 20.

lolodomo:
Tested with 
```
startlevel:20=dsl:items,managed:item,dsl:things,managed:thing,managed:itemchannellink,dsl:persist,managed:metadata
startlevel:30=persistence:services,persistence:restore,automation:scriptEngineFactories
```
in Eclipse and I can confirm that our problem with managed things remaining in NOT_YET_READY status at startup is fixed.

lolodomo:
@openhab/distro-maintainers @openhab/core-maintainers : this is a critical fix to review ASAP, we are no more able since few weeks/months to use managed things when playing with Eclipse IDE.

@lo92fr : please adjust your fix as you kept two lines for start level 20.

launch/app/runtime/services.cfg
J-N-K: IMO you. eed to remove the second line for start level 20 and should remove the `persistence:restore` marker. The other two markers rely on bundle trackers and will be set when all installed bundles (persistence or scripting) are `ACTIVE` or no bundle of the given type is installed. `persistence:restore` in principle expects a persistence bundle to be installed which might not be the case in a development environment.

J-N-K (changes requested):

lo92fr:
@J-N-K, @lolodomo,

I've push the change about duplicate line and removing persistence:restore on startlevel:30.

Laurent.

lolodomo:
I just tried again with your last changes and I confirm that my managed thing is now ONLINE after start from Eclipse IDE.

@kaikreuzer : can you please merge ?

kaikreuzer (approved):
It fixes the issues in the IDE for me as well - many thanks!

PR #1705, opened 2024-12-16
Revert LinkedTransferQueue high CPU usage workaround

wborn:
Reverts the workaround (https://github.com/openhab/openhab-distro/pull/1636) that was introduced for [JDK-8301341](https://bugs.openjdk.org/browse/JDK-8301341) because it is fixed in OpenJDK 21 (https://github.com/openjdk/jdk/pull/14317).

Related to openhab/openhab-core#4499


holgerfriedrich (approved):
LGTM

wborn:
Maybe the config value state is still  persisted as false and we need to set it to true instead so it does not wait.

PR #1709, opened 2024-12-19
Add breaking changes alert for oh-clock-card background

florian-h05:
Should be backported to 4.3.x so it is shown there as well.


holgerfriedrich (approved):
LGTM, thanks!

PR #1711, opened 2024-12-28
Enable HTTP/2 for HTTPS

florian-h05:
Depends on https://github.com/openhab/openhab-core/pull/4526.

### DIFF
--- a/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
+++ b/distributions/openhab/src/main/resources/runtime/etc/jetty.xml
@@ -79,10 +79,10 @@
 		</Arg>
 	</Call>
 
-	<New id="httpConfig" class="org.eclipse.jetty.server.HttpConfiguration">
+	<New id="http2Config" class="org.eclipse.jetty.server.HttpConfiguration">
 		<Set name="secureScheme">https</Set>
 		<Set name="securePort">
-			<Property name="org.osgi.service.http.port.secure" default="8443" />
+			<Property name="org.osgi.service.http2.port.secure" default="8443" />
 		</Set>
 		<Set name="outputBufferSize">32768</Set>
 		<Set name="requestHeaderSize">8192</Set>
@@ -139,16 +139,21 @@
 					<Array type="org.eclipse.jetty.server.ConnectionFactory">
 						<Item>
 							<New class="org.eclipse.jetty.server.SslConnectionFactory">
-								<Arg name="next">http/1.1</Arg>
 								<Arg name="sslContextFactory">
 									<Ref refid="sslContextFactory" />
 								</Arg>
+								<Arg name="next">alpn</Arg>
 							</New>
 						</Item>
 						<Item>
-							<New class="org.eclipse.jetty.server.HttpConnectionFactory">
+							<New class="org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory">
+								<Arg>h2</Arg>
+							</New>
+						</Item>
+						<Item>
+							<New class="org.eclipse.jetty.http2.server.HTTP2ServerConnectionFactory">
 								<Arg name="config">
-									<Ref refid="httpConfig" />
+									<Ref refid="http2Config" />
 								</Arg>
 							</New>
 						</Item>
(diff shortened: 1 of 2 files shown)


J-N-K:
Isn't that removing http/1.1 support for secure connections?

florian-h05:
Yes, but I see no harm in doing so:
HTTP/2 is widely supported by across major browsers, most of them have HTTP/2 support since 2015, see https://caniuse.com/http2.

J-N-K:
I'm ok with that, but I would like to hear the opinion of other @openhab/distro-maintainers and @openhab/architecture-council

florian-h05:
Let's also ask @ghys for his opinion.

holgerfriedrich:
I had to keep http/1 and TLS 1.1 for quite a while longer than I wanted due to Android-based wall displays. But it is more than ten years since Android 4 and time to get rid of those devices anyway - if still in use. So 👍 from my side.

florian-h05:
If really needed, you could use nginx to get http/1 and TLS 1.1 available for them - but given the vulnerabilities of TLS 1.1 you could also consider to use plain HTTP ;-)

distributions/openhab/src/main/resources/runtime/etc/jetty.xml
```diff
@@ -79,10 +79,10 @@
 		</Arg>
 	</Call>
 
-	<New id="httpConfig" class="org.eclipse.jetty.server.HttpConfiguration">
```
J-N-K, line 82: No need to change the name - in my example I only changed it because I added a second server configuration to keep http/1.1 support.

distributions/openhab/src/main/resources/runtime/etc/jetty.xml
```diff
@@ -79,10 +79,10 @@
...
+	<New id="http2Config" class="org.eclipse.jetty.server.HttpConfiguration">
 		<Set name="secureScheme">https</Set>
 		<Set name="securePort">
-			<Property name="org.osgi.service.http.port.secure" default="8443" />
```
J-N-K, line 85: Don't change that when you remove http/1.1 support, otherwise the setting of the port from the usual openhab-configuration will not work.

J-N-K (changes requested):

distributions/openhab/src/main/resources/runtime/etc/jetty.xml
florian-h05: Reverted :+1:

distributions/openhab/src/main/resources/runtime/etc/jetty.xml
florian-h05: Sure - I changed the other occurences back but did not notice this one here.

J-N-K (approved):
Thanks. Let's wait for more comments.

launch/app/app.bndrun
florian-h05, line 1: Do we need this change now that we don’t add them as compile time dependencies on core?

J-N-K: Probably not.

rkoshak:
I think there will be much complaining but I also think this is somethjing that probably needs to be done anyway. Never underestimate the willingness of people to keep running stuff far beyond it's end of life. 

I would request an update to the reverse proxy docs to correspond with this change, assuming end users need to adjust their nginx or apache configs to account for this.

florian-h05:
> Never underestimate the willingness of people to keep running stuff far beyond it's end of life.

If one has a browser older than 2015 (where HTTP/2 was added to all major browsers) I wish them good luck to not get some malware while surfing …

kaikreuzer:
I guess 10 years after the general availability of a successor, it should be ok to drop HTTP/1.1+TLS1.1 - so no veto from me.

holgerfriedrich (approved):
LGTM. 

I managed to connect with HTTP2 and HTTP w. keepalive.
Let's see how it works out for the bigger audience.

launch/app/app.bndrun
florian-h05: Just tested, it is not needed, the demo app still starts fine.
I have created https://github.com/openhab/openhab-distro/pull/1712 to revert this change.

ghys:
> Let's also ask @ghys for his opinion.

Sorry I was AFK for the holidays but it's a good move IMO.
Would fix a lot of things although we'd still need to trust the self-signed TLS certificate on all clients for features like microphone, camera, web notifications etc. to work.

PR #1719, opened 2025-01-07
Upgrade plugins for demo app

holgerfriedrich:
I do not use Eclipse nor the demo app. But I think this should be in line with the other plugin upgrades I did.


wborn (approved):
Thanks!

PR #1722, opened 2025-02-02
Upgrade app dependencies to match core

holgerfriedrich:
Adapt demo app to core change openhab/openhab-core#4584


holgerfriedrich:
PR in core is merged, this one could be merged as well. The changes are only affecting app, this is why CI is fine even before we have a new core build from Jenkins.

wborn (approved):

mstormi:
Are you sure that all of these dependency updates apply to core changes that were introduced AFTER OH4?

I'm experiencing [trouble with bundles not becoming active](https://github.com/openhab/openhab-core/issues/4594) in latest release (4.3.2) and I'm not the only one. This started to happen after updating to 4.3-something. 4.2 was still fine it seems.

What about Shelly libraries?
Given the low number of bindings on those instances I'm seeing the problem with, they sound like a prime suspect.
@holgerfriedrich could you please have a look at the linked issue? Thanks.

PR #1726, opened 2025-02-20
Add note to update.lst for removal of juicenet binding

jsjames:


wborn (changes requested):
There's already a 5.0.0 section for alerts, can you add it to that instead?

https://github.com/openhab/openhab-distro/blob/192c7b7ec7c02407e6c47406e3b47b9caf4a87bd/distributions/openhab/src/main/resources/bin/update.lst#L162

jsjames:
Should be good now.  I didn't realize there was a separate section for alerts.

wborn (approved):

PR #1733, opened 2025-03-15
Replace GH Action from tj-actions by custom code

holgerfriedrich:
The GitHub Actions build workflow includes external Actions from 3rd party repositories.
The functionality of tj-actions/verify-changed-files was used to detect modifications of files under source control during the build process.
The external action is removed and replaced by custom code. This is less flexible, but does not require an external action in the build toolchain.

Currently, tj-actions is not available and anyway no longer allowed by our GH Actions configuration.
This PR needs to be picked to every active branch (or branches to be rebased).

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -73,8 +73,11 @@ jobs:
           path: build.log
 
       - name: Verify Changed Files
-        uses: tj-actions/verify-changed-files@v20
         id: verify-changed-files
+        run: |
+          set -o pipefail
+          changed_files=$(echo -n "$(git diff --name-only HEAD && git ls-files --others --exclude-standard)"|tr '\n' ' ')
+          echo "changed_files=$changed_files" >> $GITHUB_OUTPUT
 
       - name: Fail on Changed Files
         if: steps.verify-changed-files.outputs.changed_files != ''


wborn (changes requested):
It does not seem to be equivalent to the previous action as it does not fail on newly added files.

The check was added to prevent PRs that cause files to be changed whenever you do a full build. That is highly annoying as you have to remove/revert them manually each time you switch branches (https://github.com/openhab/openhab-webui/pull/2830).

PR testing the old setup: https://github.com/wborn/openhab-distro/pull/1
PR testing these changes: https://github.com/wborn/openhab-distro/pull/2

holgerfriedrich:
@wborn how shall we proceed?
Right now, tj-actions is still disabled for distro and core.
I could just add it to the config to re-enable it. It seems the repo has been restored, hopefully without the malicious code.

Or shall we go with one of the options we tried? WDYT?

.github/workflows/ci-build.yml
```diff
@@ -73,8 +73,11 @@ jobs:
           path: build.log
 
       - name: Verify Changed Files
-        uses: tj-actions/verify-changed-files@v20
         id: verify-changed-files
+        run: |
+          set -o pipefail
+          changed_files=$(echo -n "$(git diff --name-only HEAD 2>/dev/null ; git ls-files --others --exclude-standard 2>/dev/null)"|tr '\n' ' ')
```
wborn, line 79: Can you use `&&` instead of `;` so the exit code is not ignored?
Is it really necessary to redirect stderr to /dev/null? It might hide the reason why commands fail and I think only stdout is assigned based on the output.

wborn:
> Or shall we go with one of the options we tried? WDYT?

If you address the review comments it can be merged.

wborn (approved):
Thanks!

.github/workflows/ci-build.yml
holgerfriedrich: Done.
I added `dev/null` because saw one error message during my experiments on the command line.
But I agree, let's go without.

PR #1736, opened 2025-03-26
Improve GitHub Actions

holgerfriedrich:
* pin external action stCarolas/setup-maven to specific SHA
* set global timeout
* comments


wborn (approved):
Great!

PR #1738, opened 2025-04-08
Fix WebSocket connection not working for HTTPS

florian-h05:
Regression from #1711.
Fixes https://github.com/openhab/openhab-webui/issues/3125.

WebSocket was not working over HTTPS, because ALPN only had h2 available as protocol, but WebSocket is based on http/1.1. This adds http/1.1 to ALPN and the required HttpConnectionFactory as well.
"Standard" HTTP requests over HTTPS continue to use HTTP/2, whilst WebSocket works as well.

Also adds links to related JavaDoc as it is super annoying to search for the JavaDoc every time you need it.


wborn (approved):
Thanks for the fix!

PR #1745, opened 2025-05-30
Add semantic tags breaking change note

andrewfg:


andrewfg:
> or something to that effect.

That ("something to that effect") is what I thought I had written. Remember that each breaking change entry can only be a single line, and the we have already used 170+ characters...

mherwege (approved):
LGTM

andrewfg:
I think you should try to get this into v5.0

kaikreuzer (approved):

PR #1746, opened 2025-06-04
Create Maven "dev" profile

Nadahar:
While the "demo app" is sufficient for debug and testing is most cases, there are times when running "full Karaf" is necessary to get a proper understanding of how things work. I've created a Maven "dev" profile that makes it easier to get a local Karaf installation of OH up and running from locally developed code, which can be used both for testing and debugging.

The "dev" profile modifies the Maven build so that only the "openhab" module is built, and it builds to a folder instead of an archive. Each subsequent build will delete the "runtime" folder inside the built folder, so that subsequent builds can be done without "clean", allowing the configuration of the "OH installation" to persist across rebuilds. It can't be used with "mvn install", use "mvn package -Pdev" instead. The built folder is a ready to start OH installation, use "start_debug" if you want to connect a debugger.

More information can be found [here](https://community.openhab.org/t/how-to-make-a-local-snapshot-build/164298/3).

### DIFF
--- a/distributions/openhab/pom.xml
+++ b/distributions/openhab/pom.xml
@@ -280,4 +280,48 @@
     </plugins>
   </build>
 
+  <profiles>
+    <profile>
+      <id>dev</id>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-clean-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>delete-runtime</id>
+                <goals>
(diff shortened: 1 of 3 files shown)


florian-h05 (approved):
LGTM and works great, thanks!
@openhab/distro-maintainers Can you please have a look?

kaikreuzer (approved):
Neat idea, thanks!

PR #1774, opened 2025-07-24
Configuration: Predefine useTags

andrewfg:
Depends on https://github.com/openhab/openhab-core/pull/4913

### DIFF
--- a/distributions/openhab/src/main/resources/conf/services/runtime.cfg
+++ b/distributions/openhab/src/main/resources/conf/services/runtime.cfg
@@ -287,6 +287,6 @@
 
 ######################### ITEM CHANNEL LINK REGISTRY ###########################
 
-# If enabled, the ItemChannelLink Registry will always try to apply any tags from linked Channels to the respective Items
+# If enabled, the ItemChannelLink Registry will always try to apply any tags from the first linked Channel of the respective Item
 #
 #org.openhab.ItemChannelLinkRegistry:useTags=false


lolodomo:
The change looks good, the parameter is commented and so default will be used 

To avoid any change of behaviour for existing OH instances, I finally believe it is better to consider it disabled by default.
Power users like me will enabled it because we know the impact and want it.

andrewfg:
Ok. So I will leave this PR as is.

distributions/openhab/src/main/resources/conf/services/runtime.cfg
```diff
@@ -284,3 +284,9 @@
 # org.openhab.webclient:maxThreadsShared = 80
 # org.openhab.webclient:minThreadsCustom = 20
 # org.openhab.webclient:maxThreadsCustom = 40
+
+######################### ITEM CHANNEL LINK REGISTRY ###########################
+
+# If enabled, the ItemChannelLink Registry will always try to apply any tags from linked Channels to the respective Items
+#
+#org.openhab.ItemChannelLinkRegistry:useTags=true
```
jimtng, line 292: AFAIK, generally the commented out setting should be the _default_ value as it applies without specifying the setting. So in this case, it should be written as `=false` because that's what happens without the setting.

See other settings, e.g. `org.openhab.voice:cacheSizeTTS`, `org.openhab.sitemap:groupMembersSorting`, autoApprove, autoIgnore, although this is not entirely consistent in this file, e.g. allowBasicAuth defaults to false. In this case, we should probably fix allowBasicAuth but in a different PR.

andrewfg:
> take precedence over a change to this parameter made through the UI

Our assumption is that we are dealing with power users who are fluent with text files, so there is no code that would allow this to be changed via the UI.

rkoshak:
> Our assumption is that we are dealing with power users who are fluent with text files, so there is no code that would allow this to be changed via the UI.

Right, but if it is uncommented by default it would change the behavior for all users with no way to change it back short of editing the file. That's my point.

lolodomo (approved):

Issue #1780, opened 2025-07-25
The addons file is never installed during an update (Windows)


yndtrud:
OH: <=5.0.0
JAVA: zulu-jdk-17.60.17-ca-jdk17.0.16-win_x64 / temurin-jdk-21.0.8+9-ca-jdk21.0.8+9-win_x64
OS: Windows Server 2022 Std Latest
HW: ASUSPN41/IntelPentiumSilverN6000@2100/8GB_DDR4-2666/256GB_SSD-WD-GREEN

Script:
```
runtime\bin\update.ps1
```

Need replace:
```
$AddonsFile = "$OHAddons\openhab-addons-$OHVersionName.kar"
```
with 
```
$AddonsFile = "$OHAddons\openhab-addons-$CurrentVersion.kar"
```

```
        # If there's an existing addons file, we need to replace it with the correct version.
        try {
            $AddonsFile = "$OHAddons\openhab-addons-$OHVersionName.kar"
            if (Test-Path -Path $AddonsFile) {
                Write-Host "Found an openHAB addons file, replacing with new version"
                DeleteIfExists $AddonsFile
                DownloadFiles $AddonsDownloadLocation "$OHAddons\openhab-addons-$OHVersionName.kar"
            }
        } catch {
            return PrintAndThrow "Could not replace the $AddonsFile" $_
        }
```

wborn:
I think you're right!
Can you create a PR to fix it?

The file is maintained here:

https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/bin/update.ps1

PR #1782, opened 2025-08-04
The addons file is never installed during an update (Windows) #1780

yndtrud:
Need replace:
```
$AddonsFile = "$OHAddons\openhab-addons-$OHVersionName.kar"
```
with 
```
$AddonsFile = "$OHAddons\openhab-addons-$CurrentVersion.kar"
```

```
        # If there's an existing addons file, we need to replace it with the correct version.
        try {
            $AddonsFile = "$OHAddons\openhab-addons-$OHVersionName.kar"
            if (Test-Path -Path $AddonsFile) {
                Write-Host "Found an openHAB addons file, replacing with new version"
                DeleteIfExists $AddonsFile
                DownloadFiles $AddonsDownloadLocation "$OHAddons\openhab-addons-$OHVersionName.kar"
            }
        } catch {
            return PrintAndThrow "Could not replace the $AddonsFile" $_
        }
```


holgerfriedrich (changes requested):
Thanks!

I suppose that this does not work if you are on the snapshot.
`CurrentVersionName` does only contain the numbers, other parts are removed.

An approach could be to store the unmodified date to a new variable around ln. 395`$CurrentVersionName = $CurrentVersion` would be recommended. Then use it here.
Not sure if this works under all circumstances.

holgerfriedrich (approved):
LGTM, but I cannot test it right now.
@kaikreuzer, @wborn can you have a look, please?
Might be good to have this before M1.

kaikreuzer (approved):
I can't test now, but the change looks logical, so let's merge.

kaikreuzer:
FTR: Disregarding the failing DCO, since this is a minor change.

PR #1785, opened 2025-08-09
Increase default max size for thingHandler thread pool

lolodomo:
Changed from 5 to 15.

Related to discussion in openhab/openhab-addons#19125


lsiepel (approved):
Thanks LGTM

Will leave this open for some days to let others comment.

kaikreuzer:
I am not sure whether this is critical enough to backport to `5.0.x` - if someone requires a bigger setting, they can easily change it manually. For all other 5.0 users, it would mean that memory usage is suddenly increased through a "small" patch only.

PR #1798, opened 2025-08-24
Resolve app after core upgrades

holgerfriedrich:
updated dependencies after recent changes in core

I just tested the app, it starts up fine.


holgerfriedrich:
now included in #19266

wborn (approved):
👍

PR #1803, opened 2025-09-06
Fix Eclipse IDE setup

Nadahar:
Ever since bndtools 7.1.0 was released, OH's Eclipse setup has been broken. The problem has been identified, but they haven't released a patched version, so the fix won't arrive until 7.2.0. However, a snapshot version of 7.2.0 exists, but it has a different problem which still breaks things.

So, until this changes, it seems the best approach is to lock bndtools to 7.0.0 until a newer working version exists.

I made this several months ago, and no longer remember why `org.eclipse.m2e.feature.feature.group` was removed, but this has been tested by me and quite a few others and seems to work like it is. I had [a post on the forum](https://community.openhab.org/t/californium-removed/163625/78) that people have used to install Eclipse in the meanwhile, and as far as I know, this has resolved the problem for all that have used it. Now however, the links in the forum post are all dead (both to the setup file itself and to the screenshots explaining how to use it), so that's no longer an option, and I figure it's time to make this "temporary permanent".

See [this thread](https://community.openhab.org/t/osgi-resolve-gives-error/160641) for some of the related problems reported.

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2352,13 +2352,12 @@
       <requirement
           name="org.eclipse.m2e.logback.feature.feature.group"/>
       <requirement
-          name="org.eclipse.m2e.feature.feature.group"/>
-      <requirement
           name="org.eclipse.wst.jsdt.feature.feature.group"/>
       <requirement
           name="org.eclipse.wst.web_ui.feature.feature.group"/>
       <requirement
-          name="bndtools.m2e.feature.feature.group"/>
+          name="bndtools.m2e.feature.feature.group"
+		  versionRange="[7.0.0,7.1.0)"/>
       <requirement
           name="org.lastnpe.m2e.feature.feature.group"/>
       <requirement


Nadahar:
> Look at my comment in the community forum, looks like it does not work with that updated file in Linux (Mint).

@lolodomo  There must be something else going wrong. I have tested it for installations for Linux, macOS and Windows (on VMs). Also, if you look at the changes I've made, I haven't "removed" any project from the file.

Nadahar:
@lolodomo I've just tested the setup file on Mint 20.3, and it works just fine as far as I can see

Eclipse installer 2

lolodomo:
Ok so I am certainly very stupid and I don't understand how to provide this file ?
Is it in the window you show ? By clicking on the green "+" icon ?

lolodomo:
Is it a file I could replace somewhere in the "eclipse installer" so that I don't have to select it ?

wborn (approved):
LGTM!

wborn:
Yes Eclipse uses the updated file in this PR @lolodomo. Can you give it a try now it's merged?

Nadahar:
@lolodomo Did you finally get this to work now?

lolodomo:
It works for me now, just following the instructions in the documentation.
Great fix.

Nadahar:
> Could you pls. check if new 7.2.0 works for you?

I don't know who "you" refers to, but if it's me, I'm very reluctant to spend lots of hours struggling just because some new version is out. Generally, I never upgrade unless there's a reason to, because life is just too short to struggle with everything that breaks in this volatile world, and I especially don't want to be among "first adopters" since then I probably can't lean on other people's experiences when there's a problem.

What I know is that people who have tested the 7.2.0 snapshot has reported that while the particular issue that was broken in 7.1.0 had been fixed, but that there was another one still preventing OH from working with 7.2.0. If this has been fixed in the final version is anybody's guess, but I wouldn't hold my breath.

I can do some testing on a VM, but so can anyone basically, and I don't have time for that right now.

Nadahar:
I understand. While testing this specific issue is relatively easy, because it won't run at all when the bug is present, it means creating or cloning a VM, running the OH Eclipse installation, cloning all repos and a lot of building. After all that, it's just to try to run the "demo app" and see if it works. But, even if that particular thing works, how can you know that something else hasn't broken? 

The fact that the bug in 7.1 was discovered shortly after release, the developers themselves were quite shocked how this bug could have "slipped by" and ended up in a release, no patch release was done, and it has been around a year of waiting. In the meanwhile, people have had to try to use the snapshot, which has been broken in other ways. This doesn't give me a lot of confidence in their "release process", it seems like you have to consider yourself pretty lucky if you have a working setup.

So, given a couple of hours messing around with a VM, I could confirm whether 7.2.0 actually allows the "demo app" to resolve - but I'd have no idea of what other subtle bugs might lurk around. I think it would be better if some adventurous soul manually upgraded their BndTools to 7.2.0 in the Eclipse they use, and gained some experience whether it actually works. But, I'm a bad candidate for this, since I have a lot of bundles open, both from `addons`, `core` and `webui`. This makes rebuild operations extremely time-consuming at times (not always, but when it "really gets out of sync", which it typically does when BndTools acts up). As such, somebody that works a bit more focused, for example only with a few add-ons, would be in a much better position to do such experiments, because rebuilds will probably take minutes, not hours.

We've been using 7.1 in Core for a while, and it doesn't seem like that has lead to any conflict with using BndTools 7.0 in Eclipse, but of course, you never know.

Maybe it would be best to ask for some guinea pigs on the forum?

Nadahar:
I took a snapshot of a Linux VM where I have OH-Eclipse set up and working last night, and then upgraded Eclipse to latest and BndTools to 7.2. My first attempt to resolve was successful, but it should be noted that I had made no changes at this point, and the "solution" was already resolved from 7.0 - so I suspect that's why it worked. I then did some changes to what bundles to include and tried to resolve again, but it failed. And it failed whatever I did, also when I undid the changes to the bundles, so that it was back to what it originally was.

The failure is different from a "real failure"; as there is no "real information" about what's wrong. The error isn't the same as with [7.1 (project is null)](https://community.openhab.org/t/osgi-resolve-gives-error/160641). It complains about being unable to resolve `org.eclipse.equinox.metatype`, very similar to this: https://community.openhab.org/t/eclipse-resolution-problem-after-update-from-git/128421

I have never seen that error with 7.0, so I'm pretty sure that it's related to 7.2, and might be the very problem that others reported with the 7.2 snapshot. That said, it might be that it's possible to solve whatever is causing it - I don't know. I tried various things to no avail, and was stupid enough (I was tired) to be led to delete the contents of `~/.p2`. That was a mistake of course, because it basically deleted Eclipse. So, I had to revert to the snapshot, and now the VM is back in its working state - but it also means that I must do all the upgrading again to get to the point where I see the error.

I generally know too little about the mysterious workings of Bnd to solve such issues. They can be extremely frustrating because it just feels like you're left with far too little information to work with to actually get somewhere.

PR #1807, opened 2025-09-16
Add breaking change alerts for JS Scripting

florian-h05:
- Warn about breaking change for UI event object. (In fact the existing code will in most cases continue to work, it will print deprecation warnings to the log.)
- Advise to manually import the library into file-based scripts and transformations and turn off auto-injection.

Signed-off-by: Florian Hotze <dev@florianhotze.com>


florian-h05:
Yes, because I committed through GitHub web. I added the DCO to the PR description so you can copy it into the commit message of the squash merge.

miloit:
@florian-h05 What exactly do I need to add to my file based rules?

florian-h05:
```js
const { actions, items, rules } = require('openhab'); // every namespace can be imported this way
```

PR #1841, opened 2025-12-15
[persistence] refine persistence update instructions

mherwege:
Based on feedback, the removal of the default strategy may require more detail in the update instructions for file based configurations.

@lolodomo @kaikreuzer I believe this should cover the required changes if any. Many will not need changes (example: https://github.com/openhab/openhab-core/pull/4682#issuecomment-3651961853).

Note that the `default = ...` in the Strategy section that has to be removed was not even in the documentation (https://www.openhab.org/docs/configuration/persistence.html).


kaikreuzer (approved):
Thanks for the clarification, @mherwege!

PR #1862, opened 2026-01-18
Update update.ps1

seliSoft:
More robust backup handling
The upgrade failed on Windows 10, because the files are copied to a file named "home" instead of a the directory home. Simmilar thing happend to conf/userdata/runtime
With the changes the process is more robust and with the correct folders it is able to restore the files later.

### DIFF
--- a/distributions/openhab/src/main/resources/bin/update.ps1
+++ b/distributions/openhab/src/main/resources/bin/update.ps1
@@ -637,10 +637,16 @@ Function Update-openHAB() {
         Write-Host -ForegroundColor Cyan "Creating backup directories in $TempBackupDir"
         DeleteIfExists $TempBackupDir $True
 
-        Write-Host -ForegroundColor Cyan "Copying directory conf, userdata and runtime to $TempBackupDirConf"
-        Copy-Item -Path $OHConf, $OHUserData, $OHRuntime -Destination $TempBackupDir -Recurse -Force -ErrorAction Stop
+        Write-Host -ForegroundColor Cyan "Copying directory conf, userdata and runtime to $TempBackupDir"
+        New-Item -Path $TempBackupDir -Name "conf" -ItemType "Directory"
+		Copy-Item -Path $OHConf -Destination $TempBackupDirConf -Recurse -Force -ErrorAction Stop
+		New-Item -Path $TempBackupDir -Name "userdata" -ItemType "Directory"
+		Copy-Item -Path $OHUserData -Destination $TempBackupDirUserData -Recurse -Force -ErrorAction Stop
+		New-Item -Path $TempBackupDir -Name "runtime" -ItemType "Directory"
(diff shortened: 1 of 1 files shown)


distributions/openhab/src/main/resources/bin/update.ps1
```diff
@@ -637,10 +637,16 @@ Function Update-openHAB() {
         Write-Host -ForegroundColor Cyan "Creating backup directories in $TempBackupDir"
         DeleteIfExists $TempBackupDir $True
 
-        Write-Host -ForegroundColor Cyan "Copying directory conf, userdata and runtime to $TempBackupDirConf"
-        Copy-Item -Path $OHConf, $OHUserData, $OHRuntime -Destination $TempBackupDir -Recurse -Force -ErrorAction Stop
+        Write-Host -ForegroundColor Cyan "Copying directory conf, userdata and runtime to $TempBackupDir"
+        New-Item -Path $TempBackupDir -Name "conf" -ItemType "Directory"
```
wborn, line 641: `$TempBackupDir` was just removed above, so it does not normally exist when this tries to create `conf` relative to it.

Pre-creating `$TempBackupDirConf` also changes the semantics of the following `Copy-Item`: when the destination directory already exists, copying `$OHConf` can create another directory level below it.

The existing `CreateDirectory` helper could keep this simpler:

```powershell
CreateDirectory $TempBackupDir

Copy-Item -Path $OHConf -Destination $TempBackupDirConf -Recurse -Force -ErrorAction Stop
Copy-Item -Path $OHUserData -Destination $TempBackupDirUserData -Recurse -Force -ErrorAction Stop
Copy-Item -Path $OHRuntime -Destination $TempBackupDirRuntime -Recurse -Force -ErrorAction Stop

CreateDirectory $TempBackupDirHome
```

This still gives the backup the fixed `conf`, `userdata`, and `runtime` names expected by the restore logic, independent of the configured source directory names.

wborn (changes requested):
This PR was reviewed by AI first, before manual maintainer review.

The change addresses a valid problem: the temporary backup needs predictable `conf`, `userdata`, `runtime`, and `home` paths because those paths are used by the restore logic, while the source directories can be configured through environment variables.

There is one issue with the current implementation that should be addressed before merging. The code removes `$TempBackupDir` and then immediately tries to create child directories using that removed directory as the `New-Item -Path`. It also pre-creates the destinations before copying the source directories, which changes the behavior of `Copy-Item` and can introduce an additional directory level.

The existing `CreateDirectory` helper can be used to keep the fix simpler and make the intended backup layout explicit.

Since this backup is also the rollback mechanism when an update fails, it would be useful to manually verify the backup and restore paths with Windows PowerShell 5.1. The current CI build runs on Ubuntu and does not exercise this Windows-specific behavior.

PR #1864, opened 2026-01-21
[audio] Reencode the doorbell.mp3 sample file

dilyanpalauzov:
This is the same as https://github.com/openhab/openhab-distro/pull/1579 , but for the other copy of doorbell.mp3.


kaikreuzer (approved):
Thanks. Yeah, I guess it is currently stored twice. Haven't tried any symlinking yet and this file here hardly ever changes, so I'm not keen on investing any time on the investigation...

PR #1884, opened 2026-03-07
gzip compress assets and images

joerg1985:
This PR will enable gzip compression for `/assets/*` and `/images/*` with more mime types.
In the past the precompressed files where used, but this is currently broken and a deprecated Jetty feature.
So i does not make sense for me to fix using the precompressed files, as the support has been removed in Jetty 10+.

Before the 5 MB javascript was not compressed:
grafik

Now the 5 MB file is compressed to 1.16 MB
grafik


kaikreuzer:
> In the past the precompressed files were used

Do they still exist anywhere and should they now be removed?

PR #1891, opened 2026-03-29
Release notes workflow: Trigger version update for Homebrew formula

florian-h05:


kaikreuzer (approved):

kaikreuzer:
I wonder if we should rename the workflow as it does much more than "creating release notes"...

florian-h05:
Might be a good idea … „Release to GitHub“?

kaikreuzer:
Or "Create Release"?

florian-h05:
„Create GitHub Release“
IMO that clarifies a bit that the actual release to JFrog, Docker, Linux packages etc. is not handled by GHA.

PR #1895, opened 2026-04-20
Backup script: Fix exclusions not working & Include Piper/Vosk/Whisper dirs in full backup

florian-h05:
- The exclusions added in #1861 weren't working.
- Do not exclude Piper, Vosk & Whisper model dirs from full backup.


holgerfriedrich (approved):
LGTM

holgerfriedrich (approved):

PR #1905, opened 2026-05-06
Adapt deployment URL specification

holgerfriedrich:
I noticed the following warnings during M3 build:

~~~
[INFO] Using alternate deployment repository jfrog::default::https://openhab.jfrog.io/openhab/libs-milestone-local
[WARNING] Using legacy syntax for alternative repository. Use "jfrog::https://openhab.jfrog.io/openhab/libs-milestone-local" instead.
~~~

This PR tries to adapt the URL. I cannot test, though.


kaikreuzer (approved):
Thanks - let's see if it works the next time. :-)

PR #1907, opened 2026-05-10
[Oomph installer] Add workarond for lastnpe in recent Eclipse

### LINKED ISSUE #1856: Eclipse 2025-12 - Oomph installer fails
Following the [official steps ](https://www.openhab.org/docs/developer/ide/eclipse.html) to setup the development environment fails with the latest Eclipse release **2025-12**. 

**Error message:**  Missing requirement: eclipse-external-annotations-m2e-plugin 2.0.0.202212171629 (org.lastnpe.m2e.core 2.0.0.202212171629) requires 'osgi.bundle; org.apache.commons.io 0.0.0' but it could not be found

See attached logs for full-details: 
[log.txt](https://github.com/user-attachments/files/24531463/log.txt)

**Workaround:** Manually select version **2025-09** in the first page of the installer.
Maybe the developer guide should mention trying an older version if the newest fails 😉

Image

holgerfriedrich:
lastnpe m2e plugin requires commons.io, which is no longer available in Eclipse starting with 2025-12, as it was renamed to commons.commons-io.

Until lastnpe is released with a new requirement in MANIFEST.MF, an older installation source is required to safisfy this import. This leads to commons.io and commons.commons-io to be installed in parallel.

Fixes #1856

### DIFF
--- a/launch/openHAB2.setup
+++ b/launch/openHAB2.setup
@@ -2372,6 +2372,10 @@
           url="https://ianbrandt.github.io/m2e-maven-dependency-plugin/"/>
       <repository
           url="https://www.lastnpe.org/eclipse-external-annotations-m2e-plugin-p2-site/m2e_2/"/>
+      <!-- workaround: lastnpe requires org.apache.commons.io which was renamed
+          to org.apache.commons.commons-io in Eclipse 2025-12 -->
+      <repository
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
Thanks!

PR #1910, opened 2026-05-14
[GHA] Schedule resolution of app dependencies

holgerfriedrich:
Introduce the same scheduled resolver and PR creation as in core and add-ons, adapted to resolve demo app dependencies.

Manually resolve app as it is currently broken.


lsiepel (approved):
Thanks, LGTM

kaikreuzer (approved):

PR #1913, opened 2026-05-25
Include core.model.yaml bundle to demo app

lolodomo:


lolodomo:
The idea is to be able to work out of the box with YAML core bundle in Eclipse without the need to adjust app.bndrun.

holgerfriedrich (approved):
LGTM

PR #1914, opened 2026-05-25
Adjust setup of start levels to consider YAML providers

### LINKED ISSUE #5602: OH start level ignoring YAML providers
I am afraid that YAML providers were not considered to control the start level at OH startup.

The setup is defined in that file:
https://github.com/openhab/openhab-distro/blob/b88286759d7495dd2ca353c99c8b2b005bf72b5c/distributions/openhab/src/main/resources/runtime/services.cfg#L16
```
# Start level definitions
startlevel:20=dsl:items,managed:item,dsl:things,managed:thing,managed:itemchannellink,dsl:persist,managed:metadata
startlevel:30=persistence:services,persistence:restore,automation:scriptEngineFactories
startlevel:40=dsl:rules,managed:rule,rules:refresh,rules:dslprovider
startlevel:50=ruleengine:start
startlevel:70=dsl:sitemap
startlevel:80=things:handler
```

I assume we should consider YAML providers to setup the start levels, the same way as for DSL providers ?
WDYT ?

PS: Note also that there is nothing for managed persistence defined for level 20.

lolodomo:
Depends on #1913
Depends on openhab/openhab-core#5603
Depends on openhab/openhab-core#5604
Depends on openhab/openhab-core#5605
Closes openhab/openhab-core#5602

### DIFF
--- a/distributions/openhab/src/main/resources/runtime/services.cfg
+++ b/distributions/openhab/src/main/resources/runtime/services.cfg
@@ -14,11 +14,11 @@ threadpool:safeCall=10
 threadpool:thingHandler=15
 
 # Start level definitions
-startlevel:20=dsl:items,managed:item,dsl:things,managed:thing,managed:itemchannellink,dsl:persist,managed:metadata
+startlevel:20=dsl:items,yaml:items,managed:item,dsl:things,yaml:things,managed:thing,managed:itemchannellink,dsl:persist,managed:persistenceserviceconfiguration,managed:metadata
 startlevel:30=persistence:services,persistence:restore,automation:scriptEngineFactories
-startlevel:40=dsl:rules,managed:rule,rules:refresh,rules:dslprovider
+startlevel:40=dsl:rules,yaml:rules,managed:rule,rules:refresh,rules:dslprovider
 startlevel:50=ruleengine:start
-startlevel:70=dsl:sitemap
+startlevel:70=dsl:sitemaps,yaml:sitemaps,managed:sitemap,yaml:pages,yaml:widgets
 startlevel:80=things:handler
 
 # Auto-update event behavior
--- a/distributions/openhab/src/main/resources/runtime/services.cfg
+++ b/distributions/openhab/src/main/resources/runtime/services.cfg
@@ -14,7 +14,7 @@ threadpool:safeCall=10
 threadpool:thingHandler=15
 
 # Start level definitions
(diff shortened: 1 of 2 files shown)


distributions/openhab/src/main/resources/runtime/services.cfg
```diff
@@ -14,11 +14,11 @@ threadpool:safeCall=10
 threadpool:thingHandler=15
 
 # Start level definitions
-startlevel:20=dsl:items,yaml:items,managed:item,dsl:things,yaml:things,managed:thing,managed:itemchannellink,dsl:persist,managed:metadata
+startlevel:20=dsl:items,yaml:items,managed:item,dsl:things,yaml:things,managed:thing,managed:itemchannellink,dsl:persist,managed=persistenceserviceconfiguration,managed:metadata
```
mherwege: Shouldn’t that be : iso =.

lolodomo: Yes, must not be = 
I will fix it.

lolodomo: Fixed

distributions/openhab/src/main/resources/runtime/services.cfg
```diff
@@ -14,11 +14,11 @@ threadpool:safeCall=10
 threadpool:thingHandler=15
 
 # Start level definitions
-startlevel:20=dsl:items,managed:item,dsl:things,managed:thing,managed:itemchannellink,dsl:persist,managed:metadata
+startlevel:20=dsl:items,yaml:items,managed:item,dsl:things,yaml:things,managed:thing,managed:itemchannellink,dsl:persist,managed:persistenceserviceconfiguration,managed:metadata,yaml:tags,managed:semantictag
```
mherwege, line 17: Did you check if this does not create problems with the semantics service depending on metadata?

lolodomo: I just checked that all start levels are reached.

lolodomo: I could remove added markers for semantic tags in case you think there is something wrong but I can't do it before this evening.

mherwege: If all start levels are reached, we should be fine I think.

kaikreuzer (approved):

PR #1929, opened 2026-06-26
Persistence extensions change alert

mherwege:
https://github.com/openhab/openhab-core/pull/4951 adjust the unit returned in some persistence extensions to be the unit of the input argument, instead of the system unit. This may break assumptions made in rules, therefore this alert.


kaikreuzer (approved):
thanks!

PR #1930, opened 2026-06-28
Add marker scripts:refresh as condition for start level 40

lolodomo:
Related to openhab/openhab-core#5665


lolodomo:
@kaikreuzer : as discussed in core PR.
I let you decide if you want to add that before or after 5.2 release.
As already mentioned, OH can work perfectly even without that change.

kaikreuzer:
Thanks. Let's stick to what we communicated: Only critical fixes to be merged before GA, so let's keep it until after the release.

holgerfriedrich (approved):

PR #1944, opened 2026-08-09
Remove pax-web patching by openhab-util library

holgerfriedrich:
This patching step has no longer worked for quite a while.

`findResources` was removed from pax-web upstream in January 2020 (commit `cf6861615`, part of the PAXWEB-1190 Http/Whiteboard refactor), replaced by the current `findEntries(...)` methods - over a year and a half *before* pax-web `8.0.0` GA even shipped (September 2021). Every pax-web 8.0.x version has no method by that name.

`ClassAdapter`'s method-matching (`"findResources".equals(name)`) therefore never fires, and `ClassPathUtilPatcher` has no verification step - it always prints `"Finished updating ..."` and exits 0 regardless of whether anything was actually patched.

@kaikreuzer Is seems you were right in https://github.com/openhab/openhab-util/pull/1#issuecomment-4186851515

### DIFF
--- a/distributions/openhab/pom.xml
+++ b/distributions/openhab/pom.xml
@@ -82,22 +82,6 @@
         </includes>
       </resource>
     </resources>
-    <pluginManagement>
-      <plugins>
-        <plugin>
-          <groupId>org.codehaus.mojo</groupId>
-          <artifactId>exec-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


wborn (approved):
Nice to see it's no longer needed!

PR #1945, opened 2026-08-13
Fix Windows startup helper script handling

### LINKED ISSUE #1940: Popups and error messages during initial startup of openHAB 5.3 snapshot in Windows 10
When running `start.bat` I get this popup after a while:

Image

"Den angivne fil blev ikke fundet" translates to something like "The specified file was not found".

I suspect this might be a regression of #1922, since I haven't updated my openHAB development system for a while, and this mentioned "runhidden.vbs" seems to have been added by that PR:

https://github.com/openhab/openhab-distro/blob/b04095e08998763f43d6cd286fe634409cd474c6/distributions/openhab/src/main/resources/bin/karaf.bat#L107

On second run, the situation changed:
```text
Launching the openHAB runtime...

                           _   _     _     ____
   ___   ___   ___   ___  | | | |   / \   | __ )
  / _ \ / _ \ / _ \ / _ \ | |_| |  / _ \  |  _ \
 | (_) | (_) |  __/| | | ||  _  | / ___ \ | |_) )
  \___/|  __/ \___/|_| |_||_| |_|/_/   \_\|____/
       |_|       5.3.0-SNAPSHOT - Build #5535

Use '<tab>' for a list of available commands
and '[cmd] --help' for help on a specific command.
To exit, use '<ctrl-d>' or 'logout'.

openhab> Error executing command: Terminal has been closed
openhab>
```

On one hand, it might have just been a small glitch to be ignored. On the other hand, isn't this message "Error executing command: Terminal has been closed" what should have been fixed by that PR?

andrewfg:
In #1922 we fixed the 'Terminal has been closed' error. The solution for Windows requires a helper Visual Basic Script file that executes some shell commands in a hidden window in order to avoid confusing users. The original solution created the helper VBS file on the fly in the Windows `temp` folder. However as reported and discussed in #1940 creating the script on the fly can create problems; specifically due to paths with spaces. And it leaves an orphan copy of the script file in the Windows `temp` folder.

In this PR we no longer create the helper Visual Basic Script on the fly, but instead we include it as a fixed part of the distribution. Furthermore the batch file and the script are modified to handle paths with spaces on Windows platforms.

Furthermore based on tests with slower machines, this PR increases the timeout durations accordingly.

Resolves #1940

### DIFF
--- a/distributions/openhab/src/main/resources/bin/karaf.bat
+++ b/distributions/openhab/src/main/resources/bin/karaf.bat
@@ -126,7 +126,11 @@ if defined LEVEL (
     )
 )
 if !SECONDS_WAITED! GEQ 30 (
-    echo Wait timed out!
+    if not exist "%STARTLEVEL_FILE%" (
+        echo Wait timed out! .. missing start-level file %STARTLEVEL_FILE%
+        goto SERVER_STARTED
+    )
+    echo Wait timed out! .. did not reach start-level %STARTED_TARGET_LEVEL%
     goto SERVER_STARTED
 )
 timeout /t 1 /nobreak >nul
@@ -152,7 +156,11 @@ if defined LEVEL (
     )
 )
 if !SECONDS_WAITED! GEQ 30 (
-    echo Wait timed out!
+    if not exist "%STARTLEVEL_FILE%" (
+        echo Wait timed out! .. missing start-level file %STARTLEVEL_FILE%
+        goto SERVER_STOPPED
+    )
+    echo Wait timed out! .. did not reach start-level %STOPPED_TARGET_LEVEL%
     goto SERVER_STOPPED
 )
 timeout /t 1 /nobreak >nul
--- a/distributions/openhab/src/main/resources/bin/karaf.bat
+++ b/distributions/openhab/src/main/resources/bin/karaf.bat
@@ -93,6 +93,11 @@ goto AFTER_CACHE_READY
 set CACHE_FOLDER=%KARAF_DATA%\cache\org.eclipse.osgi
 if exist "%CACHE_FOLDER%\.*" goto AFTER_CACHE_READY
 
+if not exist "%DIRNAME%runhidden.vbs" (
+    call :warn Missing helper script: "%DIRNAME%runhidden.vbs"
+    goto END
+)
+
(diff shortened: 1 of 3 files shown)


distributions/openhab/src/main/resources/bin/karaf.bat
```diff
@@ -103,8 +103,12 @@ set STARTED_TARGET_LEVEL=10
...
 > "%STARTLEVEL_FILE%" echo %STOPPED_TARGET_LEVEL%
 
-set RUNHIDDEN=wscript //nologo "%TEMP%\runhidden.vbs"
-> "%TEMP%\runhidden.vbs" echo CreateObject("Wscript.Shell").Run WScript.Arguments(0), 0, False
+set RUNHIDDEN=wscript //nologo "%DIRNAME%runhidden.vbs"
+
+if not exist "%DIRNAME%runhidden.vbs" (
+    call :warn Missing helper script: "%DIRNAME%runhidden.vbs"
+    goto END
+)
```
wborn: The cache-refresh lock has already been created when this branch is reached. Jumping directly to `END` leaves that lock behind, so the next startup will detect the existing lock and skip `AWAIT_CACHE_READY` entirely, including this helper-file check.

Could the helper be checked before creating the lock, or could the lock be removed before exiting here? Moving the check before creating `CACHE_REFRESH_LOCK` would probably be the simpler option.

wborn (dismissed):
This PR is being reviewed with AI first.

The overall approach looks good: shipping `runhidden.vbs` as part of the distribution avoids creating the helper dynamically in `%TEMP%` and removes the associated timing, permission, path, and cleanup concerns.

One correctness issue should be addressed before merging: when `runhidden.vbs` is missing, the new error path exits after `cache-refresh-lock` has already been created. That stale lock causes subsequent startups to skip the cache refresh path entirely, including the helper-file check.

Apart from that issue, no further problems were found in the current changes. A human maintainer review is still needed.

andrewfg:
@wborn I addressed your (AI) comment. However can you please wait for @jlaur to confirm that it fixes his #1940

andrewfg:
@jlaur can you please confirm?

jlaur (approved):
Thanks, LGTM.

andrewfg:
@wborn this is ready now

wborn (approved):
Thank you!

PR #1949, opened 2026-08-17
Fix mvn signature, use sha256sum

holgerfriedrich:
dependabot now handles `mvnw` and `mvn` upgrades.
sha512 signature is anyway useless as it is ignored by mvnw.

This PR replaces sha512 checksum by sha256. It will then be updated by dependabot as well.

Please merge this before #1947 and #1948, I want to see dependabot recreate do the upgrade correctly afterwards :-)


wborn (approved):
I also noticed it didn't update it. Let's see what happens now.

PR #1955, opened 2026-08-24
[chatgpt] Add note about HLI service per Thing

florian-h05:
Refs https://github.com/openhab/openhab-addons/pull/21468.


kaikreuzer:
Hm, this is only "breaking" for any previous 5.3.0-SNAPSHOT, right? We didn't have ChatGPT-HLI in 5.2 yet and there isn't yet even a 5.3 milestone. I'd say that in that case we should actually NOT add any note here, wdyt?

florian-h05:
ChatGPT actually had a HLI since https://github.com/openhab/openhab-addons/pull/17320, which shipped with 4.3.
That HLI was available under the `chatgpt` ID, however with no support for the chat UI until now, I don't think the change will affect that many users. I guess there weren't many users of the `chatgpt` HLI service.

kaikreuzer:
Ok, thanks for the clarification!

kaikreuzer (approved):

